Issue #88, opened 2017-02-04
Allow auto-hiatus posts to be marked hiatus


Throne3d:
> What it says on the box

Marri:
Fixed in #91

PR #91, opened 2017-02-04
Add a 'mark hiatus' button to autohiatused threads

Throne3d:
It seems slightly obscure (or perhaps more than slightly obscure) if you're trying to remove a thread from your tags owed page – currently you have to mark it some non-active status, then mark it hiatus, as the "mark hiatus" button doesn't show up when it's technically active (and hence on your tags owed page) but automatically hiatus.

This also tries to keep the text clear – if the thread is explicitly marked on hiatus, marking it 'active' might not actually make it 'active' per se (if it's been more than a month since the latest reply), so if it's on hiatus, the text will show "Unmark Hiatus" instead of "Mark In Progress".

Fixes issue #88?

### DIFF
--- a/app/views/posts/show.haml
+++ b/app/views/posts/show.haml
@@ -70,16 +70,19 @@
             %div
               = image_tag '/images/book.png'.freeze
               Mark Complete
-        - unless @post.on_hiatus?
+        - unless @post.marked_hiatus?
           = link_to post_path(@post, status: 'hiatus'.freeze), method: :put do
             %div
               = image_tag '/images/hourglass.png'.freeze
-              Put On Hiatus
+              Mark On Hiatus
         - unless @post.active?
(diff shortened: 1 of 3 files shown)


app/views/posts/show.haml
```diff
@@ -70,16 +70,19 @@
...
           = link_to post_path(@post, status: 'hiatus'.freeze), method: :put do
             %div
               = image_tag '/images/hourglass.png'.freeze
-              Put On Hiatus
+              Mark On Hiatus
         - unless @post.active?
           = link_to post_path(@post, status: 'active'.freeze), method: :put do
             %div
               = image_tag '/images/book_open.png'.freeze
-              Mark In Progress
```
Marri: Can we leave this as it was? I'd rather it be consistent.

Issue #73, opened 2017-02-04
Link to hide a thread in posts#show


Throne3d:
> Requested by Pedro: "Having the option to hide a thread from Unread within the thread would be cool."

Marri:
Fixed in #92

PR #92, opened 2017-02-04
Add a hide/unhide thread link in post#show

Throne3d:
Adds tests for some post#update, and fixes issue #73.


spec/controllers/posts_controller_spec.rb
```diff
@@ -639,6 +650,33 @@
...
+      it "marks hidden" do
+        post = create(:post)
+        user = create(:user)
+        login_as(user)
+        expect(post.ignored_by?(user)).not_to be_true
+
+        put :update, id: post.id, hidden: 'true'
+        expect(response).to redirect_to(post_url(post))
+        expect(flash[:success]).to eq("Post has been hidden")
+        expect(post.reload.ignored_by?(user)).to be_true
```
Marri: Can you also check that this doesn't muck with the read_at timestamp? I've seen that happen because of how I overloaded the views.

PR #93, opened 2017-02-04
Include content_warnings, order characters by scope

Throne3d:
Removes some N+1 queries


Throne3d:
This is done in other PRs for the most part and I don't care enough about using `includes` for content warnings on the report page.

PR #94, opened 2017-02-04
Set post defaults for ease of generation of posts in a dev environment

Throne3d:
- Set last_user in a before_create callback (instead of the controller)
- Set default content to be ''

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -87,7 +87,7 @@ class PostsController < WritableController
     preview(:post, posts_path) and return if params[:button_preview].present?
 
     @post = Post.new(params[:post])
-    @post.user = @post.last_user = current_user
+    @post.user = current_user
 
     create_new_tags if @post.valid?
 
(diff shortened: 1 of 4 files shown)


db/migrate/20161229204445_set_default_content_for_posts.rb
```diff
@@ -0,0 +1,8 @@
+class SetDefaultContentForPosts < ActiveRecord::Migration
+  def up
+    change_column_default :posts, :content, ''
```
Marri: If this is literally only useful when manually creating posts in the console, I'd rather skip the database migration, they're kind of a pain.

Throne3d: I'm not too strongly opinionated about *when* it gets merged, so… we could just wait until we plan on doing something else in a migration?

Marri: This shouldn't need a migration at all, though? Do we just need a .to_s somewhere, or something, this isn't something that ever happens in the actual correct user flow or even really in tests.

Throne3d: Hmm, good point – I'll look into it tomorrow.

Throne3d:
Now uses a `.to_s` call like suggested.

Issue #46, opened 2017-02-04
List editor-allowed markup


Throne3d:
> List the allowed markup (`<b>`, `<i>`, etc) somewhere in the post editor

Throne3d:
Fixed in #96.

PR #96, opened 2017-02-05
Improve the post editor by adding a help section ("?" button)

Throne3d:


app/views/posts/_write_post.haml
```diff
@@ -25,6 +25,8 @@
     - unless post.id.present? && !post.editable_by?(current_user)
       .view-button#rtf{class: ('selected' if current_user.default_editor == 'rtf')} Rich Text
       .view-button#html{class: ('selected' if current_user.default_editor == 'html')} HTML
+      .view-button#markup-allowed{title: "Toggle RTF editor and HTML editor\nAllowed tags: #{Glowfic::ALLOWED_TAGS.sort * ', '}"} ?
+      - # title gets overridden by JS
```
Marri: Why is the title getting overridden by Javascript?

app/views/posts/_write_post.haml
```diff
@@ -41,6 +43,18 @@
         = f.hidden_field :icon_id, id: 'reply_icon_id'
         = submit_tag (method == :put ? 'Save' : "Post"), class: 'button', id: 'submit_button', disable_with: 'Saving...'
         = submit_tag "Preview", class: 'button', id: 'preview_button', name: 'button_preview'
+      %br
+      #markup-allowed-box{style: 'display:none'}
```
Marri: Is this complex enough to be worth its own partial in case we change it later? No PR description here but I vaguely recall we might want to UI it slightly more later.

app/views/posts/_write_post.haml
Throne3d: Because it's extremely long and this is for NoScript users. When Javascript is enabled, it truncates the title but the 'spoiler' box is usable and contains the same / more information.

app/views/posts/_write_post.haml
Throne3d: Possibly? I'm not sure what we'd actually change the UI *to*, because people don't give much input on this, unfortunately, and I don't predict we'll get around to changing it anytime soon, so I think I'd be inclined against, but not particularly strongly so.

Marri:
I'll live with the NoScript thing because I don't have NoScript so I feel I should not be opining on how it works for them, but my more generic concern is that at least on the post page, I have literally no idea that the button worked, because on my screen the tags div appears out of sight. Not sure if we want to put them somewhere else or move the view window with Javascript or what?

Marri:
Sorry :( didn't mean to close it, I was used to the "Add single comment / Start full review" buttons when commenting on code, aimed for the button on the left expecting it to be "Add single comment" and it was "Close and comment".

Throne3d:
Np! Hmm… if we can check if it's out of sight, scrolling down to it could be good? Or I could improve the UI now by making it a floating box… I'm not sure about that, though, it seems slightly weird if we only have one or maybe two other (small) elements that float.

Throne3d:
(Relates to #46.)

Throne3d:
Above, one of the things marked as "Show outdated comment" shouldn't be – there was still relevant conversation there.

Throne3d:
Still to be resolved:

- [ ] Consider that NoScript users are already having trouble with lots of things and tbh this wouldn't be the biggest problem for them, so we can simplify it by removing that handling
- [ ] Consider moving the box to a partial
- [ ] Scroll to box when it's opened, to make it clearer?

Throne3d:
It still looks just as disgusting as it used to :P 
image

It at least contains all the appropriate information. I'm not sure how to go about prettifying it.

Throne3d:
I need to figure out how to make this less ugly before merging it.

Throne3d:
Updated screenshot:
image

Still ugly. I was gonna prompt people on how to make it better – maybe a popup in the middle of the screen would be better.

Throne3d:
It got far less disgusting and made friends with some other information boxes!

image

(And is now merged.)

Issue #45, opened 2017-02-04
Improve the pagination display


Throne3d:
Could be part of #43.

- Mobile view should be done by CSS (max-width) instead of browser detection
- Allow a dropdown for mobile view, if people want to be able to switch pages (perhaps make one of the bits of text into, visibly, a 'link', which when pressed produces a dropdown option)

Issue #43, opened 2017-02-04
Layout options


Throne3d:
Allow layouts to be stacked with *themes*. So that "No Icon" can be stacked with "Starry Dark", for example.

Example layouts:
- [X] No Icon
- [ ] Right-sided reply info boxes

Throne3d:
See discussion [around here](http://alicorn.elcenia.com/board/viewtopic.php?f=12&t=459&start=1870#p31995).

dhouck:
It seems like some of these are orthogonal:

 * What info do you want in character boxes (icons? Screen name? Author name?)
 * Do you want to see reply info above/to the right of/to the left of the replies? (Would anybody want “below”? It feels weird listing three of the four standard options but I can’t see a use for the fourth)
 * From #45: How do you want pagination handled? (Although that might not be layout, which to me implies just a CSS change)

Perhaps this implies layout should be a multi-select option instead of a “choose one”, or perhaps this is really three separate features.

Throne3d:
It was intended to be a multi-select, yes. "No icon" and "Right-sided reply info boxes" should really not be mutually exclusive. The plan is to split the "No icon" type things into a layout multi-select, as distinct from the style single-select.

I don't think I was proposing to have character info above replies mostly because this would suggest that left/right overrides the current "above the reply" behavior seen on mobile.

Pagination display I was expecting to be a desktop/mobile divide, as opposed to a layout thing.

dhouck:
Okay, I thought the plan was “One option for layout, one option for style”.

I don’t think I’m parsing your second paragraph correctly. I would like the options page to suggest putting the infobox above the reply, because I would like to actually put the infobox above the reply even on non-mobile. I don’t know what you thought would suggest it would be overridden on mobile, or whether you wanted that suggested or not, but I’m not sure how such override would work given the limited horizontal screen real estate. I think the ideal would be, on the options page, having an option for “Infobox position (non-mobile view)” with options for left, right, and above. Perhaps there could be an asterisk next to “non-mobile view” saying that the mobile view always puts the infobox above the reply.

I only mentioned #45 because it mentioned it might want to be part of this.

Throne3d:
Oh. It would in my opinion suggest that it would override the display everywhere, if it had "Infobox above" as well as "Infobox to the right", and the problem is that yes this would be problematic on mobile devices. We could have that layout option separate from the others with such information.

[...]

PR #97, opened 2017-02-05
Add a paginator to the top of post lists

Throne3d:
* Add a paginator to the top of post-lists; make other style changes

* Fix styling with various themes and the new compact paginator

* Darken pagination visited links (on the default style)

Addresses part of #45, but should perhaps wait until layouts (#43) are properly implemented.


Throne3d:
If we wait for layouts to be implemented, the UI problems of this being "ugly" are removed (it can be a layout and hence be optional). Accordingly this PR is stalled on that.

Throne3d:
I'll close this for now – it hasn't been touched in forever. #43 can track it, and we can steal code from here if it's ever useful.

PR #99, opened 2017-02-05
Improve the dropdowns and JS of the reply editor

Throne3d:
In the post/reply editor:

* On page load, when restoring a form (from Firefox saving), update the `active_character` dropdown to display the proper character.  
(Further commit to scope the change event to `select2`.)

* Display no icon if switching to an avatarless user account.  
(Previously the icon would not disappear if switching to a user with no icon.)

* Make the icon dropdown consistent between JS and HAML  
  * Order icons in the dropdown by gallery if `icon_picker_grouping?` – matches the JSON behavior when swapping to characters with many galleries.  
  * Display `character.icon` in the dropdown if there are no galleries – matches the JSON behavior when swapping to a character with no galleries but a default icon.

* Use `character.icons` in a few places where the HAML was mapping galleries to icons, then flattening them and sorting them – this is for speed and memory purposes.

### DIFF
--- a/app/concerns/writable.rb
+++ b/app/concerns/writable.rb
@@ -16,7 +16,7 @@ module Writable
       return user.avatar_id? unless character
       return true if character.default_icon
       return false unless character.galleries.present?
-      return character.galleries.includes(:icons).map(&:icons).flatten.present?
+      return character.icons.exists?
     end
 
     def editable_by?(editor)
--- a/app/concerns/writable.rb
+++ b/app/concerns/writable.rb
@@ -16,7 +16,7 @@ module Writable
       return user.avatar_id? unless character
       return true if character.default_icon
       return false unless character.galleries.present?
-      return character.galleries.map(&:icons).flatten.present?
+      return character.galleries.includes(:icons).map(&:icons).flatten.present?
     end
 
     def editable_by?(editor)
--- a/app/controllers/writable_controller.rb
+++ b/app/controllers/writable_controller.rb
@@ -6,7 +6,7 @@ class WritableController < ApplicationController
   def build_template_groups
     return unless logged_in?
 
-    templates = current_user.templates.includes(:characters).order('LOWER(name)')
+    templates = current_user.templates.includes(:ordered_characters).order('LOWER(name)')
     faked = Struct.new(:name, :id, :ordered_characters)
     templateless = faked.new('Templateless', nil, current_user.characters.where(:template_id => nil).to_a.sort_by(&:name))
     @templates = templates + [templateless]
--- a/app/models/gallery.rb
+++ b/app/models/gallery.rb
@@ -1,8 +1,7 @@
 class Gallery < ActiveRecord::Base
   belongs_to :user
   belongs_to :cover_icon, class_name: Icon
-  has_and_belongs_to_many :icons, after_add: :set_has_gallery, after_remove: :unset_has_gallery
-  has_and_belongs_to_many :ordered_icons, after_add: :set_has_gallery, after_remove: :unset_has_gallery, order: 'LOWER(keyword)', class_name: Icon
+  has_and_belongs_to_many :icons, after_add: :set_has_gallery, after_remove: :unset_has_gallery, order: 'LOWER(keyword)'
 
   has_many :characters_galleries
   has_many :characters, through: :characters_galleries
--- a/app/models/gallery.rb
+++ b/app/models/gallery.rb
@@ -1,7 +1,8 @@
 class Gallery < ActiveRecord::Base
   belongs_to :user
-  belongs_to :cover_icon, :class_name => Icon
+  belongs_to :cover_icon, class_name: Icon
   has_and_belongs_to_many :icons, after_add: :set_has_gallery, after_remove: :unset_has_gallery
+  has_and_belongs_to_many :ordered_icons, after_add: :set_has_gallery, after_remove: :unset_has_gallery, order: 'LOWER(keyword)', class_name: Icon
 
   has_many :characters_galleries
   has_many :characters, through: :characters_galleries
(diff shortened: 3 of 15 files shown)


app/views/posts/_editor.haml
```diff
@@ -1,3 +1,4 @@
+- character_galleries = item.character.galleries.includes(:icons).ordered if item.character && current_user.icon_picker_grouping? && item.character.galleries.count > 1
```
Marri: I think you're going to double load icons here because the later line includes an 'order' which the `includes(:icons)` isn't using.

app/concerns/writable.rb
```diff
@@ -16,7 +16,7 @@ def has_icons?
       return user.avatar_id? unless character
       return true if character.default_icon
       return false unless character.galleries.present?
-      return character.galleries.map(&:icons).flatten.present?
+      return character.galleries.includes(:icons).map(&:icons).flatten.present?
```
Throne3d: Fixing an N+1 query – might be possible to do better but I haven't touched this code before.

app/controllers/writable_controller.rb
```diff
@@ -6,7 +6,7 @@ class WritableController < ApplicationController
   def build_template_groups
     return unless logged_in?
 
-    templates = current_user.templates.includes(:characters).order('LOWER(name)')
+    templates = current_user.templates.includes(:ordered_characters).order('LOWER(name)')
```
Throne3d: The characters dropdown uses `ordered_characters` so this should include that

app/models/template.rb
```diff
@@ -1,15 +1,12 @@
 class Template < ActiveRecord::Base
   belongs_to :user, inverse_of: :templates
   has_many :characters
+  has_many :ordered_characters, order: 'LOWER(name)', class_name: Character
```
Throne3d: Make `ordered_characters` into a scope so it can be `includes`-ed

app/models/gallery.rb
```diff
@@ -1,7 +1,8 @@
 class Gallery < ActiveRecord::Base
   belongs_to :user
-  belongs_to :cover_icon, :class_name => Icon
+  belongs_to :cover_icon, class_name: Icon
   has_and_belongs_to_many :icons, after_add: :set_has_gallery, after_remove: :unset_has_gallery
+  has_and_belongs_to_many :ordered_icons, after_add: :set_has_gallery, after_remove: :unset_has_gallery, order: 'LOWER(keyword)', class_name: Icon
```
Throne3d: Add an `ordered_icons` scoped relation so it can be `includes`-ed

app/presenters/character_presenter.rb
```diff
@@ -23,18 +23,18 @@ def as_json(options={})
   end
 
   def multi_gallery_json
-    galleries = character.galleries.ordered.includes(:icons)
+    galleries = character.galleries.ordered.includes(:ordered_icons)
     galleries_json = galleries.map do |gallery|
       {
         name: gallery.name,
-        icons: gallery.icons.order('LOWER(keyword)')
+        icons: gallery.ordered_icons
```
Throne3d: This was loading icons twice because it was ordering them – include the ordered relation and use that instead

app/presenters/character_presenter.rb
```diff
@@ -23,18 +23,18 @@ def as_json(options={})
...
       }
     end
   end
 
   def single_gallery_json
     if character.galleries.present?
-      [{icons: character.galleries.map(&:icons).flatten.uniq.sort_by {|i| i.keyword}}]
+      [{icons: character.icons}]
```
Throne3d: We have a neat relation for this exact thing, so update this to use it

app/views/posts/_editor.haml
```diff
@@ -1,3 +1,4 @@
+- character_galleries = item.character.galleries.includes(:ordered_icons).ordered if item.character && current_user.icon_picker_grouping? && item.character.galleries.count > 1
```
Throne3d: Provides `character_galleries` for if it's ever used, including `ordered_icons` since that's used in any code path that uses  `character_galleries`, it seems.

app/views/posts/_editor.haml
```diff
@@ -10,7 +11,14 @@
...
       - if item.character
-        = select_tag :icon_dropdown, options_for_select(item.character.icons.map{|i| [i.keyword, i.id]}, item.icon_id), prompt: "No Icon"
+        - if current_user.icon_picker_grouping? && item.character.galleries.count > 1
+          - icons = character_galleries.map(&:ordered_icons).flatten.map{|i| [i.keyword, i.id]}
+        - else
+          - icons = item.character.icons.map{|i| [i.keyword, i.id]}
+        - unless icons.present?
+          - icon = item.character.icon
+          - icons = [[icon.keyword, icon.id]] unless icon.nil?
+        = select_tag :icon_dropdown, options_for_select(icons, item.icon_id), prompt: "No Icon"
```
Throne3d: This is the same logic as `CharacterPresenter` uses for galleries (and then the same logic as the Javascript to produce HTML from it), so maybe they should be unified somehow.

app/concerns/writable.rb
Marri: `character.icons.exists?` is probably fine here

app/models/gallery.rb
Marri: I think this should be an actual scope, rather than a has_*? It feels weird to redefine this relationship.

app/models/template.rb
Marri: ibid

app/models/gallery.rb
Throne3d: I'm not sure how to do a relation scope. If you mean a scope on the icons, that means it can't be `includes`-ed, or – it can, but it'll end up being reloaded when the scope is used, I think?

PR #101, opened 2017-02-05
Add key traversal to the gallery#add_new page

Throne3d:
(Allows users to go between input boxes using input keys, for ease of entry.)


Throne3d:
I don't think this is stalled on anything except code review. (Unless we don't want this feature?)

Issue #49, opened 2017-02-04
Character pseudonyms


Throne3d:
Examples: Glam/Sadde, Theo?/Thea?

Also allow names to be concealed for certain replies (requested [here](http://alicorn.elcenia.com/board/viewtopic.php?f=12&t=459&p=28533#p28532)).

>I think a more general "let the lines displayed to identify a character be customisable" is ideal? At least the lines that currently have "name" and "screen name"

-Pedro

Throne3d:
The pseudonym should default to whatever pseudonym that character last used in that thread. It would be useful to have a dropdown of recent pseudonyms / all pseudonyms so far used by the particular character in the thread, so you can switch back and forth easily, as necessary, but this would require more UI.

Perhaps it should go in the character picker dropdown, and it should maybe also use the same mechanism for NPC posting (using the user account, but with a customizable 'character name' bar that doesn't link anywhere?).

Marri:
Making a new issue for the minor improvements remaining after #106 got merged

PR #106, opened 2017-02-10
Character aliasing

Marri:
Remaining:
- [x] Write tests
- [x] Add alias to API endpoint join
- [x] Include in Firefox refresh hack
- [x] Add a way to create, update and delete aliases
- [x] Rename stash and rebase it up
- [x] Default to most recent?
- [x] Display in profile somewhere
- [x] ~~Replace template_name with this intelligently?~~ Follow up later, this will be more work

Fixes issue #49


app/assets/javascripts/posts.js
```diff
@@ -111,9 +116,11 @@ $(document).ready(function() {
   var displayCharID = $("#post-editor .post-character").data('character-id');
   var selectedIconID = $("#reply_icon_id").val();
   var displayIconID = $("#current-icon").data('icon-id');
+  var selectedAliasID = $("#reply_character_alias_id").val();
+  var displayAliasID = $("#post-editor .post-character").data('alias-id');
   if (selectedCharID != displayCharID) {
-    getAndSetCharacterData(selectedCharID, {restore_icon: true});
-    $("#active_character").val(selectedCharID).trigger("chosen:updated");
```
Marri: We don't use chosen any more, so I took out the trigger

app/views/posts/_editor.haml
```diff
@@ -14,13 +14,18 @@
       - elsif current_user.avatar
         = select_tag :icon_dropdown, options_for_select([[current_user.avatar.keyword, current_user.avatar.id]], current_user.avatar.id), prompt: "No Icon"
     .post-info-text
-      .post-character{class: ('hidden' unless item.character), data: {character_id: item.character_id}}= item.character.try(:name)
+      .post-character{class: ('hidden' unless item.character), data: {character_id: item.character_id || '', alias_id: item.character_alias_id || ''}}
```
Marri: These must default to '' or the attribute can be left out of the element entirely, which might break the javascript?

Throne3d: Pretty sure `nil` in the data gets made into a blank string – I at least didn't have trouble at any point that I tested it with the previous setup – but I'm not 100% sure.

Marri:
It was not doing the thing at least locally for me until I used '' instead
of nil. Might be a browser thing? Meh, whatev, it's present now.

PR #108, opened 2017-02-13
API updates

Marri:
- Don't use `'data': {}` enveloping around JSON responses
- Update Javascript not to look for the `data` key when querying the API
- Use `api!` shorthand when documenting the API
- Move replies from `api/v1/posts/:id` to their own controller, since paginating a post on a child collection doesn't make sense

### DIFF
--- a/app/assets/javascripts/posts.js
+++ b/app/assets/javascripts/posts.js
@@ -235,7 +235,7 @@ bindIcon = function() {
 
 galleryString = function(gallery, multiGallery) {
   var iconsString = "";
-  var icons = gallery["icons"];
+  var icons = gallery.icons;
 
   for (var i=0; i<icons.length; i++) {
     iconsString += iconString(icons[i]);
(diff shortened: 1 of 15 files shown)


app/presenters/reply_presenter.rb
```diff
@@ -19,8 +19,8 @@ def as_json(options={})
   def character(reply)
     return unless reply.character_id
     { id: reply.character_id,
-      name: reply.character_name,
-      screenname: reply.character_screenname }
+      name: reply.name,
```
Marri: For historical record: these are overloaded in the Writable concern to properly use the value if it was preloaded from SQL, and to load the value through the associated object if it wasn't.

app/assets/javascripts/posts.js
```diff
@@ -362,7 +360,7 @@ getAndSetCharacterData = function(characterId, options) {
     if (restore_icon)
       setIconFromId(selectedIconID);
     else
-      setIcon(character['default']['id'], character['default']['url'], character['default']['keyword'], character['default']['keyword']);
+      setIcon(resp['default']['id'], resp['default']['url'], resp['default']['keyword'], resp['default']['keyword']);
```
MichaelBlume: OT, but in JS can't you just do like resp.default.url and so forth?

Marri: ...probably! I do not often Javascript xD

Issue #65, opened 2017-02-04
Make things case-insensitive as appropriate


Throne3d:
- ~~Screennames~~ (but not unique until aliases are implemented)
- Usernames (unique)
- ~~Facecasts~~
- Emails (unique)

Throne3d:
As aliases are now implemented, screennames should probably be made case insensitive and also unique assuming there aren't (m)any characters sharing screennames.

Throne3d:
Screennames would only be case insensitive for searching and/or uniqueness, so that's not required, and facecasts for sorting and grouping, which can be done differently.

PR #113, opened 2017-02-21
Update usernames, emails, facecasts and screennames to be citexts

Throne3d:
Fixes issue #65. The uniqueness requirements for emails and usernames is explicitly tested with items of different cases.


Throne3d:
Now only changes usernames and emails to be case insensitive.

PR #119, opened 2017-02-22
Add tests for GalleriesController's #icon

Throne3d:


app/controllers/galleries_controller.rb
```diff
@@ -101,8 +101,6 @@ def update
   end
 
   def icon
-    find_gallery if params[:id] != '0'
```
Throne3d: `setup_new_icons` is called before `#icon`, which includes this call, so this is redundant.

PR #130, opened 2017-03-06
Improve character#show UI, add character descriptions

Throne3d:
This is a rather large change, I'm afraid. No new tests because it's mostly (all?) UI and cucumber is scary and complicated. It might be worth checking out the branch and checking things work as expected.

- Extract the style of the new object pages into CSS; use appropriate classes (icons and users)
- Update the `characters#show` page to use the new interface
- Move the character template to the infobox, use `cycle` instead of manually setting even/odd classes
- Add a link to the character in the breadcrumbs
- Shrink the width (from 200px to 175px) of the character info box
- Add character descriptions to both `characters#edit` and `characters#show`

## Screenshots:
**Icons and users display as before**
image
image

**The new characters interface**
image
**Galleries**
image
**Posts**
image
**Editor** (it uses a textarea, does not use TinyMCE)
image
**New character page**
image
**Character with lots of missing details**
image
**Logged-out views**
image
image

Everything seems to work fine, I think?

### DIFF
--- a/app/views/posts/_list.haml
+++ b/app/views/posts/_list.haml
@@ -6,7 +6,7 @@
 
 %span.time-loaded= pretty_time(DateTime.now.in_time_zone)
 
-%table{style: local_assigns[:style].to_s, class: local_assigns[:table_class].to_s}
+%table{class: local_assigns[:table_class].to_s}
   %thead
     %tr
       %th{colspan: col_count}= content_for :posts_title
--- a/app/views/posts/_list.haml
+++ b/app/views/posts/_list.haml
@@ -6,7 +6,7 @@
 
 %span.time-loaded= pretty_time(DateTime.now.in_time_zone)
 
-%table{style: local_assigns[:style].to_s}
+%table{style: local_assigns[:style].to_s, class: local_assigns[:table_class].to_s}
   %thead
     %tr
       %th{colspan: col_count}= content_for :posts_title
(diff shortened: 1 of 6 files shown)


Throne3d:
Resolves part of #1.

Marri:
Too tired to go through this properly right now but a) I suggest the .vtop class to fix the weird centered Description labels and b) should we maybe put more thought into which if any of the fields display when empty and which don't?

That all aside it looks lovely from the screenshots and I appreciate you very much.

Throne3d:
<3

I wasn't sure which of Nickname(s), Facecast, Setting to hide – they're reasonably frequently blank but having them *all* be blank (and hiding them all) would be weird, and I'd probably need to display a "No extra info" bar or something? I wasn't sure how to go around it.

Throne3d:
image
Rebased against master, added `vtop` to the description header, added a style for the textarea on new character (the previous style only applied to edit character).

I'm still not sure which of Nickname(s), Facecast and Setting to hide when blank – if all of them, something should presumably be displayed when they're all blank, but again I'm not sure what. The present configuration might work okay? Or possibly only "Facecast" shows when blank, since that's very rarely going to be blank, whereas the others might be.

Throne3d:
I actually went ahead and hid Nickname(s) and Setting when blank, since that seems to make sense.

Again, character with no additional information (just name):
image

app/views/posts/_list.haml
```diff
@@ -6,7 +6,7 @@
 
 %span.time-loaded= pretty_time(DateTime.now.in_time_zone)
 
-%table{style: local_assigns[:style].to_s}
+%table{style: local_assigns[:style].to_s, class: local_assigns[:table_class].to_s}
```
Marri: You can take out the style one, grep says it's only icons and users using it. (Using `ag 'partial.*list.*style'`)

PR #133, opened 2017-03-08
Move the 'set avatar' text on users#edit to a row instead of a subheader

Throne3d:
The previous text was quite noticeable, which was… good-ish, but for those of us who have set an avatar and really don't care for the message, I think it makes more sense for it to be a small bit of text inline with the rest of the table.

### DIFF
--- a/app/views/users/edit.haml
+++ b/app/views/users/edit.haml
@@ -3,9 +3,6 @@
     %tr
       %th.centered{colspan: 2} Account Settings
     %tr
-      %td.odd.centered{colspan: 2}
-        .details *Please set your avatar directly from the desired icon's page*
-    %tr
       %th.sub Username
       %td{class: cycle('even', 'odd')}= f.text_field :username, placeholder: "Username", class: 'text', disabled: current_user.salt_uuid.nil?
     %tr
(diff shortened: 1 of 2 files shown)


app/views/users/edit.haml
```diff
@@ -17,6 +14,10 @@
         = f.check_box :email_notifications, class: 'width-15 vmid'
         = f.label :email_notifications, "Email about new tags"
     %tr
+      %th.sub Avatar
+      %td{class: cycle('even', 'odd')}
+        .details Please set your avatar directly from the desired icon's page.
```
Marri: Maybe we should also hide this if you've already set an avatar?

Throne3d: That's a good idea, actually. I'll go add it.

Throne3d:
I also changed the "Set as Default" text to "Make Avatar" on `icons#show` – it's still somewhat ambiguous but I think this is better than what it was previously?

app/views/users/edit.haml
```diff
@@ -16,6 +13,11 @@
       %td{class: cycle('even', 'odd')}
         = f.check_box :email_notifications, class: 'width-15 vmid'
         = f.label :email_notifications, "Email about new tags"
+    - unless current_user.avatar.present?
```
Marri: for informative-style reasons this is generally better done with a check on avatar_id so you don't load the object, but because the layout already loads the object I do not care in this specific case.

Issue #89, opened 2017-02-04
Make gallery from selected galleryless icons


Throne3d:

Throne3d:
Fixed in #134.

PR #134, opened 2017-03-08
Allow user to select icons to add to gallery on gallery#new

Throne3d:
Also has tests to ensure `has_gallery` is set appropriately. (Unfortunately with just the `has_many_and_belongs_to` relation, saving without a title but with icon IDs ends up setting them to `has_gallery` true, despite not having a gallery for them to be part of. The workaround in this prevents that being an issue.)

### DIFF
--- a/app/controllers/galleries_controller.rb
+++ b/app/controllers/galleries_controller.rb
@@ -199,7 +199,7 @@ class GalleriesController < ApplicationController
   end
 
   def gallery_params
-    params.fetch(:gallery, {}).permit(:name, galleries_icons_attributes: [:id, :_destroy, icon_attributes: [:url, :keyword, :credit, :id, :_destroy]])
+    params.fetch(:gallery, {}).permit(:name, galleries_icons_attributes: [:id, :_destroy, icon_attributes: [:url, :keyword, :credit, :id, :_destroy]], icon_ids: [])
   end
 
   def icon_params(paramset)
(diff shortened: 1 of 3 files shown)


Throne3d:
Fixes #89, I think.

Throne3d:
**With no galleryless icons**
image
**With galleryless icons**
image

(The width is fixed at 100%, as the default for formtables is 400px so this overrides it forcefully. Unfortunately, this means that if there's only one galleryless icon it might look weird on large screens. Screenshot below.)
image

Throne3d:
This no longer passes as `set_has_gallery` skips the icons when they're set in `new` (as the gallery has no name and hence isn't valid) and then doesn't get triggered when the gallery is saved.

I think it might be worth making the GalleriesIcons model or whatever and having an `after_create` callback there – `after_add` on the relation is called even when the model isn't saved, due to how Rails handles these callbacks (ie weirdly).

Throne3d:
Transactions seem to work well for this but the new test ensuring `set_has_gallery` only occurs when the gallery is valid fails (as that line has been removed here). I put a skip in for it but we should maybe revisit that later? (Transactions could be a better way of dealing with things when they have associations, though, to prevent this sort of problem occurring elsewhere.)

Throne3d:
I've rebased this with the fact we now use `set_has_gallery` on a separate model, as well as Rails 4 strong params, and it seems to still pass tests.

PR #135, opened 2017-03-08
Add a link to the reports page

Throne3d:
I was reminded we didn't have a link to it, so I've added one. It might be worth using "Activity Reports" instead of just "Reports" to make it clear it's not a "report something bad" kind of feature?


Throne3d:
Pedro raises a good point that the whole site is in beta, and rather than blocking this on the report page not being finished, we should put this out, get feedback (because people will actually use it) and improve it as necessary.

Issue #83, opened 2017-02-04
Indicate unread posts in tags owed


Throne3d:
> Via Moriwen:
> I would find it very helpful to have something on the Tags Owed page marking which threads I have unread posts in and which I don't. I often lose track of whether I haven't gotten around to replying to something, or whether I have and the other person has already replied back; this results in me sometimes not seeing a new reply for a long time.
> http://alicorn.elcenia.com/board/viewtopic.php?f=12&t=459&start=1630#p30737

Throne3d:
Bold could work.

Throne3d:
Changing the color of the unread icon would be better, apparently. Maybe we should make it slightly gray-er if there are no unread replies, so it's consistent with other pages when you have unreads but the tags owed page still has the appropriate button.

PR #137, opened 2017-03-11
Mark read posts in tags owed with a different arrow

Throne3d:
Fix for issue #83. Discussion about this issue appears [here](http://alicorn.elcenia.com/board/viewtopic.php?f=12&t=459&start=1840#p31858) and [here](http://alicorn.elcenia.com/board/viewtopic.php?f=12&t=459&start=1850#p31885), and I think it's an acceptable change.

## Screenshots
The first post (temp4) has been marked unread. The second post (temp3) has some unread replies. The third and fourth posts (temp2 and temp1) have been fully read.

**Starry light**
image

**Starry dark**
image


app/views/posts/_list_item.haml
```diff
@@ -11,9 +11,14 @@
     - if !current_user.try(:hide_warnings) && post.content_warnings.exists?
       = image_tag "/images/exclamation.png", class: 'vmid', title: "Content Warning: " + post.content_warnings.pluck(:name).join(', '), alt: '!'
   %td.post-subject{class: klass}
-    - if logged_in? && (@show_unread || (!@unread_ids.nil? && @unread_ids.include?(post.id)))
+    - if logged_in? && ((@opened_ids.nil? || !@opened_ids.include?(post.id)) || (!@unread_ids.nil? && @unread_ids.include?(post.id)))
```
Throne3d: Oh, oops – I think this line should use `(!@opened_ids.nil? && !@opened_ids.include?(post.id))`

app/views/posts/_list_item.haml
```diff
@@ -11,9 +11,14 @@
     - if !current_user.try(:hide_warnings) && post.content_warnings.exists?
       = image_tag "/images/exclamation.png", class: 'vmid', title: "Content Warning: " + post.content_warnings.pluck(:name).join(', '), alt: '!'
   %td.post-subject{class: klass}
-    - if logged_in? && (@show_unread || (!@unread_ids.nil? && @unread_ids.include?(post.id)))
+    - if logged_in? && ((!@opened_ids.nil? && !@opened_ids.include?(post.id)) || (!@unread_ids.nil? && @unread_ids.include?(post.id)))
```
Marri: This is pretty much unreadable but legibility can be fixed later as long as it works

Marri: (should probably clarify that these are my personal gripes with things like "line is too long" and "too many conditions" and "lots of parentheses", my existing style set no good precedent here, this is not your fault I'm just griping to gripe)

app/helpers/application_helper.rb
```diff
@@ -53,6 +53,13 @@ def unread_img
     '/images/bullet_go.png'
   end
 
+  def lastlink_img
```
Marri: I suspect we'll need a cucumber test for this but I can show you how to make those later

app/views/posts/_list_item.haml
Throne3d: I agree that it's mostly illegible (which is why I added the comment on the next line). I'm not sure what to do to fix that except to try to parse the bits manually or else somehow extract it into a method.

PR #141, opened 2017-03-13
Add a post paginator to posts#hidden

Throne3d:
Teceler couldn't find a post on her `posts#hidden` page earlier – turns out it was on another page, but the paginator wasn't displayed. This should fix that happening.

(The boards list does not seem to be paginated and so does not need a paginator.)


Throne3d:
You didn't actually merge this? (You approved it, apparently, but didn't merge.)

PR #142, opened 2017-03-15
Fix reply#preview then using replies_path

Throne3d:
And associated fixes.

Adelene reported an issue in which editing a reply, pressing preview, and then saving would create a new reply. I confirmed that it happened (every time), investigated the code, and didn't really understand why it was being done the way it was. This makes it use the inbuilt system to handle the paths instead of manually setting it (by retaining the ID on the replies/posts).

I tested it manually and everything seems to work (Reply/Post saving from #new, saving from #preview, previewing from #preview and then saving from there).

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -84,10 +84,9 @@ class PostsController < WritableController
   end
 
   def create
-    preview(:post, posts_path) and return if params[:button_preview].present?
-
     @post = Post.new(params[:post])
     @post.user = current_user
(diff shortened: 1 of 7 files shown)


app/controllers/posts_controller.rb
```diff
@@ -115,12 +114,10 @@ def history
   def stats
   end
 
-  def preview(method, path)
-    @written = Post.new(params[:post])
+  def preview(written)
+    @written = written
     @post = @written
```
Marri: I'll fix it but if we're going with @post in both cases we should simplify this

PR #150, opened 2017-03-22
Remove the maximum length from passwords

### LINKED ISSUE #136: Remove the maximum length on passwords
It's bad practice to put a max length on it, I'm pretty sure. I don't see why there should be a max limit.

Throne3d:
Fixes #136.


Throne3d:
(Added & fixed tests for password length.)

PR #152, opened 2017-03-24
Create 5 boards & users, then destroy them, before tests

Throne3d:
This is done so there is never a 'Site testing' board, as that screws with various other tests and meant I had to fix them in at least one other place. Here it's used because I was writing a Cucumber feature that failed specifically because the fourth post made will be in the fourth board made will be treated as site testing and hence does not show up in the "Unread" list, breaking the test.

I don't know if there's a neater way to do this – if we could mock the IDs when we *do* want to use site testing, and guarantee it'll never expect it to be a site testing board otherwise, that'd be much nicer, but I don't know how we'd go about doing that. Stubbing the constant in Board doesn't seem to work neatly.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -31,6 +31,8 @@ end
 require 'factory_girl_rails'
 require 'rails_helper'
 require 'support/spec_test_helper'
+$LOAD_PATH << '.'
+require 'features/support/fix_boards.rb'
 
 RSpec.configure do |config|
   # rspec-expectations config goes here. You can use an alternate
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -32,6 +32,14 @@
 require 'rails_helper'
 require 'support/spec_test_helper'
 
+# make 5 boards so site_testing doesn't screw up tests
+5.times do
```
Marri: I can think about cleaner ways to do this but uh regardless there shouldn't be two of these xD

Throne3d:
I extracted it into just `features/support/fix_boards.rb` now, and removed a previous hack I'd done to fix a test. (Run `rspec -e "PostsController GET index"` and on master, if `27.times` is made into `26.times`, it should fail – here it works.)

PR #154, opened 2017-03-24
Test boards#show (with a cucumber test) for flat boards

Throne3d:
Builds on #153 since it uses the `Then(/^I should see (\d+) posts?/)` step found there. Tests the display of a flat board (e.g. Sandboxes) with 5 posts.

Fails intermittently if 9195874 isn't applied due to Postgres arbitrary ordering.

### DIFF
--- a/features/boards/view_board.feature
+++ b/features/boards/view_board.feature
@@ -1,0 +1,10 @@
+Feature: View Board
+As a reader
+I want to be able to see a list of posts in a continuity
+So that I can read this glowfic
+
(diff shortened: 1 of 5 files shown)


Throne3d:
… Testing locally it looks like this error is appearing intermittently and I'm not sure why.

Throne3d:
PostgreSQL arbitrary ordering bites us again! (`User.first` in a cucumber test to mark it read for the current user: *did not work well*. I've set up a really basic `current_user` thing which hopefully gets around it.)

PR #158, opened 2017-03-27
Save alias also for characters that aren't the most recent one

Throne3d:
Builds on #157. It adds a new parameter to the character endpoint of the API (`post_id`), which when given will provide an `active_alias` attribute in the resulting JSON that can then be used in the `posts.js` file to update to it when it gets the character. (If no active alias is found, it acts as before.)

Manual testing showed the Javascript to work properly when replying to a post (various different characters, swapping from 'no alias' to 'has alias' and back, testing preview also), and creating a post still seems to work without modifying the alias.

I wasn't sure if modifying the API endpoint for characters was the best place to put this – perhaps could've created a subroute on the posts controller? I'd be happy to move it if you want.


Throne3d:
Looks like this is done in master (though differently) – I rebased this to make the diff clearer, since it might be worth copying some of the extra tests here over to master?

Throne3d:
Code can be looked at at some future point if we want to copy some tests across.

PR #160, opened 2017-03-28
Update .eslintrc and then fix a ton of JS lint things

Throne3d:
I'm afraid this is also pretty long, but it's more predominantly:
- fixed whitespace
- fixed semicolons
- `x = function(…) {` → `function x(…) {`
- removed braces around single-statement inline `if` blocks (mostly)
- comments that `gon` and `tinyMCE` exist
- ~~camelCasing things~~ (this was a bit large so I haven't actually done it here)
- removed unused parameters (`function a(b) { console.log(b); }` will work fine given `a(1)`, `a(1, 2)` and `a(1,2,3)` – i.e. all output `1`)
- rewrote a couple of places where some ugly string addition is done to produce HTML (as it can go horribly wrong if people try to abuse it)

Sadly we don't have tests on our JS or anything.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 **/*{.,-}min.js
+**/jquery.ui.widget.js
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,7 +12,6 @@ env:
 # http://eslint.org/docs/rules/
 rules:
   # Possible Errors
-  comma-dangle: [2, never]
   no-cond-assign: 2
   no-console: 0
   no-constant-condition: 2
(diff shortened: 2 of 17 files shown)


Throne3d:
I've split out the individual changes into individual commits so they'll hopefully be easier to review.

Throne3d:
Should wait on #175 and then be rebased (and fixed).

Throne3d:
Rebased onto master (hopefully successfully) to make it easier to rebase in the future…

Throne3d:
Wrt "Check function return values", returning `true` from a jQuery event handler is the same as returning nothing – only `false` is handled specially:

> Returning false from an event handler will automatically call event.stopPropagation() and event.preventDefault(). A false value can also be passed for the handler as a shorthand for function(){ return false; }. So, $( "a.disabled" ).on( "click", false ); attaches an event handler to all links with class "disabled" that prevents them from being followed when they are clicked and also stops the event from bubbling.

- https://api.jquery.com/on/

So the places where I've removed `true` from the `return` within an event handler don't affect its functionality.

PR #164, opened 2017-03-31
Set up a current_user for cucumber to avoid intermittent errors

Throne3d:
It would probably make sense for this to come before #152 and for that to build on this (since I expect this is less controversial) but I don't really want to rewrite the history of four PRs right now. Feel free to cherry-pick just 9195874.

### DIFF
--- a/features/step_definitions/navigation.rb
+++ b/features/step_definitions/navigation.rb
@@ -1,26 +1,30 @@
+def current_user
+  @user
+end
+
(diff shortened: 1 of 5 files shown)


features/step_definitions/navigation.rb
```diff
@@ -1,26 +1,30 @@
+def current_user
```
Marri: I'm not sure a global function like this is a great idea but we can worry about that later

PR #173, opened 2017-04-11
Reorganize post javascript into sections, skip bits as necessary

Throne3d:
The JavaScript was previously executing a call to `/api/v1/characters/undefined` (which produces a 404) due to how it was handling 'there is no editor' – i.e., try to do all the usual checks ("is the displayed character ID the same as the actual value for the ID", etc.), notice a disparity, try to fix the editor appropriately.

This PR reorganizes the JavaScript to put things in the following order:

* Post viewing setup: screenname resizing, scroll to unread
* Post metadata editing setup (also for coauthors): select2 controls for `board_id`, `section_id`, `privacy`, `viewer_ids`; the various tag multi-selects; the access list dropdown hide/show; the AJAX section-list-getting; the interactivity of the submit buttons
* Writable editing setup (i.e. characters & content & icons & so on): fix active-character dropdown, only setup tinyMCE if in RTF mode, fix UI for any cached values in form, icon dropdown responsiveness, RTF/HTML editor buttons, swap character & alias responsiveness, ability to exit hover-boxes with escape or clicking off them

I don't think the reordering breaks any dependencies (or at least none obviously stick out on manual testing, and it doesn't look like I've messed anything up).

The following checks are used to determine whether to continue or stop:

* After post viewing setup, look for a `$("#post-editor")`, else skip everything else
* After post metadata editing setup, look for a `$("#post-editor .view-button")` (i.e. an RTF/HTML button), else skip everything else

### DIFF
--- a/app/assets/javascripts/posts.js
+++ b/app/assets/javascripts/posts.js
@@ -25,6 +25,8 @@ $(document).ready(function() {
   if (shouldScrollToUnread)
     $(window).scrollTop(unreadElem.offset().top);
 
+  if ($("#post-editor").length === 0) return; // Skip if there is no post / writable editor
+
   // SET UP POST EDITOR:
 
   PRIVACY_ACCESS = 2; // TODO don't hardcode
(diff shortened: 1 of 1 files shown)


Throne3d:
For a less invasive change that hopefully produces the same result, only apply the fix-UI-according-to-cached-form-data action when there is definitely a `#post-editor` present.

I don't think this is hugely more invasive and really we should group together the various setups anyway.

PR #186, opened 2017-04-13
Fix unread timestamps changing when hiding a post

Throne3d:
Builds on #185. Removes `:read_at` from the timestamps to be set on create, so it will be `nil` if the person has never read the post (e.g. they hide it without opening it).

The change to the query may not be necessary, but I thought it best to hard-code it rather than assuming `date_trunc` on nil would produce nil and that `nil < date` would be true?

### DIFF
--- a/app/concerns/viewable.rb
+++ b/app/concerns/viewable.rb
@@ -9,12 +9,12 @@ module Viewable
       return true if view.ignored
 
       if view.new_record?
-        view.read_at = at_time
+        view.read_at = at_time || Time.now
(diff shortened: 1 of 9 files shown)


Throne3d:
The `@unread_ids` and `@opened_ids` in `posts_from_relation` would likely have been broken with this change, but I fixed that and added tests to specifically ensure they still worked. Marking a post unread no longer destroys its post view, as if a post is ignored and is then marked unread, this would mark it unignored.

PR #187, opened 2017-04-13
Mark posts read even if the board is ignored

Throne3d:
I'm not sure why it's special-cased so this isn't done – people would like the arrows to keep track of where they've read up to even if they hide the board for reasons such as 'it was cluttering that page and I didn't want to typically be reminded of this board'. (If there is no reason *against* this change, I'd say there's good reason to bring it through, even if it seems weird / unaesthetic.)

Pedro is one specific person who would like this behavior, I think.

### DIFF
--- a/app/concerns/viewable.rb
+++ b/app/concerns/viewable.rb
@@ -6,7 +6,6 @@ module Viewable
 
     def mark_read(user, at_time=nil, force=false)
       view = view_for(user)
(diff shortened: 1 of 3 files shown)


Throne3d:
… This does not actually work. I'll get back to it.

Throne3d:
This now builds on #186. And should work.

PR #188, opened 2017-04-13
Show hidden status of posts in post lists

Throne3d:
By implementing the same style as on the reports page, change the opacity of the rows for hidden posts to 30%.

I don't think people are going to be *strongly* adversely affected by this and it will improve functionality for a few people. If it's controversial enough, it can be rolled back, but when I polled people in IRC they seemed generally positive about it and the opposition on the forum seemed to drop their position.

### DIFF
--- a/app/views/posts/_list_item.haml
+++ b/app/views/posts/_list_item.haml
@@ -1,5 +1,6 @@
 %tr
-  - klass = cycle('even', 'odd')
+  - klass = [cycle('even', 'odd')]
+  - klass << 'post-ignored' if current_user && post.ignored_by?(current_user)
   - replies_count = post.reply_count
   %td.post-completed{class: [klass, (post.completed? ? 'post-complete' : 'post-incomplete')]}
     - if post.completed?


Throne3d:
This should (intentionally) only do it when a postview is marked ignored. If the post is in a board and the boardview is marked ignored, this is not designed to mark it ignored. (That would require further cases, such as making sure that if you view a board that you marked ignored it is not almost completely 30% opacity.)

app/views/posts/_list_item.haml
```diff
@@ -1,5 +1,6 @@
 %tr
-  - klass = cycle('even', 'odd')
+  - klass = [cycle('even', 'odd')]
```
Marri: I don't think this works since we're inserting the array into another array below; did it work when tested? We might need a .flatten somewhere.

Throne3d: It does seem to work as is:

image

Throne3d:
Closing, allowing tracking by #43.

Issue #184, opened 2017-04-13
Post search doesn't respect visibility


Throne3d:
As in the title, it shows the presence (title and such) of posts that are invisible to the person searching.

PR #189, opened 2017-04-13
Select search results by visibility

Throne3d:
Previously, people could search for posts that they couldn't see and search for replies that they also couldn't see.

This both:

- skips in the post list (so it skips even if there is more than one page or weird things occur)
- selects the whole list if there are ≤ 1 pages (so the 'count:' value doesn't show '1' when there are no replies, and so 'no search results' shows if applicable)

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -231,6 +231,9 @@ class PostsController < WritableController
       @search_results = @search_results.where(where)
     end
     @search_results = @search_results.paginate(page: page, per_page: 25)
+    if @search_results.total_pages <= 1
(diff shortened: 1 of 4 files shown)


Throne3d:
To reproduce, see [here](https://www.glowfic.com/posts/search?utf8=%E2%9C%93&board_id=&author_id=&character_id=&setting_id=&subject=darkness&commit=Search) ("Fluidity in the darkness" is hidden) and [here](https://www.glowfic.com/replies/search?utf8=%E2%9C%93&board_id=3&author_id=&template_id=&character_id=&subj_content=scrambles+moving&commit=Search) (same). Incognito mode could be used to see, if you happen to have permission to see these posts.

Throne3d:
Fixes #184.

Throne3d:
(The promised tests are in #201.)

PR #190, opened 2017-04-13
Improve unread highlighting

### LINKED ISSUE #179: Unread highlighting
Outline in red (like with Alicorn) per account option
Optionally include something that is control-F findable; invisible or tiny text?

Throne3d:
Fixes #179.

- Make the unread highlighting a toggleable account option

Displays as before:
image
Account option:
image

- Add an invisible text 'First unread marker' to mark unread reply

Looks the same as before:
image
Partially selected:
image
Fully selected:
image
With border:
image

I could make the text unselectable, but unfortunately either way it ends up appearing if you select the replies around it and copy and paste (i.e. the text will interrupt if people copy and paste; they shouldn't be doing this anyway but still :P). Making it unselectable still makes it show up when you hover over it (cursor changes to be the selection cursor) but it'd mean double-clicking to select wouldn't not show anything. Giving it a negative z-index (so it shows behind another element and you don't get the cursor) means it's not visible when people do Ctrl+F. I'm not sure how to resolve this, but it works pretty well at present.

The reason it's given a container with `position: relative` is so the `position: absolute` adjustments are made relative to its location; if I gave *it* `position: relative`, it wouldn't be out of the flow of the page (and accordingly it'd mess with local things like the post info box).

### DIFF
--- a/app/views/replies/_single.haml
+++ b/app/views/replies/_single.haml
@@ -1,14 +1,11 @@
 -# Hack for Alicorn's unread bug
-- highlighted = ''
-- if reply.id.present? && logged_in? && current_user.id == 2 && @unread.present? && @unread.id == reply.id && @unread.class == reply.class
-  - highlighted = 'border: 1px red solid;'
+- highlighted = 'border: 1px red solid;' if reply == @unread && current_user.try(:visible_unread?)
 
 .post-container{class: reply.is_a?(Post) ? 'post-post'.freeze : (reply.is_a?(Reply) ? 'post-reply'.freeze : ''), style: highlighted}
   - if reply.is_a?(Reply) && reply.id.present?
     <a id="reply-#{reply.id}" class="noheight"> </a>
-  - if reply.id.present? && logged_in? && !@unread.nil?
-    - if reply.id == @unread.id && reply.class == @unread.class
-      <a id="unread" class="noheight"> </a>
+  - if reply == @unread
+    <a id="unread" class="noheight"> </a>
   .padding-10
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/global.css
```diff
@@ -36,7 +36,6 @@ button, .button {
 
 /* Display Helpers */
 .no-margin { margin: 0; }
-.noheight { font-size: 0; }
```
Marri: Um? We use this?

app/views/replies/_single.haml
```diff
@@ -1,14 +1,12 @@
 -# Hack for Alicorn's unread bug
-- highlighted = ''
-- if reply.id.present? && logged_in? && current_user.id == 2 && @unread.present? && @unread.id == reply.id && @unread.class == reply.class
-  - highlighted = 'border: 1px red solid;'
+- highlighted = 'border: 1px red solid;' if reply == @unread && current_user.try(:visible_unread?)
```
Marri: We should pull this into a class and make it not a hack-with-hack-comment

app/assets/stylesheets/global.css
Throne3d: … oops. I didn't mean to remove this; I accidentally added a thing and removed the wrong thing.

Throne3d: Fixed.

app/views/replies/_single.haml
Throne3d: Also fixed.

Issue #140, opened 2017-03-12
Properly escape image URLs on upload


Throne3d:
With [this image](https://d1anwqy6ci9o1i.cloudfront.net/users/5/icons/4e2a38a5-7f57-475a-96b8-d3d47bd08cf7_kato%2029%2B.png), the `%2B` to the end of the URL had not been escaped (from `+`) by the server, and so the icon failed to display properly on the site. This should be fixed.

Throne3d:
According to [the S3 docs for POST object](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPOST.html#RESTObjectPOST-requests-responses-response-elements), there should be a response of `Location` as well as `Key`, which (being a URI instead of the key name), may be properly escaped and fully functional as a URI.

I just uploaded a file to the site, put a breakpoint in the JavaScript, and found the following response:
```
Location: https://glowfic-constellation.s3.amazonaws.com/users%2F31%2Ficons%2F43f7f1d3-6933-4473-8221-e3f57d067853_dupe-avatar.jpg
Bucket: glowfic-constellation
Key: users/31/icons/43f7f1d3-6933-4473-8221-e3f57d067853_dupe-avatar.jpg
Etag: "722c62255530edb560f2b608c0752c8b"
```

Looking at the "Location", as given, [it does appear to be functional](https://glowfic-constellation.s3.amazonaws.com/users%2F31%2Ficons%2F43f7f1d3-6933-4473-8221-e3f57d067853_dupe-avatar.jpg) – despite the slashes and all the other things being escaped – so I think  it would be good to use that, especially as it seems to be pre-escaped for us. If the current system is used so Cloudfront can cache it, simply use this response and replace as appropriate.

The same URL is functional with the host replaced with the Cloudfront equivalent: [see here](https://d1anwqy6ci9o1i.cloudfront.net/users%2F31%2Ficons%2F43f7f1d3-6933-4473-8221-e3f57d067853_dupe-avatar.jpg).

Throne3d:
I'm attempting some broken characters and will list if they worked below. These are listed [on a page about S3 object metadata](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html) as being problematic or potentially problematic. They may well not be escaped, in the tests prior to the relevant PR, but whether they work or break the image is listed.

The filename is `100 hilarity*broken.png`, with the `*` replaced by the relevant special character.

- `\` does not upload
- `?` uploads but does not display
[...]

PR #192, opened 2017-04-13
Use the Location response in S3 uploads to properly escape URIs

Throne3d:
As per #140 (see [comment about this](https://github.com/Marri/glowfic/issues/140#issuecomment-294027307)), use the `Location` response to get the image's URI instead of generating it from the `Key` value, so it escapes things properly. (Replace `https://glowfic-constellation.s3.amazonaws.com` with `https://d1anwqy6ci9o1i.cloudfront.net` so it still uses Cloudfront, though.)

I would appreciate it if you could test this, as I don't think I have a way to do so. Specific things you'd be looking for:
- It uses the Cloudfront URL instead of the S3 URL
- It actually works with any icons (and it doesn't just mysteriously fail)
- It works with an icon where the filename of the icon includes a `+` (which I think fails at present – the icon won't display on the site, as the URL will be broken)

**This is mostly untested.**

### DIFF
--- a/app/assets/javascripts/galleries/add_new.js
+++ b/app/assets/javascripts/galleries/add_new.js
@@ -86,8 +86,8 @@ function bindFileInput(fileInput) {
       submitButton.prop('disabled', false);
 
       // extract key and generate URL from response
-      var key = $(data.jqXHR.responseXML).find("Key").text();
-      var url = 'https://d1anwqy6ci9o1i.cloudfront.net/' + key;
+      // replace https://glowfic-constellation.s3.amazonaws.com with Cloudfront URL
+      var url = $(data.jqXHR.responseXML).find("Location").text().replace(form.data('url'), 'https://d1anwqy6ci9o1i.cloudfront.net');
 
       // create hidden field
       var iconIndex = addNewRow();


app/assets/javascripts/galleries/add_new.js
```diff
@@ -86,8 +86,9 @@ function bindFileInput(fileInput) {
         submitButton.prop('disabled', false);
 
         // extract key and generate URL from response
+        // replace https://glowfic-constellation.s3.amazonaws.com with Cloudfront URL
         var key   = $(data.jqXHR.responseXML).find("Key").text();
-        var url   = 'https://d1anwqy6ci9o1i.cloudfront.net/' + key;
+        var url   = $(data.jqXHR.responseXML).find("Location").text().replace(form.data('url'), 'https://d1anwqy6ci9o1i.cloudfront.net');
```
Marri: Tested locally and got `https://d1anwqy6ci9o1i.cloudfront.net/users%2F1%2Ficons%2F78f49a9a-c161-4395-9ea5-f8ab4df7238e_aestrix.jpeg` instead of `https://d1anwqy6ci9o1i.cloudfront.net/users/1/icons/78f49a9a-c161-4395-9ea5-f8ab4df7238e_aestrix.jpeg`, seems to be over-escaping?

Throne3d: I also got this (if you try the link I gave) - it seems to *work* just fine and so I think that might be what it's supposed to be, just it works if you don't escape those characters.

Throne3d:
Make sure icon deletion works if this gets merged, since it seems to rely on the URL for the key and this escapes the key.

Throne3d:
Will hopefully be superseded by the "[key](https://github.com/Marri/glowfic/tree/key)" branch, where the key value is being separated from the image location.

PR #194, opened 2017-04-13
Add character duplication, fixes #18

### LINKED ISSUE #18: Ability to fork/duplicate characters
> From Moriwen: "I want to fork a new instance off of it, with the same iconset, and make all posts by that character in [these selected threads] be by the new character instead, keeping their old icons; put the new character as a new instance of the same template."

As replacing all instances of a character in a post is already a feature, this can be solved by adding a feature to duplicate a character.

Throne3d:
(Forks can be done by duplicating and replacing as appropriate. Fixes #18.)

Aliases, being tied to a single character each, are duplicated,
added to the character's list of aliases, and saved one at a
time.

The main problem I can see with this approach is that if a character alias fails to save for some reason, or the new duplicate fails to save, it will fail silently. This *shouldn't* happen (character aliases only need a name and a character, which they will do; the duplicate should have the same data as the first character).

On the bright side, if it *does* happen, and the duplicate fails to save, they should get an error that it can't find the character (since the character isn't persisted, and so the redirect to `character_path(@dup)` will fail).


app/controllers/characters_controller.rb
```diff
@@ -69,6 +69,19 @@ def update
     end
   end
 
+  def duplicate
+    @dup = @character.dup
```
Marri: Transactions!

app/controllers/characters_controller.rb
```diff
@@ -69,6 +69,19 @@ def update
...
+    @dup.galleries = @character.galleries
+    @character.aliases.find_each do |calias|
+      dupalias = calias.dup
+      @dup.aliases << dupalias
+      dupalias.save
+    end
+    @dup.save
+    flash[:success] = "Character duplicated successfully."
+    redirect_to character_path(@dup)
```
Marri: Open to debate, but do we want to go straight to edit?

Throne3d: It might make it clearer? I'll add it.

app/controllers/characters_controller.rb
Throne3d: How would I catch a transaction failing? (So I can redirect them with an error.) If that's not the point of a transaction… what is? (If an error occurs, it'll bubble up and give them a 500, and then it'll also end up rolling back all the changes?)

app/controllers/characters_controller.rb
Throne3d: (Done.)

app/controllers/characters_controller.rb
Throne3d: (I've wrapped it in `Character.transaction` anyway but it doesn't detect if there's a problem. Similarly to how `characters#do_replace` doesn't seem to.)

spec/spec_helper.rb
```diff
@@ -131,6 +131,8 @@
   end
 end
 
+RSpec::Matchers.define_negated_matcher :not_change, :change
```
Marri: Did `.not_to change{}` not work for some reason? It's not really more grammatical this way or anything?

app/controllers/characters_controller.rb
```diff
@@ -69,6 +69,22 @@ def update
...
+      @character.aliases.find_each do |calias|
+        dupalias = calias.dup
+        @dup.aliases << dupalias
+        dupalias.save
+      end
+      @dup.save
+    end
+
+    flash[:success] = "Character duplicated successfully."
+    redirect_to edit_character_path(@dup)
```
Marri: You do not actually need to make this a @variable, standard local variable is fine, but not worth changing

spec/spec_helper.rb
Throne3d: "not to change" chained with "and change" is ambiguous (is it supposed to not change, the second time, or to change?) and rspec throws an error telling me to define a negative matcher to make it more explicit.

app/controllers/characters_controller.rb
Throne3d: I think it's required if I want to then reference it in tests? (unless I get all characters before and after the duplication, find the one new character)

app/controllers/characters_controller.rb
Marri: Standard is using save! instead of save or create!/create or similar equivalents

Throne3d: Should I then rescue it and display a nice message? (If so, what's the error type?)

Throne3d:
Updated to work with the new character UI (see 'Duplicate Character'):
image

Possibly still needs to use `save!` inside the transaction so if an error occurs it throws. (This shouldn't happen because duplicates of existing characters should be valid? But may still be useful to implement, and I'd appreciate guidance on how / for you to do it. xD)

PR #216, opened 2017-04-29
Fix the default icon switching of characters

Throne3d:
- Use `data-id` instead of `id` to keep track of icon IDs
- Force icons to be 100x100 even when deselected (for consistency with forced size once selected)
- Clean up the javascript (remove unnecessary padding, put spaces before parentheses surrounding conditions, remove unused `new_gallery`)
- Fix showing galleryless icons after removing all galleries from a character
- Make the HTML using `$("<tag>").attr(...attrs).append(contents)` instead of `<tag ...attrs>contents</tag>` so as to reduce chance of bogus input
- Only bind icons once, to prevent glitches where the icon picker refuses to select the icon you provided
  (after removing/adding a gallery, all other galleries will have the event bound twice, and would trigger both a select and deselect on one click)

### DIFF
--- a/app/assets/stylesheets/galleries.css
+++ b/app/assets/stylesheets/galleries.css
@@ -1,5 +1,6 @@
 .gallery-box { cursor: pointer; width: 20px; text-align: center; } /* The + / - symbol to minimize a gallery */
 .gallery-title { height: 5px; } /* So that when a gallery is minimized there is spacing */
+.character-icon { width: 100px; height: 100px; }
 .selected-icon { border: 10px #3F7055 solid; width: 80px; height: 80px; }
 .form-table .icon { vertical-align: middle; }
 
@@ -11,7 +12,7 @@
     display: inline-block;
     text-align: center;
     line-height: 14px;
-    margin-bottom: 10px; 
+    margin-bottom: 10px;
     position: relative;
     width: 130px;
     word-wrap: break-word;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/galleries.css
```diff
@@ -1,5 +1,6 @@
 .gallery-box { cursor: pointer; width: 20px; text-align: center; } /* The + / - symbol to minimize a gallery */
 .gallery-title { height: 5px; } /* So that when a gallery is minimized there is spacing */
+.character-icon { width: 100px; height: 100px; }
```
Marri: We do have rectangular icons, is there a way to get default not forcing instead of globally forcing?

Throne3d: When they're selected they have fixed square sizes so I was doing this for consistency.

Relevant code:

```css
.selected-icon { border: 10px #3F7055 solid; width: 80px; height: 80px; }
```

I could put it in a box and have it with a maximum size, and center it, so smaller images (including non-square images) are the appropriate size, and larger images are scaled appropriately? This would preserve the grid structure.

Throne3d:
Uh. Note to self: check other locations this icon picking might be done? Specifically `gallery#add_existing`.

app/assets/stylesheets/galleries.css
Throne3d: (`character-icon` is only used on `character#_editor` and `character#show` so this shouldn't affect elsewhere as is.)

PR #217, opened 2017-04-30
Only show icon and character if relevant attachments exist

Throne3d:
Due to the way we handle retrieving reply and post associations, if the ID for the association exists but the relevant association no longer does (eg an icon is deleted), the code would previously expect it to exist and hence display an icon.

This ensures the actual attached data is loaded (eg icon URL, character name) and if not then it assumes the association doesn't exist and so doesn't try to show that data.

### DIFF
--- a/app/models/character.rb
+++ b/app/models/character.rb
@@ -22,6 +22,8 @@ class Character < ActiveRecord::Base
 
   attr_accessor :new_template_name, :group_name
 
+  after_destroy :clear_char_ids
+
   nilify_blanks
 
   def recent_posts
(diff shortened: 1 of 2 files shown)


Throne3d:
This also now clears user avatars that reference the ID.

Throne3d:
And clears character IDs when characters are destroyed. (This caused issues including breaking the reply editor if the latest reply in a post used to use the character you just deleted.)

app/views/replies/_single.haml
```diff
@@ -11,12 +11,12 @@
       <a id="unread" class="noheight"> </a>
   .padding-10
     .post-info-box
-      - if reply.icon_id
+      - if reply.icon_id && reply.url
```
Marri: Let's take these out; I'll just clean the database once we release.

Throne3d: Done

Issue #219, opened 2017-05-01
Improve the quick switcher


Throne3d:
- [ ] Let people select the account used for the first button of the quick switcher (suggested by Kappa)
- [ ] Option to force it to show the first two letters of a name instead of the default icon (maybe?)
- [X] Option to increase the number of characters shown (5 or 10, maybe more?)
- [ ] Toggleable? (esp for mobile users)
- [X] Global author character is toggleable; display author character pinned if [not disabled] (or [used in thread]?)

Throne3d:
Seems like these mostly aren't high priority or haven't shown up again since, so I think we're good to close this

PR #220, opened 2017-05-01
Allow users to toggle the pinned user character in the quick switcher

Throne3d:
Some people don't use their user accounts ever, and would prefer that the user account accordingly not display in this list.

Fixes part of #219.

With user checkbox toggled on (default):
image

With user checkbox toggled off:
image


Throne3d:
TODO: Needs param to be settable in user controller.

PR #223, opened 2017-05-05
Set character default icons to default by gallery if nil

Throne3d:
- Migrate characters' default icons to gallery default if nil (reduces database queries)
- Remove the extra logic necessary to use gallery default if a character's default icon is nil (allows people to have characters with galleries and no default icon)

I think I proposed this before and I'm not sure why it wasn't merged then. Maybe it didn't have the seeds updated?


Throne3d:
For posterity, to update the seeds I executed the migration once and had it save times it changed the icon (in a map `character_name => new_default_id`) and then used the following to update the array of characters in the `db/seeds.rb` file:

```ruby
name_id_map = {…} # as per output from that migration
file = … # the (string of the) list of character hashes from db.seeds
regex = /:name => "((\\"|[^"])*)"/
newdat = file.read.split("\n").map do |line|
  match = line.match(regex)
  abort("couldn't find a name on line: #{line}") unless match
  name = match.to_s.sub(":name => \"", '').reverse.sub('"','').reverse.gsub("\\\"", "\"")
  id = name_id_map[name]
  (puts "no match for #{name}"; next line) unless id
  line.sub(":default_icon_id => nil", ":default_icon_id => #{id}")
end.join("\n")
```

After doing this I then found all characters with duplicate names (where the hash keys wouldn't have been unique) and conveniently they all had screennames; I mapped specifically these by `screenname => new_default_id` and then amended the rest of the list manually.

(The list of screennames for duplicate-name characters to IDs: )
```ruby
{"colorshapes"=>409, "more_capable"=>2949, "brickswithoutclay"=>1961, "calendarofcrime"=>1992, "if_inconvenient"=>2592, "wearsthecoats"=>3869, "whatyoumayexpect"=>3875, "pina_colada"=>1250, "hesperide"=>2544, "bitofafiction"=>1920, "causas_rerum"=>2050, "pyth"=>3381, "pythbox"=>3388, "most_capable"=>2989, "missnicegirl"=>2941, "withmypowers"=>3932, "prima_sequentem"=>3331, "summa_iniuria"=>3563, "unmirrored"=>3832, "unmired"=>3819, "unmarred"=>3813, "point_b"=>1290, "shesepankh"=>1427}
```

This is not necessary when performing on an active database, just to manually update the seeds file.

Throne3d:
There are currently 508 characters with no default icon on the Constellation, of which 352 have galleries; this migration should not take very long to run. (Loading 508 characters with their galleries, checking for presence of galleries, loading an icon for each that *does*, and then updating the character. It took 2.6 seconds to apply locally with ~250 characters that needed updating.)

app/models/character.rb
```diff
@@ -26,7 +26,7 @@ class Character < ActiveRecord::Base
   nilify_blanks
 
   def icon
-    @icon ||= default_icon || galleries.detect(&:default_icon).try(:default_icon)
```
Marri: We should just change uses of character.icon to character.default_icon but I'll merge for now.

Throne3d: Done in #224.

PR #230, opened 2017-05-12
Refactor the unread warning

Marri:
Uses an id instead of a boolean so it'll show the warning as many times as necessary; shows the unread posts themselves; provides both a new-window and same-window jump URL option (per lintamande's request)

### DIFF
--- a/app/controllers/replies_controller.rb
+++ b/app/controllers/replies_controller.rb
@@ -109,7 +109,8 @@ class RepliesController < WritableController
       if most_recent_unseen_reply.present?
         reply.post.mark_read(current_user, reply.post.read_time_for(@unseen_replies))
         num = @unseen_replies.count
-        flash[:error] = "There have been #{num} new #{'reply'.pluralize(num)} since you last viewed this post."
+        pluraled = num > 1 ? "have been #{num} new replies" : "has been 1 new reply"
+        flash[:error] = "There #{pluraled} since you last viewed this post."
         @last_seen_id = most_recent_unseen_reply.id
         preview(reply) and return
       end
--- a/app/controllers/replies_controller.rb
+++ b/app/controllers/replies_controller.rb
@@ -101,11 +101,16 @@ class RepliesController < WritableController
     reply = Reply.new(params[:reply])
     reply.user = current_user
 
-    if reply.post.try(:first_unread_for, current_user) && !params[:unread_warned]
-      # There are unreads and the user has not been warned.
-      params[:unread_warned] = true
-      flash[:error] = "There are unread replies not shown here. (Click 'post' again if you wish to ignore this warning.)"
-      preview(reply) and return
+    if reply.post.present?
+      last_seen_reply_id = params[:last_seen]
+      @unseen_replies = reply.post.replies.order('id asc').paginate(page: 1, per_page: 5)
+      @unseen_replies = @unseen_replies.where('id > ?', last_seen_reply_id) if last_seen_reply_id.present?
+      most_recent_unseen_reply = @unseen_replies.last
+      if most_recent_unseen_reply.present?
+        flash[:error] = "There have been #{@unseen_replies.count} new replies since you last viewed this post."
+        @last_seen_id = most_recent_unseen_reply.id
+        preview(reply) and return
+      end
     end
 
     if reply.save
--- a/app/controllers/replies_controller.rb
+++ b/app/controllers/replies_controller.rb
@@ -103,11 +103,13 @@ class RepliesController < WritableController
 
     if reply.post.present?
       last_seen_reply_id = params[:last_seen]
-      @unseen_replies = reply.post.replies.order('id asc').paginate(page: 1, per_page: 5)
+      @unseen_replies = reply.post.replies.order('id asc').paginate(page: 1, per_page: 10)
       @unseen_replies = @unseen_replies.where('id > ?', last_seen_reply_id) if last_seen_reply_id.present?
       most_recent_unseen_reply = @unseen_replies.last
       if most_recent_unseen_reply.present?
-        flash[:error] = "There have been #{@unseen_replies.count} new replies since you last viewed this post."
+        reply.post.mark_read(current_user, reply.post.read_time_for(@unseen_replies))
+        num = @unseen_replies.count
+        flash[:error] = "There have been #{num} new #{'reply'.pluralize(num)} since you last viewed this post."
         @last_seen_id = most_recent_unseen_reply.id
         preview(reply) and return
       end
--- a/app/helpers/writable_helper.rb
+++ b/app/helpers/writable_helper.rb
@@ -1,14 +1,13 @@
 module WritableHelper
   def unread_warning
-    return @unread_warning unless @unread_warning.nil?
-    # Get unread_reply separately from @post.first_unread_for because that caches (and sometimes does "halfway down the page you just loaded" when we want 'first unread after the whole page has loaded')
-    viewed_at = @post.last_read(current_user) || @post.board.last_read(current_user)
-    unread_reply = if viewed_at
-      @post.replies.where('created_at > ?', viewed_at).order('id asc').first
-    else
-      @post
-    end
-    return (@unread_warning = false) unless unread_reply
-    @unread_warning = content_tag :a, 'Post has unread replies', href: post_or_reply_link(unread_reply), class: 'unread-warning', target: '_blank'
+    return unless @replies.present?
+    return if @replies.total_pages == page
+    'Post has unread replies ' + \
+    content_tag(:a, '(View)', href: unread_path(@post, warn_id: @last_seen_id.to_i), class: 'unread-warning') + ' ' + \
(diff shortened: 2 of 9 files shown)


app/models/post.rb
```diff
@@ -111,6 +111,15 @@ def first_unread_for(user)
     @first_unread ||= reply
   end
 
+  def last_seen_reply_for(user)
+    return @last_seen if @last_seen
+    return unless replies.exists? # unlike first_unread_for we don't care about the post
+    viewed_at = last_read(user) || board.last_read(user)
+    return unless viewed_at
+    reply = replies.where('created_at <= ?', viewed_at).order('id desc').first
+    @last_seen ||= reply
```
Throne3d: It doesn't affect anything but I don't think this needs `||=`? (`=` should suffice, since you already checked for `@last_seen` existing.)

spec/controllers/replies_controller_spec.rb
```diff
@@ -70,12 +70,32 @@
...
+      create(:reply, post: reply_post)
 
-      post :create, reply: {post_id: reply_post.id, user_id: user_id}
+      post :create, reply: {post_id: reply_post.id, user_id: reply_post.user_id}
       expect(response.status).to eq(200)
-      expect(flash[:error]).to eq("There are unread replies not shown here. (Click 'post' again if you wish to ignore this warning.)")
+      expect(flash[:error]).to eq("There have been 1 new replies since you last viewed this post.")
+    end
+
+    it "handles multiple creations with " do
```
Throne3d: multiple creations with…?

app/views/posts/_write_reply.haml
```diff
@@ -4,9 +4,7 @@
     .view-button#rtf{class: ('selected' if (params[:editor_mode] || current_user.default_editor) == 'rtf')} Rich Text
     .view-button#html{class: ('selected' if (params[:editor_mode] || current_user.default_editor) == 'html')} HTML
     %br
-    - if unread_warning
-      = unread_warning
-      = hidden_field_tag :unread_warned, true
+    = hidden_field_tag :last_seen, @last_seen_id
```
Throne3d: This no longer warns the user before submitting the reply that there are unread replies. (If someone spends a while composing a tag only to find they failed to notice there was a new page in the thread, and that they were halfway through their coauthor's series of tags, that's a problem that would previously have been solved by the warning.)

app/views/replies/preview.haml
```diff
@@ -1,3 +1,17 @@
+- if @unseen_replies.present?
+  .post-header
+    .padding-15
+      Unseen Replies
+      = link_to post_path(@post, page: 'unread', anchor: 'unread'), target: '_blank' do
+        .view-button Unread (new tab) &raquo;
+      = link_to post_path(@post, page: 'unread', anchor: 'unread') do
+        .view-button.float-left Unread &raquo;
```
Throne3d: I don't think two buttons are necessary here; UI usually provides either a text link (same tab) with an associated "new window" glyph (new tab), or it just sticks to one or the other. It doesn't really seem useful to have both, and I'd stick to just the `target: '_blank'` one.

spec/controllers/replies_controller_spec.rb
```diff
@@ -70,12 +70,32 @@
...
       reply_post = create(:post)
+      login_as(reply_post.user)
+      reply_post.mark_read(reply_post.user)
+      create(:reply, post: reply_post)
 
-      post :create, reply: {post_id: reply_post.id, user_id: user_id}
+      post :create, reply: {post_id: reply_post.id, user_id: reply_post.user_id}
       expect(response.status).to eq(200)
-      expect(flash[:error]).to eq("There are unread replies not shown here. (Click 'post' again if you wish to ignore this warning.)")
+      expect(flash[:error]).to eq("There have been 1 new replies since you last viewed this post.")
```
Throne3d: I'd want to pluralize this properly in the controller – "There has been 1 new reply"?

app/views/replies/preview.haml
```diff
@@ -1,3 +1,17 @@
...
+
+  = render partial: 'replies/single', collection: @unseen_replies, as: :reply
+  - if @unseen_replies.total_pages > 1
+    .post-ender ... and #{@unseen_replies.total_entries - @unseen_replies.per_page} more ...
```
Throne3d: It would help if this linked to the rest of the replies. (Yes, there are the above button(s), as well as the permalink buttons, but when there are 5 replies on the page, the top button gets scrolled out of view, so it's less convenient.)

Throne3d: Also, if there are more replies and you press 'Post' again, it just reloads the page to show the rest of them. This is reasonable behavior, I guess, but could be inconvenient if people would like to post despite the other replies.

(It will also interact weirdly with people who have hidden boards from unread and are trying to still reply in them – nothing gets marked read, they can't read further in the thread, so if they respond from the last page it'll work fine but if they respond from a previous page I expect it'll go through all the ""unread"" replies. Could be fixed by pushing PRs #186 and #187.)

app/controllers/replies_controller.rb
```diff
@@ -101,11 +101,16 @@ def create
...
-      flash[:error] = "There are unread replies not shown here. (Click 'post' again if you wish to ignore this warning.)"
-      preview(reply) and return
+    if reply.post.present?
+      last_seen_reply_id = params[:last_seen]
+      @unseen_replies = reply.post.replies.order('id asc').paginate(page: 1, per_page: 5)
+      @unseen_replies = @unseen_replies.where('id > ?', last_seen_reply_id) if last_seen_reply_id.present?
+      most_recent_unseen_reply = @unseen_replies.last
+      if most_recent_unseen_reply.present?
+        flash[:error] = "There have been #{@unseen_replies.count} new replies since you last viewed this post."
+        @last_seen_id = most_recent_unseen_reply.id
```
Throne3d: As `last_seen_id` increases when you resubmit the form but it doesn't actually mark the replies *read*, resubmitting multiple times (and accordingly reading the unread replies in 5-per-submission increments) will not impact whether you've 'read' those replies, as far as the site is concerned. This seems weird. (Maybe just mark the post read up to the new `@last_seen_id`, to fix this?)

spec/controllers/replies_controller_spec.rb
Throne3d: Resolved

app/controllers/replies_controller.rb
Throne3d: Resolved

app/models/post.rb
Throne3d: Resolved

app/helpers/writable_helper.rb
```diff
@@ -1,14 +1,13 @@
...
-    else
-      @post
-    end
-    return (@unread_warning = false) unless unread_reply
-    @unread_warning = content_tag :a, 'Post has unread replies', href: post_or_reply_link(unread_reply), class: 'unread-warning', target: '_blank'
+    return unless @replies.present?
+    return if @replies.total_pages == page
+    'Post has unread replies ' + \
+    content_tag(:a, '(View)', href: unread_path(@post, warn_id: @last_seen_id.to_i), class: 'unread-warning') + ' ' + \
+    content_tag(:a, '(New Tab)', href: unread_path(@post), class: 'unread-warning', target: '_blank')
```
Throne3d: Won't this say "Post has unread replies" even if the user has seen the last reply? (Perhaps it should say "You are not on the last page" or something equivalent?)

app/views/replies/preview.haml
Throne3d: Resolved

spec/controllers/replies_controller_spec.rb
```diff
@@ -77,18 +77,18 @@
 
       post :create, reply: {post_id: reply_post.id, user_id: reply_post.user_id}
       expect(response.status).to eq(200)
-      expect(flash[:error]).to eq("There have been 1 new replies since you last viewed this post.")
+      expect(flash[:error]).to eq("There have been 1 new reply since you last viewed this post.")
```
Throne3d: "have been" isn't pluralized properly?

spec/controllers/replies_controller_spec.rb
Throne3d: Resolved

app/views/posts/_write_reply.haml
Throne3d: Resolved

Throne3d:
Also I'd suggest renaming the test "requires post read if no unread_warned param" to "requires post read if no last_seen param"?

spec/controllers/replies_controller_spec.rb
Throne3d: Resolved

app/helpers/writable_helper.rb
Throne3d: Resolved

spec/controllers/replies_controller_spec.rb
Throne3d: Resolved

app/helpers/writable_helper.rb
Throne3d: Resolved

PR #232, opened 2017-05-13
Upgrade Ruby to 2.4.1

Throne3d:
Using 2.3.3 means I can't compile Ruby on Arch Linux anymore :cry: (because it uses an old OpenSSL version or something and so the library throws a complaint). Also it's generally good to be up to date.


Throne3d:
… Hmm, it looks like this might actually require Rails 4.

Throne3d:
Closing for now.

PR #233, opened 2017-05-13
Improve reply unread warning by using read status instead of form data

Throne3d:
I was going to fix an issue whereby going through preview meant it thought you hadn't read the post, but instead I decided to fix the whole thing where it sends `last_seen` in the form rather than just figuring it out itself.

This seems to work fine, on testing. I expect it will overall reduce more obscure edge cases.

`@last_seen_id` hasn't been totally removed (from WritableController, mostly) because it's still used as a parameter for those links warning you you're not on the last page of the post.

### DIFF
--- a/spec/controllers/replies_controller_spec.rb
+++ b/spec/controllers/replies_controller_spec.rb
@@ -71,7 +71,7 @@ RSpec.describe RepliesController do
       expect(flash[:error][:message]).to eq("Your post could not be saved because of the following problems:")
     end
 
-    it "requires post read if no last_seen param" do
+    it "requires post read" do
       reply_post = create(:post)
       login_as(reply_post.user)
       reply_post.mark_read(reply_post.user)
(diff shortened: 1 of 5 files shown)


spec/controllers/replies_controller_spec.rb
```diff
@@ -111,18 +113,6 @@
...
-      login_as(user)
-      character = create(:character)
-      reply_post = create(:post)
-
-      expect(character.user_id).not_to eq(user.id)
-      post :create, reply: {character_id: character.id, post_id: reply_post.id}, unread_warned: true
-      expect(response).to redirect_to(post_url(reply_post))
-      expect(flash[:error][:message]).to eq("Your post could not be saved because of the following problems:")
-    end
-
```
Throne3d: Due to how `last_seen` only takes effect when there's an unread *reply* in the post, this test didn't break. It's no longer applicable, though, so I've removed it.

spec/controllers/replies_controller_spec.rb
```diff
@@ -137,20 +127,6 @@
...
-      reply_post = create(:post)
-      expect(Reply.count).to eq(0)
-
-      post :create, reply: {post_id: reply_post.id}, unread_warned: true, content: 'test!'
-
-      reply = Reply.first
-      expect(reply).not_to be_nil
-      expect(response).to redirect_to(reply_url(reply, anchor: "reply-#{reply.id}"))
-      expect(flash[:success]).to eq("Posted!")
-    end
```
Throne3d: Likewise.

PR #237, opened 2017-05-20
Upgrade to Rails 4.0

Throne3d:
So all tests now pass but I need to clean up the commit history and also check that I haven't left out params (and maybe consistently use strong params instead of the weird `attr_accessible` thing I'm doing currently).


Throne3d:
This should now not be a work in progress! I have tested relatively thoroughly the models I have upgraded and no tests are broken and I think it works pretty well.

Throne3d:
I have read through the [upgrade guide](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-3-2-to-rails-4-0) also and I seem to have covered everything there.

app/views/layouts/application.haml
```diff
@@ -22,10 +22,10 @@
     %meta{name: "application-name", content: "Glowfic Constellation"}
     %meta{name: "msapplication-config", content: asset_path('favicons/browserconfig.xml')}
     %meta{name: "theme-color", content: "#ffffff"}
-    = favicon_link_tag asset_path('favicons/favicon.ico')
-    = favicon_link_tag asset_path('favicons/favicon-32x32.png'), type: 'image/png', rel: 'icon', sizes: '32x32'
-    = favicon_link_tag asset_path('favicons/favicon-16x16.png'), type: 'image/png', rel: 'icon', sizes: '16x16'
-    = favicon_link_tag asset_path('favicons/apple-touch-icon.png'), type: 'image/png', rel: 'apple-touch-icon', sizes: '180x180'
+    = favicon_link_tag 'favicons/favicon.ico'
```
Marri: Why/how does this change? I would like to be more sure than I am presently that this will somehow yield properly asset-managed urls? Like, does this in fact go to glowfic.com/favicons/favicon-ridiculoushashhere.ico, or just favicon.ico?

Gemfile
```diff
@@ -18,7 +19,8 @@ gem 'nilify_blanks'
 gem 'nokogiri'
 gem 'pg'
 gem 'pg_search'
-gem 'rails', '3.2.22.5'
+gem 'protected_attributes'
```
Marri: What's this for?

db/migrate/20170519123223_add_request_uuid_to_audits.rb
```diff
@@ -0,0 +1,10 @@
+class AddRequestUuidToAudits < ActiveRecord::Migration
+  def self.up
+    add_column :audits, :request_uuid, :string
```
Marri: I assume this is now somehow required by 4.0? Do we have to do anything more? Are we sure audits are still getting made?

Gemfile
Throne3d: Means we can use "attr_accessible" and "attr_protected" now - they've been removed from rails by default with 4.0 and it also now requires strong params (meaning if you don't do anything in the controller, it will crash with an error, since the params are not whitelisted). Alternatively this thing makes it optionally done in the models with attr_accessible etc, which I've left in place for replies and posts and some other models.

db/migrate/20170519123223_add_request_uuid_to_audits.rb
Throne3d: Yes, it is required. No, the audits gem just says to run a particular rails generate command after each upgrade and when I didn't it crashed (on boot, I think). I haven't checked but I'm pretty sure; I'll do that and we should have tests for it?

Gemfile
Throne3d: We could upgrade every model and controller to use strong params instead, if you'd rather?

app/views/layouts/application.haml
Throne3d: This routes through `/assets/favicons/favicon.ico` for me in dev mode, which does serve the appropriate asset.

The exact error, on this line:

```haml
= favicon_link_tag asset_path('favicons/favicon.ico')
```

is:

> Asset names passed to helpers should not include the "/assets/" prefix. Instead of "/assets/favicons/favicon.ico", use "favicons/favicon.ico"
> app/views/layouts/application.haml:25:in '_app_views_layouts_application_haml__3992509873192304109_46982837423520'
app/controllers/application_controller.rb:78:in 'set_timezone'

I assume this is because `asset_path` autogenerates the prefix, and it doesn't look like it's adding a hash, so I'd expect this change to make it work as it suggests.

Throne3d: [Here](http://guides.rubyonrails.org/v4.1/asset_pipeline.html#coding-links-to-assets) seems to not expect `asset_path` to be used despite sprockets being involved.

db/migrate/20170519123223_add_request_uuid_to_audits.rb
Throne3d: Audits definitely appear to still be getting made.

Throne3d:
In case I missed something, we should probably (when pushing this, or in the couple of days after) check for "Can't mass-assign protected attributes" or "Unpermitted parameters:" (the two signs I've seen so far for having failed to whitelist an attribute).

app/models/character.rb
```diff
@@ -54,7 +54,7 @@ def valid_template
       errors.add(:template, "must be yours") if template.present? && template.user_id != user.id
       return
     end
-    @template = Template.new(user: user, name: new_template_name)
+    @template = Template.new(name: new_template_name, user: user)
```
Throne3d: this change is due to an intermediate thing I meant to revert (but ended up reverting in reverse I guess?), oops. shouldn't affect anything

app/models/galleries_icon.rb
```diff
@@ -2,4 +2,12 @@ class GalleriesIcon < ActiveRecord::Base
...
   belongs_to :gallery
   accepts_nested_attributes_for :icon, allow_destroy: true
+
+  after_destroy :unset_has_gallery
+
+  def unset_has_gallery
+    icon = Icon.find_by_id(icon_id)
+    return if icon.galleries.present?
+    icon.update_attributes(has_gallery: false)
+  end
```
Throne3d: Using `gallery#edit` to destroy a `GalleriesIcon` wasn't unsetting `has_gallery`… I'm not sure if this is a patch that should be applied to the Rails 3 thing; I'll investigate

app/models/message.rb
```diff
@@ -9,6 +9,8 @@ class Message < ActiveRecord::Base
 
   after_create :set_thread_id, :notify_recipient
 
+  attr_accessible :recipient, :recipient_id, :parent, :parent_id, :thread, :thread_id, :subject, :message, :visible_inbox, :visible_outbox, :marked_inbox, :marked_outbox, :read_at, :unread # TODO: figure if marked still used
```
Throne3d: Note to self: there's a non-urgent `TODO` here

app/models/message.rb
```diff
@@ -46,7 +48,9 @@ def site_message?
   end
 
   def self.send_site_message(user_id, subject, message)
-    Message.create(recipient_id: user_id, sender_id: 0, subject: subject, message: message)
+    msg = Message.new(recipient_id: user_id, subject: subject, message: message)
+    msg.sender_id = 0
+    msg.save
```
Throne3d: Can't mass-assign `sender_id` even outside a controller when using `attr_accessible` (I could get around this by making `Message` use strong params but I have inconsistently migrated to strong params and it's not necessary)

app/models/post_viewer.rb
```diff
@@ -3,4 +3,5 @@ class PostViewer < ActiveRecord::Base
   belongs_to :user
 
   validates_presence_of :user, :post
+  attr_accessible :user, :user_id, :post, :post_id
```
Throne3d: Otherwise it fails to add `post_id` on creation through setting `viewer_ids` – not sure if there's a simpler way around this.

app/models/template.rb
```diff
@@ -1,8 +1,8 @@
 class Template < ActiveRecord::Base
   belongs_to :user, inverse_of: :templates
-  has_many :characters, order: 'LOWER(name) ASC'
+  has_many :characters, -> { order('LOWER(name) ASC') }
 
-  validates_presence_of :name
+  validates_presence_of :name, :user
```
Throne3d: (Template should require a user.)

app/models/user.rb
```diff
@@ -5,7 +5,6 @@ class User < ActiveRecord::Base
   MAX_USERNAME_LEN = 80
 
   attr_accessor :password, :password_confirmation
-  attr_protected :crypted
```
Throne3d: We're using strong params for `User` now so `:crypted` is just not on the whitelist

config/routes.rb
```diff
@@ -7,7 +7,7 @@
 
   # Accounts
   match '/login' => 'sessions#new', :as => :login, :via => :get
-  match '/login' => 'sessions#create', :as => :login, :via => :post
+  match '/login' => 'sessions#create', :via => :post
```
Throne3d: This doesn't cause any existing tests to fail and clicking the login button still works for me; I'm pretty sure we don't need the `POST` route to be named, at least not separately from the `GET` route?

db/migrate/20170519123223_add_request_uuid_to_audits.rb
Throne3d: I'm not sure where the audit documentation actually is. There's [this](https://github.com/collectiveidea/audited) which looks very similar but isn't guaranteed to work on even Rails 4.0 and it seems to *think* further changes should be necessary for audits to work… but they don't seem to be…?

spec/controllers/aliases_controller_spec.rb
```diff
@@ -99,7 +99,7 @@
 
   describe "DELETE destroy" do
     it "requires login" do
-      delete :destroy, id: -1
+      delete :destroy, id: -1, character_id: -1
```
Throne3d: Alias destroy isn't a route without a character ID

app/models/galleries_icon.rb
Throne3d: Test fails on master also; I've moved it out to PR #240.

Throne3d:
I have now reordered some commits and cleaned up the history and extracted some irrelevant bits to hopefully make this easier.

config/routes.rb
Marri: Naming's just for code reasons; this should work fine.

PR #243, opened 2017-05-23
Add tests to ensure model attributes are set on create and update

Throne3d:
This also fixes `post_viewers` being invalid on post#create in Rails 4 (by adding an `inverse_of` attribute).

### DIFF
--- a/app/concerns/viewable.rb
+++ b/app/concerns/viewable.rb
@@ -36,7 +36,7 @@ module Viewable
     def reload
       @view = nil
       @first_unread = nil
-      @visible = false # for post.visible_to? in tests
+      @visible = nil # for post.visible_to? in tests
       super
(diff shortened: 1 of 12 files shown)


app/concerns/viewable.rb
```diff
@@ -36,6 +36,7 @@ def last_read(user)
     def reload
       @view = nil
       @first_unread = nil
+      @visible = false # for post.visible_to? in tests
```
Marri: This seems like it should almost certainly be nil, not false?

Throne3d: Oh, I guess, but I'm like 80% sure `||=` actually treats them the same.

Throne3d: (Changed.)

PR #248, opened 2017-05-29
Upgrade to Rails 4.2

Throne3d:
It could be worth investigating [the new HTML sanitizer](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#rails-html-sanitizer) so we aren't using a gem to do this; perhaps it'll reduce memory usage overall, while still using the same underlying framework (of Nokogiri).


Throne3d:
Rails now by default [does not bind to 0.0.0.0](https://fullstacknotes.com/make-rails-4-2-listen-to-all-interface/) meaning you'll have trouble if you try to connect to it and it's running in a virtualbox.

To get around this, you can add the following to `config/boot.rb` (but note you can't use the web console like this):
```ruby
require 'rails/commands/server'

module Rails  
  class Server
    alias :default_options_bk :default_options
    def default_options
      default_options_bk.merge!(Host: '0.0.0.0')
    end
  end
end 
```

Might be worth just using Firefox in the VM itself if you do this.

Throne3d:
Probably less WIP now. In fact, so un-WIP that I'm looking for feedback. xD

config/environments/development.rb
```diff
@@ -31,14 +31,17 @@
...
+
   # Debug mode disables concatenation and preprocessing of assets.
   # This option may cause significant delays in view rendering with a large
   # number of complex assets.
   config.assets.debug = false
 
   # Asset digests allow you to set far-future HTTP expiration dates on all assets,
   # yet still be able to expire them through the digest params.
-  config.assets.digest = false
+  config.assets.digest = true
```
Marri: This concerns me. Have you been able to test this at all? We don't use the asset pipeline in development.

config/environments/production.rb
```diff
@@ -75,6 +77,9 @@
   # Use default logging formatter so that PID and timestamp are not suppressed.
   config.log_formatter = ::Logger::Formatter.new
 
+  # Do not dump schema after migrations.
+  # config.active_record.dump_schema_after_migration = false
```
Marri: Why is this commented out? Shouldn't we stop this in prod?

Marri:
Re: 0.0.0.0, can you not use `heroku local` instead? Probably a more accurate local env anyway.
Re: HTML, it looks like you have to take steps to _not_ use it, or am I missing something?

config/environments/development.rb
Throne3d: I've been able to load pages in development and it shows digests on the assets that are attached to the page? They load just fine, and it was also in the recommended (default) config when I ran `rake rails:update`.

config/environments/production.rb
Throne3d: Because I wasn't sure that we wanted this, for sure. I can uncomment it just fine.

Throne3d:
Re: HTML, we currently use `Sanitize`, which is a separate gem, to sanitize our HTML, which was us taking steps not to use the default Rails sanitizer (at some point in the past, because it wasn't good enough for our purposes). So yes, you do, but we've already taken them.

Re: 0.0.0.0… I've never tried `heroku local`. I'll poke around a little.

Throne3d:
Looks like I can use `heroku local`! Problem is it doesn't output any debug info to console, like the standard `SELECT * FROM characters …`? This happens even if I add to `config/environments/development.rb` the line `config.log_level = :debug`, which I think shouldn't be necessary. Perhaps this is because it also uses Puma, when running `heroku local`. (Also the port changes to `5000` instead of `3000` but that's a minor thing.)

Marri:
IIRC it logs to log/development.rb, you'd have to tail

Throne3d:
This should be ready for review, I think?

PR #256, opened 2017-06-18
Fix report zone issues

Throne3d:
This should fix the tests to work in any timezone without forcing the server timezone pre-tests, and also tests reading things across the world (Hawaii, UTC, Auckland, which should test the full day-or-two at any moment in time).

### DIFF
--- a/spec/controllers/reports_controller_spec.rb
+++ b/spec/controllers/reports_controller_spec.rb
@@ -65,49 +65,52 @@ RSpec.describe ReportsController do
     end
 
     context "reading" do
-      it "does not mark read for today's unfinished report" do
-        user = create(:user)
-        expect(user.report_view).to be_nil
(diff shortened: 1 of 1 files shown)


Throne3d:
… Hm, of course it fails tests.

I'll look into it.

Marri:
Once tests pass this can merge

Throne3d:
I think I have fixed tests! I wasn't totally sure what I was doing before nor what I was meant to be testing precisely nor how the timezone thing was interacting (and also there was the "it passed on my machine" thing and I was tired), but I think I've now figured that out (see commit message) and done it in the tests.

Throne3d:
This also only touches tests, not application code, which is why I assume you approved it, so I'll merge it if tests do pass just fine on Travis as well as my machine.

Issue #252, opened 2017-06-12
Editing icons/characters does not update flat views


Throne3d:
Check on this. I don't recall if they do or not. I don't expect they would, skipping validations, but maybe they do.

Throne3d:
Relatedly: character name changes, screenname changes.

This could be solved (somewhat) by allowing people to flag posts as needing a flat view regeneration, if changes like that are made. Alternatively we could just make sure to requeue all relevant posts.

Semi-relatedly is to paginate flat view (not in display, just in generation), so we put a hard cap on how many things we have to regenerate at a time when a post is just being updated. This prevents RoR-esque collapses.

Marri:
#631 solves this for icons and character replacements but not for changing the information of the character/icon model itself such as character name or screenname. As mentioned, #796 may provide the thing well enough.

PR #262, opened 2017-07-04
Improve flat mode with new div classes and add update times

Throne3d:
Some flat mode styling was broken as the posts and replies used to use even/odd instead of post-reply and post-post (see, e.g., blockquotes), and they also didn't include edit times. This should fix that (but might need old posts to have their flat modes regenerated at some point).

### DIFF
--- a/app/views/posts/_generate_flat.haml
+++ b/app/views/posts/_generate_flat.haml
@@ -1,7 +1,6 @@
 - 1.upto((replies.count(:all) / 250.0).ceil).each do |index|
   - replies.paginate(per_page: 250, page: index).each do |reply|
-    - klass = cycle('even'.freeze, 'odd'.freeze)
-    .post-container{class: klass}
+    .post-container{class: 'post-reply'}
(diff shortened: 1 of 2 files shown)


Marri:
I was semi intentional about the edited date because it was more calculation to do in an already large, slow process, for not much user value. Has it been bothering people?

Throne3d:
I think `pretty_time` (seeing as it basically seems to be a `strftime`) shouldn't add much overhead to this, especially relative to how long it takes to generate the HTML itself?

I was planning on migrating to flat modes for my scraper (for the EPUB generation, which should occur at some point since I haven't updated them in a while), but my scraper uses the edited timestamps so it can display relevant information in the EPUBs themselves. Also I don't yet use flat mode, really, but I think I'd be bothered by not knowing when a post has been edited since I expect that information to be visible if relevant. I don't think it was ever mentioned that we were not displaying that sort of data.

Throne3d:
See #252.

Throne3d:
(Stalled on me figuring out how to get it to regenerate all the flat views.)

PR #285, opened 2017-08-08
Scrape Effulgence

Throne3d:
- Removes the special post list variant for Effulgence
- `scraper.rb` ⇒ `scraper_incandescence.rb`
- Fixes threaded scraping with multi-page threads if the journal doesn't use the default theme (by making it append `style=site` to the URL params)
- Adds a `scrape_threads!(threads)` method to `PostScraper` which allows specifying particular sub-threads (in an array) of a post to scrape; this is used later in the Effulgence scraper and may be useful for other applications? There's a test, which also makes sure multi-page threads work.
- Adds a `scraper_effulgence.rb` to enable scraping Effulgence. This special-cases 'repealing' to group the first list of threads into a 'guest list' post, since they're all very short, and imports posts inside sections which have threads distinctly, prefixing the titles. (e.g. "ma fête: before the guests arrive", "ma fête: the pile of Jokers", "ma fête: eventually extracts herself" – all are part of the continuity section "welcoming".)

~~Currently a work in progress while I try to import it locally and check a few things.~~


Throne3d:
[`scrape_threads!` might not be super necessary but I wanted it to be able to import a threaded post and selectively pick which threads it import. I could move that out and then try to find or fabricate some other fixtures for my test, if you'd prefer not to have that.]

lib/post_scraper.rb
```diff
@@ -16,17 +16,17 @@ class PostScraper < Object
...
-  attr_reader :url
+  attr_accessor :url, :post, :html_doc
 
   def initialize(url, board_id=nil, section_id=nil, status=nil, threaded_import=false, console_import=false)
     @board_id = board_id || SANDBOX_ID
     @section_id = section_id
     @status = status || Post::STATUS_COMPLETE
-    url = url + (if url.include?('?') then '&view=flat' else '?view=flat' end) unless url.include?('view=flat')
+    url = url + (if url.include?('?') then '&view=flat' else '?view=flat' end) unless url.include?('view=flat') || threaded_import
     url = url + '&style=site' unless url.include?('style=site')
```
Marri: This probably isn't safe any more; it assumes that you're guaranteed a ? because of the previous line. I'd move the if statement to be this block, and the view=flat to be the 'well you're guaranteed now" one.

Throne3d: Fixed

Throne3d:
I am mostly happy with this now! It functions I think as expected.

PR #287, opened 2017-08-12
Fix icon matching in PostScraper with accessibility descriptions

Throne3d:
Kappa has various icons with accessibility descriptions, which Dreamwidth places into icon `title` attributes in parentheses next to the keyword. These should be stripped when matching for duplicates, as this does.

Otherwise we get a lot of parenthesized version of keywords in galleries:

image

- https://www.glowfic.com/galleries/609

### DIFF
--- a/lib/post_scraper.rb
+++ b/lib/post_scraper.rb
@@ -217,11 +217,28 @@ class PostScraper < Object
       icon = tag.character.icons.where(keyword: keyword).first
       tag.icon = icon and return if icon
 
+      # split out the last " (...)" from the keyword (which should be at the very end), if applicable, for without_desc
+      without_desc = nil
+      if keyword.end_with?(')')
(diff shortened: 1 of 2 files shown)


lib/post_scraper.rb
```diff
@@ -217,11 +217,28 @@ def set_from_icon(tag, url, keyword)
       icon = tag.character.icons.where(keyword: keyword).first
       tag.icon = icon and return if icon
 
+      # split out the last " (...)" from the keyword (which should be at the very end), if applicable, for without_desc
+      without_desc = nil
+      if keyword.end_with?(')')
+        lbracket = keyword.rindex(' (')
+        if lbracket && lbracket > 0 # without_desc must be non-empty
```
Marri: This can be .to_i > 0 but whatev I'll merge for now

Issue #282, opened 2017-08-07
Investigate InvalidAuthenticityToken errors


Throne3d:
Currently people are getting mysterious `InvalidAuthenticityToken` errors. Authenticity tokens are a thing done automatically by Rails when using destructive methods (e.g. POST), and we currently have session to never expire, so we're not entirely sure what's up with this. It started approximately when we moved to Rails 4.

Ideas to investigate:

- Somehow the session is still expiring, possibly due to a long time elapsing without the form submission

Things to fix:

- [x] Users are logging out and back in (possibly _being logged out_, since we've been having some issues with that), which makes the old form invalid; we could fix this with a JavaScript note to reload the page (see #296)

Keywords: CSRF

Throne3d:
Latter suggestion repros.

Throne3d:
Recommendation to check for log-in, as our cookies appear to be HTTP-only (that is, transport-only, as opposed to being dynamically modified, meaning JavaScript cannot tamper with them), is to set a `localStorage` value for `loggedIn` → `true`, and in other tabs use the `storage` event ([see here](https://developer.mozilla.org/en-US/docs/Web/Events/storage)) to catch the change and display a message to reload.

Throne3d:
See #269. Also something needs to be done about page timeouts, but I'm not sure how to do that without implementing a `POST /ping` endpoint or such.

Throne3d:
It looks better since upgrading to Rails 5 but it's still definitely not fixed.

Throne3d:
We could rescue from InvalidAuthenticityToken exceptions by presenting the user with a page saying their session has expired, and also by clearing their session. I think this is only not recommended when you're not sure you're clearing their session successfully, so possibly we should clear the session and then _redirect_, so we don't have traces of the old session left in the request.

teceler:
Handled with #1102

PR #296, opened 2017-08-15
Display a warning when the user logs in/out in another tab

Throne3d:
Uses `gon` to be notified of login status. This is in an attempt to reduce the prevalence of InvalidAuthenticityToken errors (see issue #282), by notifying users when their forms become invalid due to a change in login status.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -22,6 +22,38 @@ $(document).ready(function() {
     width: '100%',
     minimumResultsForSearch: 10,
   });
+
+  // Set localStorage if login status has changed
+  // storage values are usually strings, cast to be sure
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/application.js
```diff
@@ -22,6 +22,38 @@ $(document).ready(function() {
...
+      localStorage.setItem(loggedInKey, wasLoggedIn);
+    }
+  }
+
+  // Watch for login status change
+  // Display warning prompting user to reload when it occurs
+  window.addEventListener('storage', function(e) {
+    // skip non-login-status changes
+    if (e.key !== loggedInKey) return;
+    console.log(e.key, '–', e.oldValue, '→', e.newValue);
```
Marri: ?

PR #315, opened 2017-08-25
Make the background of the post editor dark in dark modes

Throne3d:
This is being resurrected from an old branch, so might contain some weird things at present.

- [x] Make the TinyMCE editor background dark in dark modes
- [x] Make the TinyMCE buttons dark in dark modes (needs to be checked with Aestrix *so* hard)
- [ ] Make other form boxes and buttons dark in dark modes


Throne3d:
Will build on #395.

PR #325, opened 2017-08-26
Improve board and board section organization

Throne3d:
- Remove posts further from the section reordering in `boards#edit`
- Reword new section creation message
- Improve responsiveness of loading / saved notifications
- Overhaul reordering logic, client-side, and disable arrows when busy or non-functional (at boundaries)
- Add reordering to boards for unsectioned posts


Throne3d:
Still potentially to do:

- Return full ordering and other data when reordering performed, so the layout is *fully accurate* to what the server has, instead of attemptedly calculated.

Throne3d:
Todo: test endpoint with various sample data; make sure it rejects invalid data sent (from multiple boards, from multiple sections, from a section without a `section_id`, from no section with a `section_id`, from posts that don't exist, from sections that don't exist, from boards I cannot edit, from boards that don't exist (?)) and that it does the right thing and returns the right data (including new boards if created in the meantime! also maybe go have it map the sections to IDs and compare the response so that if a board is created and destroyed you still get notified of weirdness!).

PR #328, opened 2017-08-26
Fix more Javascript linting

Throne3d:


Throne3d:
Cursory test of the relevant pages still works.

PR #342, opened 2017-08-28
[WIP] Add gallery groups to characters

### LINKED ISSUE #330: Gallery tags

Throne3d:
Builds on #341, so requires everything there to be done first.

Fixes #330.

TODO:

- [X] Convert subquery in tags API controller to something more intelligible (joins?)
- [X] ~~Display a warning if user has manually excluded galleries that were in a group, so it is not easy to miss; maybe send list of grouped IDs?~~ … no, instead, make it not exclude, make sure it does not exclude, make sure it's just the added on/off toggle.
- [x] Show galleries on character page by AJAX (see "TagsController offers correct galleries…")
  - [x] Fix `character#new` if you add and remove a character group!!! [done? seems to have instead been a problem with adding a character group twice, maybe? not sure...]
- [x] Test *everything*, all the things
- [x] Test *all the things* if the form save fails, especially all the stuff around `ungrouped_gallery_ids`
- [x] Expose characters on the relevant tag pages
- [x] Leave a note about where the test code is duplicated (for model tags)
- [X] Rebase onto #341 once merged.
- [x] Make sure character duplicate contains correct `ungrouped_gallery_ids` and `gallery_group_ids`

Tests:

- [x] The "galleries" list, in the character editor, corresponds to galleries that are not added by groups
- [x] Removing galleries from this list either flags it "added by group" (if it can be found in an attached group), or removes it from the character (if it cannot be); adding them creates the relevant tie if it doesn't already exist, and flags it "not added by group"
- [x] Adding a gallery at the same time you remove the group should keep the gallery and not the others
- [x] Adding a character to a group: does not add another user's galleries, does not add galleries twice, adds galleries from the given group
- [x] Adding a gallery to a group: does not add to another user's characters, does not add to characters twice, adds to characters from the given group
- [x] TagsController offers correct galleries for tag#show with a GalleryGroup and appropriate wants
- Other controller tag-related things (for CharactersController)
  - [x] Persists if saving fails
  - [x] Successfully submits the tags
- Tag controller handles it well
  - [x] Shown characters in tags#show
  - [x] Count characters in tags#index
- [x] editor loads? try doing `render_views` for characters#edit and characters#new
- [x] character duplicate

### DIFF
--- a/app/assets/javascripts/characters/editor.js
+++ b/app/assets/javascripts/characters/editor.js
@@ -1,8 +1,14 @@
 /* global gon, createTagSelect */
 var galleryIds;
+var galleryGroupIds = [];
+var galleryGroups = {};
 
 $(document).ready(function() {
-  galleryIds = $("#character_gallery_ids").val() || [];
+  gon.gallery_groups.forEach(function(galleryGroup) {
+    galleryGroups[galleryGroup.id] = galleryGroup;
+  });
+  galleryIds = $("#character_ungrouped_gallery_ids").val() || [];
+  galleryGroupIds = $("#character_gallery_group_ids").val() || [];
 
   $("#character_setting_ids").select2({
     width: '100%',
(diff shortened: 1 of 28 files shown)


Throne3d:
It actually wasn't WIP when I merged it, I just forgot to change the title. Oops.

PR #359, opened 2017-08-31
Improve post metadata

Throne3d:
WIP on broken table layout à cause de the subheader


Throne3d:
Broken table layout fixed.

Issue #23, opened 2017-02-04
Better sanitize post HTML (specifically divs)


Throne3d:
See reply 168901. Lintamande copied a post (as a quote) into a reply, and currently classes are not stripped from divs.

Perhaps classes should be stripped from divs or the CSS around `post-content`s should be adjusted for such a scenario. Alternatively see the issue about a proper quotes system.

Throne3d:
Let's restrict the `.post-container` formatting to be for `.post-containers` which are direct children of `#content`, how about.

PR #378, opened 2017-09-05
Fix post container formatting

Throne3d:
A check for all locations that contain `.post-container` in the views produces various documents, which are then included in various locations. ("root element" means the `.post-container` appears as a direct child of the root document, or in the case of `posts#flat`, an explicit child of `#content`; numbers given are line numbers.)

- messages#single (10, 18: root elements) hence messages#preview (16, 17: root elements), messages#show (10: root element)
- posts#flat (28: root element)
- posts#new_import (4: root element)
- posts#generate_flat (4: root element)
- replies#single (1: root element) hence posts#history (18, 25: root elements), posts#preview (6: root element), posts#show (129 root, 134 **not root** but gets converted to be root on display, 137 root), replies#preview (8, 18: root elements)

As stated, in `posts#show`, line 134 is a non-root element (the expander for the main post on subsequent pages). The first commit here makes the JavaScript convert the whole expander into the HTML of the post on expansion, so it'll display identically to the first page, instead of being `#content > .post-expander > .hidden > .post-content`. (This has a side effect of fixing the previous erroneous padding found on that post when expanded.)

This is currently a WIP while I fix all the CSS and the notes above are places for me to check that everything displays properly.

### DIFF
--- a/app/assets/javascripts/posts/show.js
+++ b/app/assets/javascripts/posts/show.js
@@ -10,7 +10,7 @@ $(document).ready(function() {
 
   $(".post-expander").click(function() {
     $(this).children(".info").remove();
-    $(this).children(".hidden").show();
+    $(this).get(0).outerHTML = $(this).children('.hidden').html();
   });
 
   // Dropdown menu code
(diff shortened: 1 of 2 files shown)


Throne3d:
Hmmm. It looks like the quote in the given ID (see #23) uses inline CSS from the copy-pasting. Ew.

This PR has got a whole lot smaller in scope, then, since I'm just modifying one small part of the CSS.

PR #383, opened 2017-09-06
Show progress of file upload on gallery#add_new

Throne3d:
I can't actually test this since I don't have an environment that supports file uploads. There are a couple of other changes I want to do to uploads, too, as in #8.


Throne3d:
Looks like it's showing progress in bytes or something? I will investigate changing this.

Throne3d:
Okay, this works and now does it per item and also shows the number that failed! I'm happy to merge it.

Issue #57, opened 2017-02-04
Tests: Write them


Throne3d:
An ongoing process.

- [X] RSpec *(mostly)*
- [ ] Fix the TODOs in RSpec
- [ ] DRY up our controller specs (#586)
- [x] Feature specs (move from Cucumber to RSpec & Capybara?)
- [ ] Thorough helper specs
- [ ] View specs
- [ ] Thorough mailer specs (might be complete?)
- [x] Job specs? (Might require using ActiveJob for our jobs)
- [ ] Javascript (#389)

Throne3d:
RSpec tests are mostly done, now. Cucumber up next?

Throne3d:
We're investigating using RSpec (and Capybara) for request and feature specs, as opposed to Cucumber (which we currently have set up for the feature specs).

Request specs are like controller specs, except in RSpec 3.5+ (and Rails 5.0+), controller tests (functional) are discouraged in favor of request specs (integrational), since the full routing, session and response parts, as well as other bits, of the stack, are made sure to work with the relevant request. This apparently helps to reduce edge case issues, though we may find it harder to test some things, such as where we currently access a controller's `assigns`, since it looks like this is removed. (We may need to artificially select the new model, in cases where new models are made, or select the model-to-be-updated in cases where that applies – I'm not sure.)

Feature specs are designed to test the application from a web browser's perspective, totally driven externally, integrationally. Apparently you can also test JavaScript integration using Capybara (and I believe it's by default set up to use Chrome as the browser it tests on, with screenshot support for failures!?). It seems useful to look into this.

Other things we might want to look into:

- [ ] Proper helper specs (go through helpers thoroughly)
- [ ] DRY up our controller specs (well, request specs when we get around to that), especially the areas that are like "requires login", "requires ID", "requires valid ID"
- [ ] View specs (which are less integrational and make sure the views contain proper data for the relevant assigns!)
- [ ] Improved mailer specs (we have these to an extent, but I'm not sure it's as comprehensive as it could/should be – do we test each location they're created, all the relevant params, and then that the relevant jobs produce the actual emails accurately with the relevant views and so on? We may want to look into fixtures with some sample data too.)
- [ ] Job specs? (Might require using ActiveJob for our jobs)

Marri:
I generally test both jobs and mailers, did I not?

Throne3d:
[...]

PR #389, opened 2017-09-06
[WIP] Start testing the javascript

Throne3d:
Based on some quick research, it seems like [Jasmine](https://jasmine.github.io/) and [Teaspoon](https://github.com/jejacks0n/teaspoon) are pretty good choices for Rails Javascript testing (mostly unit and integrational, respectively).

I've added the bare-bones (quickly) to have it run these test frameworks and will look into adding some actual tests shortly.

Addresses more of #57.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ before_script:
 script:
   - RAILS_ENV=test bundle exec rails db:migrate --trace
   - bundle exec rspec spec
+  - bundle exec rake teaspoon
 branches:
   only:
   - master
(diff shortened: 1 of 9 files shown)


Throne3d:
I've now added an actual test! [This](https://github.com/Marri/glowfic/pull/389/commits/63b24e96c733a91ad9bbb42c678e426cdbbffc4f) ensures that the global.js ready handler definitely sets the localStorage key from the gon data.

Throne3d:
Prefer to test this with integrational tests via (probably) Capybara instead. We could investigate mocha if we want to test the javascript by itself but it's a bit pointless when most of it is 80%+ dependent on the page.

PR #391, opened 2017-09-06
Upgrade gem seed_dump to ~> 3.2

Throne3d:


Throne3d:
Looks like `Object.true?` is stolen from the `seed_dump` gem, and was removed when it got upgraded to version 3.0. See [this relevant commit](https://github.com/rroblak/seed_dump/commit/399190df6792f5d6fbb2466e7f6842b7b0c36700).

I'll look into removing all the instances of `be_true` in our tests and make them into `eq(true)` instead?

PR #408, opened 2017-09-08
Also reset character_template_attributes_id as necessary

Throne3d:


app/assets/javascripts/characters/editor.js
```diff
@@ -37,10 +38,19 @@ $(document).ready(function() {
...
         $("#character_template_attributes_name").val($("#character_template_id option:selected").text());
+        $("#character_template_attributes_id").val(oldTemplate);
+      } else {
+        $("#character_template_attributes_name").val('');
+        $("#character_template_attributes_id").val('');
       }
     }
   });
 
+  $("#character_template_id").change(function() {
```
Marri: What's this for? It seems like we'd either need to edit the New Template check box here as well, or check the value of the dropdown, or something?

Throne3d: It's so that if you use the `#character_template_id` dropdown to set the character's template, the nested attributes don't mess with things – if you use an empty ID and name, it doesn't seem to set anything in the nested attributes, ignoring them. I tested it manually, some, and it seemed to work on most main use cases, but I was thinking of doing some tests with data that would be included by this javascript to make sure it did the right thing in the controller.

PR #409, opened 2017-09-08
Image upload refactoring

Marri:
Improvements include:
* Storing the icon URL and S3 key separately, since S3 actions are taken on the key but the URL needs to be properly escaped to handle characters like + in icon filenames.
* Migration to add the new s3_key column will properly parse it out of existing icons and backfill.
* Supports showing a loading gif while icon uploads process; on edit pages, rolls back the image to the original image if upload fails.
* Adds multi-file-upload support. Installs it into the gallery#edit page.
* File upload pages now keep track of files they have uploaded, and attempts to intelligently delete unused ones upon form submission / page close / row deletion.
** NOTE: page close is on hold for now, since the async nature of Javascript makes this trickier than I'd like.
* API endpoint for deleting S3 files added, with tests and docs

Remaining:
- [x] Call deleteUnused() on form submit
- [x] Call deleteUnused() on row deletion (galleries#add only)
- [x] Tests for API endpoint
- [x] Generate docs from API endpoint tests
- [x] Fix up the conflicts with the progress bar to not be quite so hacky

### DIFF
--- a/app/assets/javascripts/galleries/update_existing.js
+++ b/app/assets/javascripts/galleries/update_existing.js
@@ -1,16 +1,13 @@
-var originalUrl;
-$(document).ready(function() {
-  originalUrl = $("#edit-icon").attr('src');
-});
-
 function addUploadedIcon(url, s3_key, data, fileInput) {
   $("#icon_conf").show();
   $("#icon_url_field").hide();
   $("#icon_url").hide().val(url);
   $("#icon_s3_key").val(s3_key);
-  $("#edit-icon").attr('src', url).css('height', '');
+  $("#loading").hide();
+  $("#edit-icon").attr('src', url).show().removeClass('uploading-icon');
 }
 
-function setLoadingIcon() {
-  $("#edit-icon").attr('src', '/images/loading.gif').css('height', '20px');
+function setLoadingIcon(fileInput) {
+  $("#edit-icon").hide().addClass('uploading-icon');
+  $("#loading").show();
 }
--- a/app/assets/javascripts/galleries/update_existing.js
+++ b/app/assets/javascripts/galleries/update_existing.js
@@ -1,6 +1,16 @@
-function addUploadedIcon(url, data) {
+var originalUrl;
+$(document).ready(function() {
+  originalUrl = $("#edit-icon").attr('src');
+});
+
+function addUploadedIcon(url, s3_key, data, fileInput) {
   $("#icon_conf").show();
   $("#icon_url_field").hide();
   $("#icon_url").hide().val(url);
-  $("#edit-icon").attr('src', url);
+  $("#icon_s3_key").val(s3_key);
+  $("#edit-icon").attr('src', url).css('height', '');
+}
+
+function setLoadingIcon() {
+  $("#edit-icon").attr('src', '/images/loading.gif').css('height', '20px');
 }
--- a/app/assets/javascripts/galleries/uploader.js
+++ b/app/assets/javascripts/galleries/uploader.js
@@ -1,26 +1,26 @@
-/* global addUploadedIcon */
+/* global addUploadedIcon, originalUrl, setLoadingIcon, addCallback, failCallback */
+
+var uploadedIcons = {};
+
 $(document).ready(function() {
-  bindFileInput($("#icon_files"));
-});
-
-function updateBox(progressBox, done, total, failed) {
-  if (!progressBox) return;
-  var progress = parseInt(done / total * 100, 10);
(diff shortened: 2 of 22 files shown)


app/controllers/api/v1/icons_controller.rb
```diff
@@ -0,0 +1,27 @@
+class Api::V1::IconsController < Api::ApiController
+  before_filter :login_required, only: :destroy
```
Throne3d: should this be `s3_delete`?

app/controllers/api/v1/icons_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    description 'Processes S3 uploads that were not used'
+  end
+
+  api! 'Given an S3 key that the user has not turned into an icon, deletes the file from S3'
+  param :s3_key, String, required: true, desc: 'S3 object key'
+  error 401, "You must be logged in"
+  error 403, "Icon does not belong to the user"
+  error 422, "Invalid parameters provided: s3 key is in use."
+  def s3_delete
+    unless params[:s3_key].starts_with?("users/#{current_user.id}")
```
Throne3d: can we check for the trailing slash too, so user 3 can't delete user 31's (unused but uploaded) icons? (probably not a major problem but still)

db/migrate/20170907180029_add_s3_key_to_icon.rb
```diff
@@ -1,17 +1,25 @@
 class AddS3KeyToIcon < ActiveRecord::Migration
+  CLOUDFRONT_URL = 'https://d1anwqy6ci9o1i.cloudfront.net/'
```
Throne3d: <3

Throne3d:
I would callbackify better by actually passing them as arguments but then I'd be duplicating code between `galleries/add_new.js` and `galleries/update_existing.js` that you may have to edit later (so for now I've chosen not to but you may wish to do that to make it less "random global appearing from nowhere" and more encapsulated).

Throne3d:
Relevant issue: #140.

spec/factories.rb
```diff
@@ -68,6 +68,7 @@
 
     factory :uploaded_icon do
       url { "https://d1anwqy6ci9o1i.cloudfront.net/users/#{user.id}/icons/nonsense-fakeimg.png" }
```
Throne3d: This probably needs to be escaped?

spec/models/icon_spec.rb
```diff
@@ -96,7 +96,7 @@ def delete_key(key)
     end
   end
 
-  context "#use_asset_host" do
+  context "#use_icon_host" do
```
Throne3d: Also in the model?

app/models/icon.rb
```diff
@@ -1,21 +1,25 @@
...
-  has_and_belongs_to_many :galleries
+  has_many :reply_drafts
+  has_and_belongs_to_many :galleries # TODO do these joins delete?
 
   validates_presence_of :url, :user, :keyword
   validate :url_is_url
   validate :uploaded_url_not_in_use
   nilify_blanks types: [:string, :text, :citext] # nilify_blanks does not touch citext by default
 
+  before_validation :use_asset_host
```
Throne3d: also this? (→ `use_icon_host`)

app/assets/javascripts/galleries/update_existing.js
```diff
@@ -1,6 +1,16 @@
...
   $("#icon_url_field").hide();
   $("#icon_url").hide().val(url);
-  $("#edit-icon").attr('src', url);
+  $("#icon_s3_key").val(s3_key);
+  $("#edit-icon").attr('src', url).css('height', '');
 }
+
+function setLoadingIcon() {
+  $("#edit-icon").attr('src', '/images/loading.gif').css('height', '20px');
+}
```
Throne3d: lack of newline is still weird

app/assets/javascripts/galleries/uploader.js
```diff
@@ -1,26 +1,26 @@
...
+$(document).ready(function() {
   var form = $('form.icon-upload');
   var submitButton = form.find('input[type="submit"]');
   var formData = form.data('form-data');
-  var progressBox = fileInput.closest('td').find('.progress-box');
-  var done = 0;
-  var total = 0;
-  var failed = 0;
 
+  $.each($(".icon_files"), function( index, fileInput ) {
```
Throne3d: Not that it's important but I think most of these `$.each($(selector), function(…){…})` (and same for `$.map`) can be done in the format `$(selector).each(function(…){…})`

Throne3d:
I think you force-pushed out the thing I did to extract the gallery code thingy

PR #411, opened 2017-09-08
Make sure users with blank emails don't need unique emails

Throne3d:
Added tests to ensure. Fix is pending.

### DIFF
--- a/app/models/character.rb
+++ b/app/models/character.rb
@@ -31,7 +31,7 @@ class Character < ActiveRecord::Base
 
   acts_as_tag :label, :setting, :gallery_group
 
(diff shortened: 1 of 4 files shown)


Throne3d:
No longer WIP.

The Rails 4.2 upgrade caused the citext columns in PG to be recognized as citext, instead of the default 'string' – nilify_blanks by default only applies to string and text types (see DEFAULT_TYPES here: <https://github.com/rubiety/nilify_blanks/blob/master/lib/nilify_blanks.rb#L7>), and accordingly we should override that to apply to citext columns too, fixing the bug with blank emails being considered non-unique.

PR #437, opened 2017-09-12
[WIP] Use Rails 4.2's built-in sanitizer (based on Loofah) instead of a separate gem

Throne3d:
(See #248.)

As a result of the upgrade to Rails 4.2, we have access to the [rails-html-sanitizer](https://github.com/rails/rails-html-sanitizer) gem! This means we can dump the `sanitize` gem in favor of using the built-in gem (based on Loofah instead) and a custom class to enable the old behavior of specifying specific attributes for specific nodes.

[Loofah claims to be faster than Sanitize](https://github.com/flavorjones/loofah#compare-and-contrast), too, so we may drop in rendering times, as well as the fact we're very likely to reduce memory usage with this (at least some) since added gems are annoying.

Note that nokogumbo is removed (not nokogiri – I'm not sure of the difference between the two), and the old tests needed very slight changes (to whitespace) to ensure they still passed.

This also adds two new tests: the `width` attribute is cleared on `<a>` but not on `<hr>` (so we have selective attribute scrubbing by node name), and the `style="color: red;"` attribute is not cleared on `<a>` (so we have CSS definitely permitted on nodes).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,6 @@ gem 'resque'
 gem 'resque_mailer'
 gem 'resque-retry'
 gem 'resque-web', require: 'resque_web'
-gem 'sanitize'
 gem 'sass-rails'
 gem 'select2-rails'
 gem 'test-unit', '~> 3.0' # required by Heroku for production console
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -84,7 +84,6 @@ GEM
     concurrent-ruby (1.0.5)
     crack (0.4.3)
       safe_yaml (~> 1.0.0)
-    crass (1.0.2)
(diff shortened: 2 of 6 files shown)


Throne3d:
I'm not sure why we use `Sanitize.fragment` in the API controller's param validations? @Marri (I have tentatively removed it and it didn't break anything…? Do we rely on sanitizing Apipie's response and if so why?)

Throne3d:
(WIP while I query what Sanitize is used for in the Apipie thingy.)

Throne3d:
I should be more careful about removing the "WIP" from the title before merging. (It was not WIP anymore.)

Issue #448, opened 2017-09-14
Upgrade to Rails 5.0


Throne3d:
~~Main upgrade (#449):~~
- [x] Comb over [the upgrade notes](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-2-to-rails-5-0) to generate the rest of this TODO list
- [x] Upgrade the gem 4.2 to 5.0
- [x] Run `rails app:update` to update Rails configs
- [x] Upgrade the audit gem's configs ([link](https://github.com/collectiveidea/audited#upgrading))
- [x] Make an `ApplicationRecord` for models to inherit from by default ([link](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#active-record-models-now-inherit-from-applicationrecord-by-default))
- [x] Run `rails generate rspec:install` to get fresh updated RSpec configs, then merge with old modified ones: "In Rails 5.x, controller testing has been moved to its own gem which is rails-controller-testing. Using assigns in your controller specs without adding this gem will no longer work."
- [x] Upgrade `web-console` to `~> 3.0` (doesn't support Rails 4 properly)
- [x] A lot of things have been moved to `rails` from `rake`; make sure common commands still work
- [x] ~~Consider adding `prepend: true` to `protect_from_forgery` to keep the old behavior~~
- [x] Check digests show in assets on staging when pushed there ([per](https://github.com/Marri/glowfic/pull/449/files#diff-1d98f0039bbc6df92adee882ed99d993L35))
- [x] ~~Check we don't need actionmailer not to deliver in development ([per](https://github.com/Marri/glowfic/pull/449/files#diff-a2f09e1db7a47eb67c5849478a97f3d7L19))~~

Deprecation (#457):
(*Check the many deprecation warnings, see #457 for more info:*)
- [x] `ActiveRecord::Base.raise_in_transactional_callbacks=` (set to true, now forced true and the config does nothing; removed)
- [ ] Parameters are now objects instead of hashes and comparing equality is deprecated set to be removed in Rails 5.1
- [ ] `alias_method_chain` deprecated (in Presentable)
- [x] `before_filter`, `around_filter`, `after_filter` in basically all the controllers
- [x] Positional parameters in tests (`get :show, id: 1` vs `get :show, params: { id: 1 }`)
- [x] Rails target versions in migration superclasses (`< ActiveRecord::Migration[4.2]`)

Throne3d:
[...]

PR #449, opened 2017-09-14
Upgrade to Rails 5.0

Throne3d:
See #448.

- [x] Comb over [the upgrade notes](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-2-to-rails-5-0) to generate the rest of this TODO list
- [x] Upgrade the gem 4.2 to 5.0
- [x] Run `rails app:update` to update Rails configs
- [x] Upgrade the audit gem's configs ([link](https://github.com/collectiveidea/audited#upgrading))
- [x] Make an `ApplicationRecord` for models to inherit from by default ([link](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#active-record-models-now-inherit-from-applicationrecord-by-default))
- [x] Run `rails generate rspec:install` to get fresh updated RSpec configs, then merge with old modified ones: "In Rails 5.x, controller testing has been moved to its own gem which is rails-controller-testing. Using assigns in your controller specs without adding this gem will no longer work."
- [x] Upgrade `web-console` to `~> 3.0` (doesn't support Rails 4 properly)
- [x] A lot of things have been moved to `rails` from `rake`; make sure common commands still work (`rails db:migrate`, `rails spec`…)
- [x] ~~Consider adding `prepend: true` to `protect_from_forgery` to keep the old behavior~~
- [x] Check digests show in assets on staging when pushed there ([per](https://github.com/Marri/glowfic/pull/449/files#diff-1d98f0039bbc6df92adee882ed99d993L35))
- [x] ~~Check we don't need actionmailer not to deliver in development ([per](https://github.com/Marri/glowfic/pull/449/files#diff-a2f09e1db7a47eb67c5849478a97f3d7L19))~~

For review:
- [x] Pick out changes that are compatible with Rails 4.2 to make this less huge as a PR
- [x] Pick out changes that are only about deprecation warnings to make this less huge as a PR (test locally since Travis output is not all visible)


config/spring.rb
```diff
@@ -0,0 +1,6 @@
+%w(
+  .ruby-version
+  .rbenv-vars
+  tmp/restart.txt
+  tmp/caching-dev.txt
+).each { |path| Spring.watch(path) }
```
Throne3d: We don't actually use `spring`, this is just part of the default configuration setup.

config/puma.rb
```diff
@@ -1,19 +1,49 @@
...
-    config = ActiveRecord::Base.configurations[Rails.env] ||
-                Rails.application.config.database_configuration[Rails.env]
-    config['pool'] = ENV['RAILS_MAX_THREADS'] || 5
-    ActiveRecord::Base.establish_connection(config)
-  end
+  ActiveRecord::Base.establish_connection if defined?(ActiveRecord)
 end
+
+# Allow puma to be restarted by `rails restart` command.
+plugin :tmp_restart
```
Throne3d: This file is mostly reordering, except for the removal of `rackup DefaultRackup` (not required in later Puma versions), and the modification of the `on_worker_boot` section (that appears to be for an old version of Rails and isn't part of the default config?).

config/initializers/new_framework_defaults.rb
```diff
@@ -0,0 +1,25 @@
...
+
+# Make Ruby 2.4 preserve the timezone of the receiver when calling `to_time`.
+# Previous versions had false.
+ActiveSupport.to_time_preserves_timezone = false
+
+# Require `belongs_to` associations by default. Previous versions had false.
+Rails.application.config.active_record.belongs_to_required_by_default = false
+
+# Do not halt callback chains when a callback returns false. Previous versions had true.
+ActiveSupport.halt_callback_chains_on_return_false = true
```
Throne3d: These can be changed as necessary when we're more confident about moving to proper Rails 5.0 defaults.

config/initializers/application_controller_renderer.rb
```diff
@@ -0,0 +1,8 @@
+# Be sure to restart your server when you modify this file.
+
+# ActiveSupport::Reloader.to_prepare do
+#   ApplicationController.renderer.defaults.merge!(
+#     http_host: 'example.org',
+#     https: false
+#   )
+# end
```
Throne3d: I'm not sure what this is about, but it's part of the standard configuration setup for Rails 5.0 (and it's all commented out).

config/environments/production.rb
```diff
@@ -32,16 +25,20 @@
...
 
-  # Asset digests allow you to set far-future HTTP expiration dates on all assets,
-  # yet still be able to expire them through the digest params.
-  config.assets.digest = true
-
   # `config.assets.precompile` and `config.assets.version` have moved to config/initializers/assets.rb
 
+  # Enable serving of images, stylesheets, and JavaScripts from an asset server.
+  config.action_controller.asset_host = ENV['ASSET_HOST']
+
```
Throne3d: File updated per default config except for the clear sections we've modified (e.g. `ASSET_HOST`)

config/environments/development.rb
```diff
@@ -9,45 +9,46 @@
...
-  config.assets.raise_runtime_errors = true
+  # Suppress logger output for asset requests.
+  config.assets.quiet = true
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+  # Use an evented file watcher to asynchronously detect changes in source code,
+  # routes, locales, etc. This feature depends on the listen gem.
+  # config.file_watcher = ActiveSupport::EventedFileUpdateChecker
```
Throne3d: Do we want this?

config/environments/development.rb
```diff
@@ -9,45 +9,46 @@
...
-  # }
+
+  config.action_mailer.perform_caching = false
 
   # Print deprecation notices to the Rails logger.
   config.active_support.deprecation = :log
 
   # Raise an error on page load if there are pending migrations.
-  # config.active_record.migration_error = :page_load
+  config.active_record.migration_error = :page_load
```
Throne3d: This is default and we probably want it?

config/cable.yml
```diff
@@ -0,0 +1,9 @@
+development:
+  adapter: async
+
+test:
+  adapter: async
+
+production:
+  adapter: redis
+  url: redis://localhost:6379/1
```
Throne3d: ActionCable got merged into Rails proper for 5.0, and seems to allow WebSocket connections, so if we want to set up to have some realtime notification system we mmmight use this? (Default config.)

config/application.rb
```diff
@@ -72,9 +72,6 @@ class Application < Rails::Application
     # config.i18n.load_path += Dir[Rails.root.join('my', 'locales', '*.{rb,yml}').to_s]
     # config.i18n.default_locale = :de
 
-    # Do not swallow errors in after_commit/after_rollback callbacks.
-    config.active_record.raise_in_transactional_callbacks = true
-
```
Throne3d: Deprecated, ignored, and keeps the same functionality we had before.

app/controllers/api/api_controller.rb
```diff
@@ -1,4 +1,5 @@
 class Api::ApiController < ActionController::Base
+  include Rails::Pagination
```
Throne3d: `paginate` now only automatically includes itself in `ActionController::API` – when I tried swapping the API controllers to that, they were unhappy about how we included Authentication and various other things (needs cookies, and needs to provide the `helper_method` function, and various other things). It says for applications where that's annoying, you can just include it manually like this.

Gemfile
```diff
@@ -64,6 +65,7 @@ group :test do
   gem 'codeclimate-test-reporter', '~> 1.0.0'
   gem 'factory_girl_rails'
   gem 'json'
+  gem 'rails-controller-testing'
```
Throne3d: For access to `response` and `assigns` – got extracted into a gem with Rails 5.0

Gemfile
```diff
@@ -4,7 +4,7 @@ ruby '2.4.1'
 
 gem 'api-pagination'
 gem 'apipie-rails'
-gem 'audited-activerecord', '~> 4.0'
+gem 'audited', '~> 4.5'
```
Throne3d: Looks like `audited` is the new version of `audited-activerecord`, and the one that works with Rails 5.0

config/environments/production.rb
```diff
@@ -32,16 +25,20 @@
   # Do not fallback to assets pipeline if a precompiled asset is missed.
   config.assets.compile = false
 
-  # Asset digests allow you to set far-future HTTP expiration dates on all assets,
-  # yet still be able to expire them through the digest params.
-  config.assets.digest = true
-
```
Throne3d: Is not part of the default config – I plan to test this on staging to check if we still have digests?

config/environments/production.rb
```diff
@@ -14,16 +14,9 @@
...
-  # Enable Rack::Cache to put a simple HTTP cache in front of your application
-  # Add `rack-cache` to your Gemfile before enabling this.
-  # For large-scale production use, consider using a caching reverse proxy like
-  # NGINX, varnish or squid.
-  # config.action_dispatch.rack_cache = true
-
   # Disable serving static files from the `/public` folder by default since
   # Apache or NGINX already handles this.
-  config.serve_static_files = ENV['RAILS_SERVE_STATIC_FILES'].present?
-  config.static_cache_control = 'public, max-age=31536000'
```
Throne3d: Do we need this? We could probably get it again by copying the config from `test.rb` below but I think it caused issues for me when I modified TinyMCE's `tinymce.css` – that didn't change in either of my browsers until I manually loaded *it* and cleared the cache on that page, instead of just Ctrl+Shift+R on the main site. (Alternatively, should we look into putting `tinymce.css` into the assets folder, and getting TinyMCE to use the digest properly?)

config/environments/production.rb
```diff
@@ -77,6 +73,16 @@
...
+  # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new 'app-name')
+
+  if ENV["RAILS_LOG_TO_STDOUT"].present?
+    logger           = ActiveSupport::Logger.new(STDOUT)
+    logger.formatter = config.log_formatter
+    config.logger = ActiveSupport::TaggedLogging.new(logger)
+  end
+
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
```
Throne3d: File updated per default config (except e.g. `ASSET_HOST`)

db/structure.sql
```diff
@@ -1893,151 +1913,80 @@ CREATE INDEX user_index ON audits USING btree (user_id, user_type);
...
+('20170814042150'),
+('20170819222420'),
+('20170821210336'),
+('20170826211901'),
+('20170828144608'),
+('20170907180029'),
+('20170911235423'),
+('20170914191425');
 
-INSERT INTO schema_migrations (version) VALUES ('20170914191425');
```
Throne3d: Automatic changes

spec/spec_helper.rb
```diff
@@ -98,7 +95,7 @@
     # Use the documentation formatter for detailed output,
     # unless a formatter has already been configured
     # (e.g. via a command-line flag).
-    config.default_formatter = 'doc'
+    config.default_formatter = "doc"
   end
 =end
```
Throne3d: New defaults

spec/spec_helper.rb
```diff
@@ -141,7 +138,7 @@
 RSpec::Matchers.define_negated_matcher :not_change, :change
 
 # Monkey patches the controller response objects to return JSON
-module ActionController
+module ActionDispatch
```
Throne3d: To fix `response.json` monkey patch

config/environments/production.rb
Throne3d: Digests seem to be a new default (tested on staging).

config/environments/production.rb
Marri: I'm pretty confident we do, yes, because omg will my AWS bill get horrible if this isn't cached. Sounds like we need tinymce.css in the asset folder (and like, probably more things in public/ as well tbh?)

Throne3d: I was under the impression this was only for assets in the 'public' folder but I might well be wrong.

I'll add it back in anyway. (But yeah, if we have CSS or JS in public we should probably move them, I'll make an issue.)

config/cable.yml
Marri: What is ActionCable and do we actually use it?

config/environments/development.rb
Marri: Do we still have the ability to start the server, modify a file, and have the change get auto-picked up?

config/environments/production.rb
```diff
@@ -32,16 +28,20 @@
...
+  config.action_controller.asset_host = ENV['ASSET_HOST']
+
   # Specifies the header that your server uses for sending files.
   # config.action_dispatch.x_sendfile_header = 'X-Sendfile' # for Apache
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
+  # Mount Action Cable outside main process or domain
+  # config.action_cable.mount_path = nil
+  # config.action_cable.url = 'wss://example.com/cable'
+  # config.action_cable.allowed_request_origins = [ 'http://example.com', /http:\/\/example.*/ ]
```
Marri: ???

config/puma.rb
```diff
@@ -1,19 +1,49 @@
-workers Integer(ENV['WEB_CONCURRENCY'] || 2)
-threads_count = Integer(ENV['RAILS_MAX_THREADS'] || 5)
+# Puma can serve each request in a thread from an internal thread pool.
+# The `threads` method setting takes two numbers a minimum and maximum.
+# Any libraries that use thread pools should be configured to match
+# the maximum value specified for Puma. Default is set to 5 threads for minimum
+# and maximum, this matches the default thread size of Active Record.
+#
+threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i
```
Marri: Why is this a block? .fetch(key, default) should work?

config/puma.rb
Marri: Context https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server#on-worker-boot

config/cable.yml
Throne3d: It's apparently a thing to do with WebSockets (see [here](https://github.com/rails/rails/tree/master/actioncable)), and not as far as I know, but this is part of the default config and I assume we'd have to do something complicated to remove actioncable as a gem and make this config no longer necessary.

config/environments/production.rb
Throne3d: this is all commented out and part of the default? it's actioncable and we don't in fact want to mount the websocket handler outside the main process (or maybe even at all) so i left it there, commented

config/puma.rb
Throne3d: I'm not sure, but this is part of the updated default puma config

config/environments/development.rb
Throne3d: Pretty sure we do – I'll check.

config/puma.rb
Throne3d: Added a `pool` attribute to `config/database.yml` per documentation in link: <https://github.com/Marri/glowfic/pull/449/files#diff-e31bdf70b15c8f84344c332efe06900d>

PR #451, opened 2017-09-15
Use labeled parameters instead of positional in controller tests

Throne3d:
About to be deprecated in Rails 5.0, so might as well fix it now.


Throne3d:
… Not available on Rails 4.2. Sure. Oops.

PR #453, opened 2017-09-15
Specify Rails versions in migrations to avoid future deprecation warning

Throne3d:
Accepted values: 4.2 (, 5.0)


Throne3d:
Not permitted on Rails 4.2.

PR #457, opened 2017-09-15
Fix deprecation warnings for Rails 5.0

Throne3d:
Builds on #449.

Deprecation:
(*Check the many deprecation warnings, see #448 for more info:*)
- [x] `ActiveRecord::Base.raise_in_transactional_callbacks=` (set to true, now forced true and the config does nothing; removed)
- [ ] Parameters are now objects instead of hashes and comparing equality is deprecated set to be removed in Rails 5.1
- [ ] `alias_method_chain` deprecated (in Presentable)
- [x] `before_filter`, `around_filter`, `after_filter` in basically all the controllers
- [x] Positional parameters in tests (`get :show, id: 1` vs `get :show, params: { id: 1 }`)
- [x] Rails target versions in migration superclasses (`< ActiveRecord::Migration[4.2]`) #

### DIFF
--- a/db/migrate/20150413062555_create_users.rb
+++ b/db/migrate/20150413062555_create_users.rb
@@ -1,4 +1,4 @@
-class CreateUsers < ActiveRecord::Migration
+class CreateUsers < ActiveRecord::Migration[4.2]
   def change
     create_table :users do |t|
       t.string :username, :null => false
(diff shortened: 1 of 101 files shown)


Throne3d:
I'll look at the rest in future PRs.

PR #467, opened 2017-09-19
Simplify and modularize posts#editor javascript

Throne3d:
It was previously quite complicated to keep track of all the individual components (and the linting was complaining about the complexity of various methods), so I've tried to extract the sections into appropriately-labelled functions (by function, instead of clustering all the chronologically-related things together), as well as restructuring other bits of code and logic to make this more straightforwards.

Risky because it's hard to test. I'm gonna put it on staging and ask some people to poke at it.


Throne3d:
Swapping between characters without aliases and with 0, 1 and 2 galleries and (no, a) default works fine – adding aliases into the mix also seems to work but has been tested perhaps less thoroughly.

Accordingly I've merged this, relatively happy with it.

PR #475, opened 2017-09-22
Convert jobs to ActiveJob

Throne3d:
- [ ] Have Rails serialize the models itself? (It can do that now; we don't need to pass `.id` and get it to retrieve the model.)
- [ ] Use the `async` queue adapter in development instead of Resque? (`heroku local` runs Resque anyway, so it might not be necessary, but with `async` presumably it also gets run on `rails s`.)


Throne3d:
Fixes part of #476.

Gemfile.lock
```diff
@@ -442,7 +439,7 @@ DEPENDENCIES
   redis (~> 3.0)
   redis-rails
   resque
-  resque-retry
+  resque-scheduler
```
Throne3d: Allows ActiveJob to use `enqueue_at` so we can schedule jobs for a specific time. (See <https://github.com/rails/rails/blob/6a8e7fd8d8cbfa335899153d455082cde906a579/activejob/lib/active_job/queue_adapters/resque_adapter.rb#L34>.)

Gemfile
```diff
@@ -27,7 +27,7 @@ gem 'redis-rails'
 gem 'redis', '~> 3.0'
 gem 'resque'
 gem 'resque_mailer'
-gem 'resque-retry'
+gem 'resque-scheduler'
```
Throne3d: Allows ActiveJob to use `enqueue_at` so we can schedule jobs for a specific time. (See <https://github.com/rails/rails/blob/6a8e7fd8d8cbfa335899153d455082cde906a579/activejob/lib/active_job/queue_adapters/resque_adapter.rb#L34>.)

app/jobs/application_job.rb
```diff
@@ -0,0 +1,24 @@
+require 'resque/errors'
+class ApplicationJob < ActiveJob::Base
```
Throne3d: Mostly a renamed BaseJob

config/environments/development.rb
```diff
@@ -26,6 +26,11 @@
     config.cache_store = :null_store
   end
 
+  # Use a real queuing backend for Active Job (and separate queues per environment)
+  config.active_job.queue_adapter     = :resque
+  # config.active_job.queue_name_prefix = "glowfic_#{Rails.env}"
+  config.action_mailer.perform_caching = false
+
```
Throne3d: Development should also queue to Resque (otherwise I think it runs jobs inline).

config/environments/test.rb
```diff
@@ -40,6 +40,8 @@
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
 
+  config.active_job.queue_adapter = :test
+
```
Throne3d: (Otherwise assertions in tests fail)

config/environments/development.rb
Throne3d: We might want to look into `async` as `queue_adapter` for development: <http://api.rubyonrails.org/classes/ActiveJob/QueueAdapters.html#module-ActiveJob::QueueAdapters-label-Async+and+Inline+Queue+Adapters>

Throne3d:
Also to look into: have Rails serialize the models itself? (It can do that now; we don't need to pass `.id` and get it to retrieve the model.)

Throne3d:
Jobs queue just fine on staging as is.

Issue #477, opened 2017-09-23
Clean up the CSS


Throne3d:
- [ ] SCSS is better than CSS because it supports variables and nesting; use it
- [ ] Start using variables for shared colors of the palette; where necessary for layouts break these down into sub-divided variables that in the default layout refer to one single Standard Color
- [ ] Likewise break font sizes and so on down into variables
- [ ] Use `em` for font sizes? Also use relative font sizes in `@media` queries
- [ ] Use CSS instead of manually modifying the HTML for the paginators

Throne3d:
See also #857

Marri:
Used [this cleaner](https://medium.com/carwow-product-engineering/clean-your-css-from-unused-selectors-353f0a65ca88), it was naive but it did the job decently well.

If we install it, it has some gotchas around we might want to fix:
- not finding unused IDs
- not finding dynamic class names (didn't seem to affect much)
- not catching ui-* or select2-* classes are actually used (by JQuery UI and Select2 respectively)
- outputting only class names into classes_found.txt and just printing the class>count results; ideally that list goes into a file as well

PR #478, opened 2017-09-23
Improve CSS

Throne3d:
In response to #477. PR opened so I can keep track of to-dos that crop up and weren't planned in the related issue.

- [x] Check `.breadcrumbs` display properly in each theme now they've been decoupled from `.subber` (also look for subbers outside `#content`?? except I'm pretty sure everything is inside `#content` except layout things)
- [x] Relatedly: `.sub` (paginator seems to use this, despite not being a `th` like all the other occurrences)
- [x] Make sure the relevant of `th`, `sub` and `subber` work properly on the various layouts
- [x] Separate `subber` and the thing at the end of `form-table`s (since they use different colors in some layouts and are semantically distinct)
- [x] Combine `list-header` and `content-header` (ex-`post-header`)
- [x] Put the padding into `content-header` instead of using `padding-15`
- [x] Put the padding into `post-subheader` instead of using `padding-15`

Tasks:
- [x] SCSS is better than CSS because it supports variables and nesting; use it
- [x] Start using variables for shared colors of the palette; where necessary for layouts break these down into sub-divided variables that in the default layout refer to one single Standard Color
- [ ] Likewise break font sizes and so on down into variables
- [ ] Use `em` for font sizes? Also use relative font sizes in `@media` queries
- [ ] Use CSS instead of manually modifying the HTML for the paginators
- [x] The "no icon" thing in the header should display with the same formatting as if you have an icon (there are slight differences at present)
- [x] The header should display identically between Firefox and Chrome if possible
- [ ] Links in the nav-bottom header are hard to read on default layout
- [ ] Apply more the principle "style by *what* it is, not *where* it is"
- [ ] Combine `th` and `head` if possible?
- [ ] Get rid of the magic numbers still left in `replies.scss`


Throne3d:
It's at a mergeable stage though there are still things left to do; I'll merge it now and make a note for the remaining bits.

PR #493, opened 2017-10-01
Improve the paginator display (mainly CSS & Javascript)

Throne3d:
- Center the paginator when it wraps to a new line (as it seems to without the `clear` take into account the position of the left float, even though it's on a separate line; it's weird)
- Hide paginator links when the screen narrows, so the paginator doesn't wrap over several lines
- Use the mobile paginator at the 400px breakpoint, instead of based on browser detection, and relatedly also use it when the paginator wraps over several lines and no more links can be feasibly hidden

Other considered solutions were to modify font sizes or padding, but since there are often upwards of ten links on the single line at any one time (the current page, the four to either side, the two at either end of the paginator – this adds to 13, plus the "‹ Previous" and "Next ›" links, which gets bad especially with three-digit page numbers), I considered that it would not be the best fix; it would likely involve reducing the hitbox of the visible links (making them gradually more and more unuseful), whereas the solution of hiding links is more graceful from a user perspective. (Admittedly it is ugly that we duplicate the logic to calculate visible page links.)

### DIFF
--- a/app/assets/javascripts/posts/show.js
+++ b/app/assets/javascripts/posts/show.js
@@ -54,10 +54,10 @@ $(document).ready(function() {
   paginators.each(function() { reflowPaginator(this); });
   var resizeDebounce = null;
   $(window).resize(function() {
-    clearTimeout(resizeDebounce);
-    resizeDebounce = setTimeout(function() {
+    window.cancelAnimationFrame(resizeDebounce);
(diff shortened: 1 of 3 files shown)


Throne3d:
I've pushed it to staging and am looking for some more feedback before I merge this, since I may have broken something with my latest round of changes; people do seem generally in favor of it.

PR #496, opened 2017-10-01
Replace character.setting with the Setting model

Marri:
Remaining:
- [x] Don't duplicate the Javascript from posts/search, make it somehow common
- [x] Make sure every use of partials on characters/index.haml uses `includes(:settings)`

### DIFF
--- a/app/controllers/characters_controller.rb
+++ b/app/controllers/characters_controller.rb
@@ -151,7 +151,7 @@ class CharactersController < ApplicationController
       name = alt.name
       name += ' | ' + alt.screenname if alt.screenname
       name += ' | ' + alt.template_name if alt.template_name
-      name += ' | ' + alt.setting if alt.setting
+      name += ' | ' + alt.settings.pluck(:name).join(' & ') if alt.settings.present?
       [name, alt.id]
     end
     @alt = @alts.first
@@ -310,6 +310,9 @@ class CharactersController < ApplicationController
   def build_tags
     @gallery_groups = @character.gallery_groups.order('character_tags.id asc') if @character.try(:persisted?)
     @gallery_groups ||= @character.try(:gallery_groups) || []
+
+    @settings = @character.settings.order('character_tags.id asc') if @character.try(:persisted?)
+    @settings ||= @character.try(:settings) || []
   end
 
   def build_template
@@ -319,6 +322,18 @@ class CharactersController < ApplicationController
   end
 
   def character_params
-    params.fetch(:character, {}).permit(:default_icon_id, :name, :template_name, :screenname, :setting, :template_id, :pb, :description, ungrouped_gallery_ids: [], gallery_group_ids: [], template_attributes: [:name, :id])
+    params.fetch(:character, {}).permit(
+      :default_icon_id,
+      :name,
+      :template_name,
+      :screenname,
+      :template_id,
+      :pb,
+      :description,
+      setting_ids: [],
+      ungrouped_gallery_ids: [],
+      gallery_group_ids: [],
+      template_attributes: [:name, :id]
+    )
   end
 end
--- a/app/views/characters/_editor.haml
+++ b/app/views/characters/_editor.haml
@@ -9,9 +9,7 @@
   %td{class: cycle('even', 'odd')}= f.text_field :screenname, placeholder: "Screen Name", class: 'text'
 %tr
   %th.sub Setting
-  %td{class: cycle('even', 'odd')}= f.text_field :setting, placeholder: "Setting", class: 'text'
-  /%td{class: cycle('even', 'odd')}= f.collection_select(:setting_ids,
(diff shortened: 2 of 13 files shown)


app/views/characters/_editor.haml
```diff
@@ -9,9 +9,7 @@
   %td{class: cycle('even', 'odd')}= f.text_field :screenname, placeholder: "Screen Name", class: 'text'
 %tr
   %th.sub Setting
-  %td{class: cycle('even', 'odd')}= f.text_field :setting, placeholder: "Setting", class: 'text'
-  /%td{class: cycle('even', 'odd')}= f.collection_select(:setting_ids,
-  /  Setting.all, :id, :name, {selected: @character.settings.pluck(:id)}, {multiple: true})
+  %td{class: cycle('even', 'odd')}= f.collection_select(:setting_ids, Setting.all, :id, :name, {selected: @character.settings.pluck(:id)}, {multiple: true})
```
Throne3d: `@character.settings` not `Setting.all` since we don't need them all loaded and it hits the endpoint?

app/controllers/characters_controller.rb
```diff
@@ -319,6 +319,6 @@ def build_template
   end
 
   def character_params
-    params.fetch(:character, {}).permit(:default_icon_id, :name, :template_name, :screenname, :setting, :template_id, :pb, :description, ungrouped_gallery_ids: [], gallery_group_ids: [], template_attributes: [:name, :id])
+    params.fetch(:character, {}).permit(:default_icon_id, :name, :template_name, :screenname, :template_id, :pb, :description, ungrouped_gallery_ids: [], setting_ids: [], gallery_group_ids: [], template_attributes: [:name, :id])
```
Throne3d: This is getting long, would suggest putting the individual attributes on new lines (not a blocker or anything, minor)

db/migrate/20171001035221_move_setting_to_tags.rb
```diff
@@ -0,0 +1,23 @@
+class MoveSettingToTags < ActiveRecord::Migration[5.0]
+  def up
+    print "\n"
+    Character.where.not(setting: nil).each do |character|
```
Throne3d: I think `find_each` is preferred for large sets

db/migrate/20171001035221_move_setting_to_tags.rb
```diff
@@ -0,0 +1,23 @@
...
+      character.settings << setting
+      print "#{character.name} migrated\n"
+    end
+    print "\n"
+    remove_column :characters, :setting
+  end
+  
+  def down
+    add_column :characters, :setting, :text
+    Setting.all.each do |setting|
```
Throne3d: again `find_each`

db/migrate/20171001035221_move_setting_to_tags.rb
```diff
@@ -0,0 +1,23 @@
+class MoveSettingToTags < ActiveRecord::Migration[5.0]
+  def up
+    print "\n"
+    Character.where.not(setting: nil).each do |character|
+      setting = Setting.where(name: character.setting).first
+      setting ||= Setting.create!(user: character.user, name: character.setting)
+      character.settings << setting
+      print "#{character.name} migrated\n"
```
Throne3d: This looks like it's going to clutter the output A Lot and so far migrations don't do that (at least not a ton) – could we output `\r(#{index} / #{total_count})` instead, so it looks like a progress thing? (`\r` resets the line instead of doing a linebreak, then you'd presumably fetch and save `total_count` up the top and increment `index` as you go along, so it shows `(1/250)` then `(2/250)`…)

db/migrate/20171001035221_move_setting_to_tags.rb
```diff
@@ -0,0 +1,23 @@
+class MoveSettingToTags < ActiveRecord::Migration[5.0]
+  def up
+    print "\n"
+    Character.where.not(setting: nil).each do |character|
+      setting = Setting.where(name: character.setting).first
+      setting ||= Setting.create!(user: character.user, name: character.setting)
```
Throne3d: Can we cache settings by name here for efficiency purposes? (This is looking to be a lot of `where`s)

db/migrate/20171001035221_move_setting_to_tags.rb
```diff
@@ -0,0 +1,23 @@
...
+      character.settings << setting
```
Throne3d: Also I'm wary of adding associations without saving – there are _some_ circumstances (which I think do not apply here) where it doesn't save the change. (`.update_attributes(settings: [setting])` should work the same since the characters only have one setting?)

db/migrate/20171001035221_move_setting_to_tags.rb
Marri: Even after 100% migrating we won't have enough Settings to exceed a single batch, but I'll use it above (which will take 3 batches)

Issue #295, opened 2017-08-15
Indexes and collections, distinct from continuities


Throne3d:
Requested [here](http://alicorn.elcenia.com/board/viewtopic.php?f=12&t=459&start=2310#p34126) by Alicorn, something that "worked mechanically exactly like a continuity except a) not being called that ("an index"?) and b) allowing threads to be in more than one of them. Possibly also c) the ability to add threads to them without being involved in creating the thread".

dhouck:
This way the Particularly Good Threads section on Alicornutopia could be moved to a Constellation index.

Unfortunately, I don’t think it would be reasonable to make it smart enough to autogenerate author indexes, since those often contain information not available in other metadata.

Throne3d:
We'd need to have freely-editable indexes for that, I think? (Unless there's some meta post to suggest them in.) It'd also need some way to attach a little index-specific summary to the post, which current continuities don't have.

Throne3d:
Indexes were added in #519

PR #519, opened 2017-10-28
Indexes part 1

Marri:
Skeleton code for indexes. Includes absolutely no UI, for safety of migration running.

Addresses first part of #295

### DIFF
--- a/db/migrate/20171027225408_create_indexes.rb
+++ b/db/migrate/20171027225408_create_indexes.rb
@@ -11,21 +11,21 @@ class CreateIndexes < ActiveRecord::Migration[5.0]
     add_index :indexes, :user_id
 
     create_table :index_sections do |t|
-    	t.integer :index_id, null: false
-    	t.citext :name, null: false
-    	t.text :description
-    	t.integer :section_order
-    	t.timestamps null: false
+      t.integer :index_id, null: false
+      t.citext :name, null: false
+      t.text :description
+      t.integer :section_order
+      t.timestamps null: false
     end
     add_index :index_sections, :index_id
 
     create_table :index_posts do |t|
-    	t.integer :post_id, null: false
-    	t.integer :index_id, null: false
+      t.integer :post_id, null: false
+      t.integer :index_id, null: false
       t.integer :index_section_id
       t.text :description
-    	t.integer :section_order
-    	t.timestamps null: false
+      t.integer :section_order
+      t.timestamps null: false
     end
     add_index :index_posts, :index_id
     add_index :index_posts, :post_id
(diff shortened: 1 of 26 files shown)


db/migrate/20171027225408_create_indexes.rb
```diff
@@ -0,0 +1,33 @@
+class CreateIndexes < ActiveRecord::Migration[5.0]
+  def change
+    create_table :indexes do |t|
+      t.integer :user_id, null: false
+      t.citext :name, null: false
+      t.text :description
+      t.integer :privacy, null: false, default: 0
+      t.boolean :open_to_anyone, null: false, default: false
```
Throne3d: I assume this (`open_to_anyone`) is about editing privileges instead of just a regular privacy option? I might go for something like `community_editable` to stop it from being confused?

db/migrate/20171027225408_create_indexes.rb
```diff
@@ -0,0 +1,33 @@
...
+    	t.integer :section_order
+    	t.timestamps null: false
+    end
+    add_index :index_sections, :index_id
+
+    create_table :index_posts do |t|
+    	t.integer :post_id, null: false
+    	t.integer :index_id, null: false
+      t.integer :index_section_id
+      t.text :description
```
Throne3d: indentation is weird

spec/models/index_section_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    expect(section2.section_order).to eq(2)
+    section3 = create(:index_section, index: index)
+    expect(section3.section_order).to eq(3)
+    section1.index = create(:index)
+    section1.save
+    expect(section0.reload.section_order).to eq(0)
+    expect(section2.reload.section_order).to eq(1)
+    expect(section3.reload.section_order).to eq(2)
+  end
+end
```
Throne3d: Should we copy similar tests for `IndexPosts` across?

db/migrate/20171027225408_create_indexes.rb
Marri: Yeah, it's about "who can add things to this index", I had named it this to be consistent with boards (and in case I ever made it a non-boolean and built in an IndexAuthors structure, so that the name wouldn't change) but if you think it's confusing I can make it community_editable, sure

Throne3d: … Oh, good point, `open_to_anyone` is about people who can add to the board on a regular board anyway. This should be fine.

spec/models/index_section_spec.rb
Throne3d: (Done.)

db/migrate/20171027225408_create_indexes.rb
```diff
@@ -11,23 +11,23 @@ def change
...
       t.text :description
-    	t.integer :section_order
-    	t.timestamps null: false
+      t.integer :section_order
+      t.timestamps null: false
     end
     add_index :index_posts, :index_id
     add_index :index_posts, :post_id
-  end
+    end
```
Throne3d: Indentation is weird again? :P

PR #533, opened 2017-11-11
No more Canons

Marri:
aljsdflkjdflskjdflskjfldks

Remove canon, make settings link to each other instead, sulk.


app/models/canon.rb
```diff
@@ -1,9 +1,2 @@
 class Canon < Tag
```
Marri: Can't delete the class till the migration has run; similarly, cannot update CharacterTag or Character (could hack around TagTag but need the non-duplicating logic for characters)

Throne3d: Can't you just place the class into the migration itself, since that's the last time we're gonna use it? (That way the migration will work even after we remove the class, too.)

Throne3d: (Idk around CharacterTag and Character.)

PR #534, opened 2017-11-11
Tags

Marri:
- Removes remaining Canon code
- Deletes Setting<>Setting joins when a Setting is deleted
- Spaces the sections in Tag#show properly
- Fixes the confirmation prompt upon deleting a Tag
- Locks deletion privileges to only owners and admins
- Preserves view and filters on tags#index page when deleting a tag
- Allow owners to rename tags
- Tag#merge_with admin method now updates Setting<>Setting joins correctly

### DIFF
--- a/db/migrate/20171111163658_canon_refactor.rb
+++ b/db/migrate/20171111163658_canon_refactor.rb
@@ -1,6 +1,6 @@
 class CanonRefactor < ActiveRecord::Migration[5.0]
   def up
-    Canon.all.each do |canon|
+    Tag.where(type: "Canon").each do |canon|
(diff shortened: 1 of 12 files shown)


db/migrate/20171111163658_canon_refactor.rb
```diff
@@ -1,6 +1,6 @@
 class CanonRefactor < ActiveRecord::Migration[5.0]
   def up
-    Canon.all.each do |canon|
+    Tag.where(type: "Canon").each do |canon|
```
Throne3d: <3

PR #545, opened 2017-11-13
Add moderator notes for editing replies

Marri:
Audits for moderators part one: editing replies

### DIFF
--- a/app/assets/javascripts/posts/editor.js
+++ b/app/assets/javascripts/posts/editor.js
@@ -425,7 +425,7 @@ function getAndSetCharacterData(characterId, options) {
 
   // Handle special case where just setting to your base account
   if (characterId === '') {
-    var avatar = gon.current_user.avatar;
+    var avatar = gon.editor_user.avatar;
     var data = {aliases: [], galleries: []};
     if (avatar) {
       data.default = avatar;
--- a/app/assets/javascripts/posts/editor.js
+++ b/app/assets/javascripts/posts/editor.js
@@ -12,9 +12,7 @@ $(document).ready(function() {
     formButtons.not(this).data('disable-with', '').prop('disabled', true);
   });
 
-  if ($("#post-editor .view-button").length === 0) return; // Skip if there is no writable editor (no RTF/HTML buttons)
-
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/posts/editor.js
```diff
@@ -12,9 +12,7 @@ $(document).ready(function() {
     formButtons.not(this).data('disable-with', '').prop('disabled', true);
   });
 
-  if ($("#post-editor .view-button").length === 0) return; // Skip if there is no writable editor (no RTF/HTML buttons)
-
-  setupWritableEditor();
+  if ($("#post-editor .view-button").length > 0) setupWritableEditor();
```
Marri: Unrelated, but stumbled across it; should be the same functionality, but less of a gotcha in future if someone adds code below the return.

app/controllers/replies_controller.rb
```diff
@@ -166,8 +166,22 @@ def update
...
-    @reply.skip_post_update = true unless @reply.post.last_reply_id == @reply.id
-    @reply.save
+    if current_user.id != @reply.user_id && @reply.audit_comment.blank?
+      flash[:error] = "You must provide a reason for your moderator edit."
+      editor_setup
+      render action: :edit and return
+    end
+
+    @reply.audit_comment = nil if @reply.changes.empty? # don't save an audit for a note and no changes
+    @reply.skip_post_update = true unless @reply.post.try(:last_reply_id) == @reply.id
```
Marri: Also not moderation relevant, but added the missing reply save validation check here, along with a test; since my test checks by setting post to `nil`, also needed a `try()`.

app/controllers/writable_controller.rb
```diff
@@ -119,7 +120,7 @@ def display_warnings?
 
   def editor_setup
     use_javascript('posts/editor')
-    build_template_groups
+    build_template_groups(@reply.try(:user) || @post.try(:user))
```
Marri: `@reply` must come first here because most reply controller methods set `@post` as well, and the user isn't necessarily the same.

PR #553, opened 2017-11-16
Show template

### LINKED ISSUE #535: Show character template on Settings#show page
We have a list of characters tagged with the setting; include their template.

teceler:
fixes #535, fixes the same problem on un-template-grouped character page, and adds a test for characters on setting tags.

### DIFF
--- a/spec/controllers/tags_controller_spec.rb
+++ b/spec/controllers/tags_controller_spec.rb
@@ -93,7 +93,7 @@ RSpec.describe TagsController do
         expect(assigns(:galleries)).to match_array([gallery])
       end
 
-      it "succeeds with valid character tag" do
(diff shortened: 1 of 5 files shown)


spec/controllers/tags_controller_spec.rb
```diff
@@ -93,22 +93,39 @@ def create_tags
         expect(assigns(:galleries)).to match_array([gallery])
       end
 
-      it "succeeds with valid character tag" do
+      it "gallery group succeeds with valid character tag" do
```
Throne3d: To be more correct, I think this "gallery group" bit should be pulled out into a separate context. (Currently it's not grammatical – "it gallery group succeeds with valid character tag", whereas "context gallery group, it succeeds with valid character tag" is more so.)

PR #564, opened 2017-11-17
Clean up search HAML & darken search inputs on dark modes

Throne3d:
Builds on #561.

- Search table should not have ID "signup"
- TD should not be nested inside another TD
- Input labels should be labels
- `username` should not be an ambiguous element ID

### DIFF
--- a/app/views/characters/search.haml
+++ b/app/views/characters/search.haml
@@ -14,15 +14,15 @@
       %tr
         %td.width-300.padding-10.vtop.dark
           = form_tag search_characters_path, :method => :get do
-            %table#signup
+            %table
               %tr
-                %td Author
+                %td= label_tag :author_id, 'Author'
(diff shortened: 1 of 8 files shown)


app/views/characters/search.haml
```diff
@@ -14,15 +14,15 @@
       %tr
         %td.width-300.padding-10.vtop.dark
           = form_tag search_characters_path, :method => :get do
-            %table#signup
+            %table.search-form
```
Marri: I think this was for styling reasons, did formatting change too?

Throne3d: The only style that applies to `#signup` is `#signup .sub { width: 90px; }` – I'll look if there are any `.sub`s but I didn't notice a change, at least.

Throne3d: Looking at the differences between master and this branch doesn't show any visible change (except for a darkened input box). It looks like `#signup` wasn't necessary.

PR #566, opened 2017-11-18
Add a webhook for Travis CI

Throne3d:


Throne3d:
Doesn't work since Travis and Discord don't use a compatible format. There are various workarounds on Google but a lot of them seem to be downloading a script and executing it, which I thought – "maybe not" to (at least for now).

Issue #476, opened 2017-09-22
Address changes after Rails 5.0


Throne3d:
After #448 there are some things still to be addressed – this is a continuation issue.

Deprecation:
- [x] `ActiveRecord::Base.raise_in_transactional_callbacks=` (set to true, now forced true and the config does nothing; removed)
- [x] Parameters are now objects instead of hashes and comparing equality is deprecated set to be removed in Rails 5.1
- [x] `alias_method_chain` deprecated (in Presentable)
- [x] `before_filter`, `around_filter`, `after_filter` in basically all the controllers
- [x] Positional parameters in tests (`get :show, id: 1` vs `get :show, params: { id: 1 }`)
- [x] Rails target versions in migration superclasses (`< ActiveRecord::Migration[4.2]`)
- [x] `uniq!` → `distinct!` (`app/controllers/messages_controller.rb:14`)

Later:
- [x] Fix API tests to use `as: :json` and `response.parsed_body` instead of monkey patching `ActiveDispatch::TestResponse#json` - #1847
- [x] Investigate ActiveJob, convert jobs to it if relevant [#475]
- [x] Make `ApplicationJob` for ActiveJob to inherit from ([link](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#activejob-now-inherits-from-applicationjob-by-default)) [#475]
- [x] Comb over `before_…` callbacks for returning `false` – this should now `throw(:abort)` instead
- [x] Handle the upcoming change where `belongs_to` requires an association by default gracefully ([link](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#active-record-belongs-to-required-by-default-option))
- [x] Investigate output options for `db:schema:dump`? ([link](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#configure-the-output-of-db-structure-dump))
- [x] Remove manual dependency on `redis ~> 3.0` (#507)
- [ ] Move to request specs per #57 (#728)
- [x] Have Rails serialize the models for jobs? (It can do that now; we don't need to pass `.id` and get it to retrieve the model.)
- [x] Use the `async` queue adapter in development instead of Resque? (`heroku local` runs Resque anyway, so it might not be necessary, but with `async` presumably it also gets run on `rails s`.)

Throne3d:
Now has remaining tasks from #475

Throne3d:
[...]

PR #567, opened 2017-11-18
Make belongs_to associations explicitly state optional / required

teceler:
Update `belongs_to` as per #476

**Edit by throne**: managed to make fewer tests fail (<https://travis-ci.org/Marri/glowfic/builds/304084617> → <https://travis-ci.org/Marri/glowfic/builds/304093736>) by way of making, for each of `character_tag`, `gallery_tag` and `post_tag`, the `belongs_to :tag…` association `optional: true`; this is a hack and should be fixed at some point:

> Setting the join `*_tag` lines that refer to `:tag` to have `optional:
false` fails.
> 
> I am not sure why Rails knows the post has content_warnings, and it has
post_tags to match, but the post_tag for the new content warning has no
tag attached and so it errors. It seems like it should know it's the new
one. I cannot find a relevant place to add `inverse_of` to fix it.

Relatedly, `galleries_icon` should require an icon, but doing this breaks the ability to delete icons from a gallery using nested parameters, for some reason.

### DIFF
--- a/app/concerns/writable.rb
+++ b/app/concerns/writable.rb
@@ -2,12 +2,11 @@ module Writable
   extend ActiveSupport::Concern
 
   included do
-    belongs_to :character
-    belongs_to :icon
-    belongs_to :user
-    belongs_to :character_alias
+    belongs_to :character, optional: true
+    belongs_to :icon, optional: true
+    belongs_to :user, optional: false
+    belongs_to :character_alias, optional: true
 
-    validates_presence_of :user
     validate :character_ownership, :icon_ownership
 
     def has_icons?
--- a/app/models/board.rb
+++ b/app/models/board.rb
@@ -9,10 +9,10 @@ class Board < ApplicationRecord
   has_many :favorites, as: :favorite, dependent: :destroy
   belongs_to :creator, class_name: User, optional: false
 
-  has_many :board_authors
-  has_many :board_coauthors, -> { where(cameo: false) }, class_name: BoardAuthor
+  has_many :board_authors, inverse_of: :board
+  has_many :board_coauthors, -> { where(cameo: false) }, class_name: BoardAuthor, inverse_of: :board
   has_many :coauthors, class_name: User, through: :board_coauthors, source: :user
-  has_many :board_cameos, -> { where(cameo: true) }, class_name: BoardAuthor
+  has_many :board_cameos, -> { where(cameo: true) }, class_name: BoardAuthor, inverse_of: :board
   has_many :cameos, class_name: User, through: :board_cameos, source: :user
 
   validates_presence_of :name
--- a/app/models/board.rb
+++ b/app/models/board.rb
@@ -7,7 +7,7 @@ class Board < ApplicationRecord
   has_many :posts
   has_many :board_sections, dependent: :destroy
   has_many :favorites, as: :favorite, dependent: :destroy
-  belongs_to :creator, class_name: User
+  belongs_to :creator, class_name: User, optional: false
 
   has_many :board_authors
   has_many :board_coauthors, -> { where(cameo: false) }, class_name: BoardAuthor
@@ -15,7 +15,7 @@ class Board < ApplicationRecord
   has_many :board_cameos, -> { where(cameo: true) }, class_name: BoardAuthor
   has_many :cameos, class_name: User, through: :board_cameos, source: :user
 
-  validates_presence_of :name, :creator
+  validates_presence_of :name
 
   after_destroy :move_posts_to_sandbox
 
--- a/app/models/character_tag.rb
+++ b/app/models/character_tag.rb
@@ -1,10 +1,9 @@
 class CharacterTag < ApplicationRecord
-  belongs_to :character, inverse_of: :character_tags
-  belongs_to :tag
-  belongs_to :label, foreign_key: :tag_id
-  belongs_to :setting, foreign_key: :tag_id
-  belongs_to :gallery_group, foreign_key: :tag_id
-  validates_presence_of :character
+  belongs_to :character, inverse_of: :character_tags, optional: false
+  belongs_to :tag, optional: false
+  belongs_to :label, foreign_key: :tag_id, optional: true
+  belongs_to :setting, foreign_key: :tag_id, optional: true
+  belongs_to :gallery_group, foreign_key: :tag_id, optional: true
 
   after_create :add_galleries_to_character
   after_destroy :remove_galleries_from_character
--- a/app/models/character_tag.rb
+++ b/app/models/character_tag.rb
@@ -1,6 +1,6 @@
 class CharacterTag < ApplicationRecord
   belongs_to :character, inverse_of: :character_tags, optional: false
-  belongs_to :tag, inverse_of: :character_tags, optional: true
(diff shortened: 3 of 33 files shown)


app/concerns/writable.rb
```diff
@@ -2,10 +2,10 @@ module Writable
   extend ActiveSupport::Concern
 
   included do
-    belongs_to :character
-    belongs_to :icon
-    belongs_to :user
-    belongs_to :character_alias
+    belongs_to :character, optional: true
+    belongs_to :icon, optional: true
+    belongs_to :user, optional: true
```
Throne3d: user not optional – see below, `validates_presence_of :user`

app/models/board.rb
```diff
@@ -7,16 +7,16 @@ class Board < ApplicationRecord
...
   has_many :board_authors
   has_many :board_coauthors, -> { where(cameo: false) }, class_name: BoardAuthor
   has_many :coauthors, class_name: User, through: :board_coauthors, source: :user
   has_many :board_cameos, -> { where(cameo: true) }, class_name: BoardAuthor
   has_many :cameos, class_name: User, through: :board_cameos, source: :user
 
-  validates_presence_of :name, :creator
-
+  validates_presence_of :name
+
```
Throne3d: accidental spacing, oops?

app/models/character_tag.rb
```diff
@@ -1,10 +1,9 @@
 class CharacterTag < ApplicationRecord
-  belongs_to :character, inverse_of: :character_tags
-  belongs_to :tag
-  belongs_to :label, foreign_key: :tag_id
-  belongs_to :setting, foreign_key: :tag_id
-  belongs_to :gallery_group, foreign_key: :tag_id
-  validates_presence_of :character
+  belongs_to :character, inverse_of: :character_tags, optional: false
+  belongs_to :tag, optional: true
```
Throne3d: tag not optional (a `character_tag` joins a character and a tag, so, it has to have both attached)

app/models/flat_post.rb
```diff
@@ -1,5 +1,4 @@
 class FlatPost < ApplicationRecord
-  belongs_to :post, inverse_of: :flat_post
+  belongs_to :post, inverse_of: :flat_post, optional: false
```
Throne3d: trailing linebreak can be removed

app/models/post.rb
```diff
@@ -12,10 +12,10 @@ class Post < ApplicationRecord
   STATUS_HIATUS = 2
   STATUS_ABANDONED = 3
 
-  belongs_to :board, inverse_of: :posts
-  belongs_to :section, class_name: BoardSection, inverse_of: :posts
-  belongs_to :last_user, class_name: User
-  belongs_to :last_reply, class_name: Reply
+  belongs_to :board, inverse_of: :posts, optional: false
+  belongs_to :section, class_name: BoardSection, inverse_of: :posts, optional: true
+  belongs_to :last_user, class_name: User, optional: true
```
Throne3d: I'm not sure about this one but should check with @Marri.

Throne3d:
We are probably missing some `inverse_of`s which will fix the Travis build

app/concerns/writable.rb
```diff
@@ -2,10 +2,10 @@ module Writable
...
-    belongs_to :character
-    belongs_to :icon
-    belongs_to :user
-    belongs_to :character_alias
+    belongs_to :character, optional: true
+    belongs_to :icon, optional: true
+    belongs_to :user, optional: false
+    belongs_to :character_alias, optional: true
 
     validates_presence_of :user
```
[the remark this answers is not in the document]
teceler: Which part can go away?

app/models/gallery_tag.rb
```diff
@@ -1,7 +1,7 @@
 class GalleryTag < ApplicationRecord
-  belongs_to :gallery, inverse_of: :gallery_tags
-  belongs_to :tag
-  belongs_to :gallery_group, foreign_key: :tag_id
+  belongs_to :gallery, inverse_of: :gallery_tags, optional: false
+  belongs_to :tag, inverse_of: :gallery_tags, optional: true
+  belongs_to :gallery_group, foreign_key: :tag_id, optional: false # This is currently required but may not continue to be
```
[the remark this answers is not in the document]
teceler: it looks like tag-not-being-required was necessary to make a test stop failing bizarrely

app/concerns/writable.rb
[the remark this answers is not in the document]
Throne3d: The `validates_presence_of` bit, since `user` is optional in the above

app/models/gallery_tag.rb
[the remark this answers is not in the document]
Throne3d: Tag should be required (but breaks stuff for some reason :( ), STI bit is temporarily marked required because it's the only possible STI in the list.

app/models/post_tag.rb
```diff
@@ -1,8 +1,7 @@
 class PostTag < ApplicationRecord
-  belongs_to :post, inverse_of: :post_tags
-  belongs_to :user
-  belongs_to :tag
-  belongs_to :setting, foreign_key: :tag_id
-  belongs_to :content_warning, foreign_key: :tag_id
-  belongs_to :label, foreign_key: :tag_id
+  belongs_to :post, inverse_of: :post_tags, optional: false
+  belongs_to :tag, inverse_of: :post_tags, optional: true
```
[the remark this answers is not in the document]
Throne3d: It broke things when it was not optional – see [here](https://travis-ci.org/Marri/glowfic/builds/304084617#L1368) vs [here](https://travis-ci.org/Marri/glowfic/builds/304093736#L1368), for some reason it thinks new tags are invalid?

(I wonder if the problem is somehow in the Taggable concern. I am hoping not but maybe we should get to doing something about that – i.e. fixing it from what I implemented – so it's more easily understood.)

app/models/post.rb
```diff
@@ -12,28 +12,30 @@ class Post < ApplicationRecord
...
+  belongs_to :last_reply, class_name: Reply, optional: true
   has_one :flat_post
   has_many :replies, inverse_of: :post, dependent: :destroy
   has_many :post_viewers, inverse_of: :post, dependent: :destroy
   has_many :viewers, through: :post_viewers, source: :user
   has_many :reply_drafts, dependent: :destroy
+  has_many :favorites, as: :favorite, dependent: :destroy
+
```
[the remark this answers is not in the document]
Throne3d: Separating the tags from the rest – this list was getting long.

app/models/post.rb
Throne3d: … Did I? Huh, uh, okay.

config/initializers/new_framework_defaults.rb
```diff
@@ -19,4 +19,4 @@
 ActiveSupport.to_time_preserves_timezone = false
 
 # Require `belongs_to` associations by default. Previous versions had false.
-Rails.application.config.active_record.belongs_to_required_by_default = false
+Rails.application.config.active_record.belongs_to_required_by_default = true
```
Throne3d: This line and the comment above (and the empty line above it) should be removed – it defaults to `true` and this is just a temporary upgrade help file, we want to get to removing the TODOs from the list.

Throne3d:
Relevant issues: #595, #596

PR #584, opened 2017-11-21
Clean up css classes named by color

teceler:
* Makes `.link-box`es use semantic class names
* Splits the two uses of `.dark` unto their own clases
* Folds `.green`'s color styling into a few pre-existing things
* Creates three new semantic classes to cover the rest of `.green`
* Cleans up `#search_form` (unrelated)

### DIFF
--- a/app/assets/stylesheets/galleries.scss
+++ b/app/assets/stylesheets/galleries.scss
@@ -72,3 +72,21 @@
   margin: 0;
 }
 .loading-icon { width: 16px; height: 16px; }
+
+/* Various tables with icons in previously styled with .green */
+.icons-box {
+  background-color: #B0BBA9;
+}
+
+/* Icon credit, previously styled with .green */
+.icon-credit {
+  background-color: #B0BBA9;
+  text-align: center;
+}
+
+/* The preview icon boxes on replace icon and replace character */
+.replace-icon {
+  @extend .icons-box;
+  width: 150px;
+  text-align: center;
+}
--- a/app/assets/stylesheets/global.scss
+++ b/app/assets/stylesheets/global.scss
@@ -83,7 +83,6 @@ button, .button {
 
 /* Displaying forum threads */
 .green { background-color: #B0BBA9; }
-.dark { background-color: #8B8687; }
 .even, .odd, .post-post, .post-reply, #post-editor {
   padding: 5px;
 }
--- a/app/assets/stylesheets/global.scss
+++ b/app/assets/stylesheets/global.scss
@@ -82,7 +82,6 @@ button, .button {
 }
 
 /* Displaying forum threads */
-.green { background-color: #B0BBA9; }
 .even, .odd, .post-post, .post-reply, #post-editor {
   padding: 5px;
 }
--- a/app/assets/stylesheets/global.scss
+++ b/app/assets/stylesheets/global.scss
@@ -109,10 +109,12 @@ button, .button {
 
   img { vertical-align: text-top; }
 }
-.link-box.green { background-color: #3f7055; }
-.link-box.red { background-color: #880000; }
-.link-box.blue { background-color: #4455AA;}
-.link-box.orange { background-color: #EA8A40;}
+.link-box.action-new { background-color: #3f7055; }
+.link-box.action-delete { background-color: #880000; }
+.link-box.action-edit { background-color: #4455AA;}
+.link-box.orange { background-color: #EA8A40;} // TODO: This does not appear to be used anywhere
+.link-box.action-favorite { background-color: #4455AA;}
+.link-box.action-dismiss { background-color: #880000; }
 #post-form-wrapper { margin-top: 5px; }
 
 /* Inline lists */
--- a/app/assets/stylesheets/global.scss
+++ b/app/assets/stylesheets/global.scss
@@ -56,7 +56,8 @@ button, .button {
 .width-150 { width: 150px; }
 .width-250 { width: 250px; }
 .width-300 { width: 300px; }
-.spacer { height: 2px; padding: 0; }
+.spacer, .space-alt { height: 2px; padding: 0; }
+.spacer-alt { background-color: #8B8687; }
 .centered { text-align: center !important; }
 .left-align { text-align: left; }
 .right-align { text-align: right; }
--- a/app/assets/stylesheets/global.scss
+++ b/app/assets/stylesheets/global.scss
@@ -56,7 +56,7 @@ button, .button {
 .width-150 { width: 150px; }
 .width-250 { width: 250px; }
 .width-300 { width: 300px; }
-.spacer, .space-alt { height: 2px; padding: 0; }
+.spacer, .spacer-alt { height: 2px; padding: 0; }
 .spacer-alt { background-color: #8B8687; }
 .centered { text-align: center !important; }
 .left-align { text-align: left; }
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -302,6 +302,7 @@ td, .table-list li { font-size: 14px; }
 .user-right-content-box { width: calc(100% - 200px); }
 .icon-right-content-box { width: calc(100% - 150px); }
 .character-right-content-box { width: calc(100% - 175px); }
+#user-avatar { text-align: center; }
 
 /* Not yet deserving of its own CSS file */
 #stats ul { margin-top: 0; margin-bottom: 0; padding-left: 15px; }
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -329,5 +329,13 @@ td, .table-list li { font-size: 14px; }
 #homepage-content {
   padding: 20px;
 }
+
+/* Search Input */
+.search-box {
+  width: 300px;
+  padding: 10px;
+  vertical-align: top;
+  background-color: #8B8687;
+}
 .search-icon-preview { vertical-align: middle; }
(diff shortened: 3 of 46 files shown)


app/assets/stylesheets/layout.scss
```diff
@@ -302,6 +302,7 @@ td, .table-list li { font-size: 14px; }
 .user-right-content-box { width: calc(100% - 200px); }
 .icon-right-content-box { width: calc(100% - 150px); }
 .character-right-content-box { width: calc(100% - 175px); }
+#user-avatar { text-align: center !important; background-color: #B0BBA9; }
```
Marri: Is this the right background color in the other layouts?

teceler: I think I have `#user-avatar` on the 'things replacing green' list in them. May put that into icon-boxes on it anyway though, for ease of use.

app/assets/stylesheets/global.scss
```diff
@@ -109,10 +108,12 @@ button, .button {
...
   img { vertical-align: text-top; }
 }
-.link-box.green { background-color: #3f7055; }
-.link-box.red { background-color: #880000; }
-.link-box.blue { background-color: #4455AA;}
-.link-box.orange { background-color: #EA8A40;}
+.link-box.action-new { background-color: #3f7055; }
+.link-box.action-delete { background-color: #880000; }
+.link-box.action-edit { background-color: #4455AA;}
+.link-box.orange { background-color: #EA8A40;} // This doesn't.... seem to actually be used anywhere?
```
Throne3d: I'd add a "TODO" note before this so it doesn't get lost

app/assets/stylesheets/layout.scss
```diff
@@ -302,6 +302,7 @@ td, .table-list li { font-size: 14px; }
 .user-right-content-box { width: calc(100% - 200px); }
 .icon-right-content-box { width: calc(100% - 150px); }
 .character-right-content-box { width: calc(100% - 175px); }
+#user-avatar { text-align: center !important; }
```
Throne3d: Is the `!important` necessary? I try to avoid it wherever possible. (I assume this is taken from, like, `.centered` or something)

app/assets/stylesheets/layout.scss
```diff
@@ -329,3 +330,11 @@ td, .table-list li { font-size: 14px; }
 #homepage-content {
   padding: 20px;
 }
+
+/* Search */
```
Throne3d: Is this the left sidebar where you input your search, or the right box that shows the results? ("Search input", "Search results")

app/assets/stylesheets/galleries.scss
```diff
@@ -73,3 +73,20 @@
...
+.icons-box {
+  background-color: #B0BBA9;
+}
+
+/* Icon-related text previously styled with .green */
+.icon-text {
+  background-color: #B0BBA9;
+}
+
+/* The specific icon boxes on replace icon and replace character */
```
Throne3d: The small icon boxes? I think these are styled semantically similarly to, say, the icon box on a post?

app/assets/stylesheets/layouts/river.scss
```diff
@@ -17,12 +19,12 @@ body { background-color: #E0E0E0; }
...
 .post-expander { background-color: #A4B0BD; }
 .pagination .current { background: #718EB9; }
-.post-icon, #current-icon-dropdown, .green { background-color: #A9B4BB; }
+.post-icon, #current-icon-dropdown, #reply-icon-selector, .replace-icon, .icons-box, .icon-text { background-color: #A9B4BB; }
 a { color: #3F5970; }
 a:visited { color: #284150; }
 .flash.breadcrumbs, .form-table-ender { background-color: #A3A8A9; }
 .subber { background-color: #B9B6AD; }
-.link-box.green { background-color: #3F5370; }
+.link-box.action-new { background-color: #3F5370; }
```
Throne3d: Can we move this up to be with the other `link-box`es? (sorry that it wasn't done before)

app/assets/stylesheets/layouts/starrydark.scss
```diff
@@ -49,9 +49,9 @@ $ODD_COLOR: #25262E;
   }
 }
 
-.post-navheader .view-button, .view-button.selected, .view-button, .link-box.green, .link-box.blue, .link-box.red, .tag-item { background-color: #463C65; }
+.post-navheader .view-button, .view-button.selected, .view-button, .link-box.action-new, .link-box.action-edit, .link-box.action-favorite, .link-box.action-delete, .link-box.action-dismiss, .tag-item { background-color: #463C65; }
 .flash.breadcrumbs .link-box, .subber .link-box { background-color: #5C5778; }
-#post-editor .green { background-color: #484357; color: $FONT_COLOR; }
+#reply-icon-selector { background-color: #484357; color: $FONT_COLOR; } // replies.scss
```
Throne3d: Is the comment here still necessary? (It's not in, say, `starry.scss`)

app/assets/stylesheets/galleries.scss
```diff
@@ -73,3 +73,20 @@
...
 }
 .loading-icon { width: 16px; height: 16px; }
+
+/* Various tables with icons in previously styled with .green */
+.icons-box {
+  background-color: #B0BBA9;
+}
+
+/* Icon-related text previously styled with .green */
+.icon-text {
```
Throne3d: This is for, specifically, `credit` on an icon's page, and for the "No galleries yet" box, I think? What does the credit thing look like – I don't think these are semantically similar?

Throne3d: I think I would put the "No galleries yet" thing as just a piece of plain text in a `.icons-box` – it feels to me like it's the same element, semantically, as the thing that lists icons and galleries, it just doesn't have galleries in.

The credit thing should probably be styled together with the icon, if it can be? I actually have not looked at that page yet, uh, apologies for the comments that are partially guessing.

app/assets/stylesheets/global.scss
```diff
@@ -57,6 +57,7 @@ button, .button {
 .width-250 { width: 250px; }
 .width-300 { width: 300px; }
 .spacer { height: 2px; padding: 0; }
+.spacer-alt { height: 2px; padding: 0; background-color: #8B8687; }
```
Throne3d: I would modify these two lines to be like so:

```css
.spacer, .spacer-alt { height: 2px; padding: 0; }
.spacer-alt { background-color: #8B8687; }
```

That is, keep the layout structure the same between them, forcibly (so nobody accidentally changes one and misses the other), and then style the `spacer-alt` with a different color after.

app/assets/stylesheets/layout.scss
teceler: ...huh i thought #user-avatar was there to start with. I might just flip that back to using .centered. (and yeah centered does that)

app/assets/stylesheets/galleries.scss
teceler: I /can/ but then I would have to put the specified padding and stuff back too.

app/assets/stylesheets/layouts/river.scss
teceler: that is a thing I probably would have done when I stripped river.scss but i can do it now, sure

app/assets/stylesheets/layouts/starrydark.scss
teceler: ...that's. part of my notes to self for refactoring/stripping starrydark. why is it here -- oh, I think I was noting that for that, okay.

app/assets/stylesheets/galleries.scss
teceler: it is those two things! I can put the 'no galleries yet' in an .icons-box yeah. and. uh. idk about the credit? it was styled separately when i found it.

app/assets/stylesheets/layout.scss
Throne3d: It seems sensible to start extracting `.centered` into the relevant locations, especially if we can avoid `!important` in those areas, because I really really dislike `!important`. xD Is `#user-avatar` in another file, maybe?

app/assets/stylesheets/galleries.scss
Throne3d: Hm, okay, I would keep them separate, then. Same thing as I mentioned elsewhere for `!important`, though – if you can remove that and it doesn't change the UI, I would much appreciate that. (And the comment comment (ha) still stands.)

app/assets/stylesheets/galleries.scss
Throne3d: Maybe check for uses of `.credit` and if you don't see it anywhere, change this selector to `.credit`? If you do see it elsewhere, I'd rename this to some third thing, which makes it clear it only applies to `credit` in that one location wherever it was, since `icon-text` sounds like it could be talking about the keywords below icons, say, when you're adding existing icons to a gallery.

app/assets/stylesheets/layout.scss
teceler: it's in users #show and did apparently pre-exist, just as a class that doesn't seem to have had anything attached? ...uh, unless it had javascript attached

app/assets/stylesheets/galleries.scss
teceler: did the comment change and the `!important` thing

app/assets/stylesheets/layout.scss
Throne3d: After a quick look, it seems like it's `.user-avatar`, and it's only used on `users#show` – there doesn't seem to be any JavaScript interacting with it, so this change should be good without the `!important`,

PR #587, opened 2017-11-22
Mods

Marri:
Next round: editing characters!
- Turn off the UI for default character icon and creating templates if the editor is not the owner
- Character editor uses correct templates/galleries rather than defaulting to current user's
- Correct edit/delete permissions
- Requires audit comment when editing a character you don't own
- Incidentally give Throne's seed user mod permissions while I'm in here

### DIFF
--- a/app/views/characters/_editor.haml
+++ b/app/views/characters/_editor.haml
@@ -15,19 +15,21 @@
   %th.sub.vtop Template
   %td{class: klass}
     = f.collection_select :template_id, @templates, :id, :name, {include_blank: '— Choose Template —'}, {class: 'text chosen-select', disabled: params[:new_template].present?}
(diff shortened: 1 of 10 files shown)


app/views/characters/_editor.haml
```diff
@@ -39,6 +41,10 @@
 %tr
   %th.sub.vtop Description
   %td{class: cycle('even', 'odd')}= f.text_area :description, placeholder: 'Description', class: 'text', cols: 35
+- if current_user.id != @character.user_id
```
Throne3d: `unless … ==`?

Marri:
Ah, yeah, I think I'd had multiple checks in here and took one out and
didn't fix it. Ah well.

PR #592, opened 2017-11-23
Tag refactor

Marri:
We used to build all our specialized tag logic on top of the model in a concern; this moves the specialized logic into the controllers (still in a concern, just for the controller) and also moves the view logic into the tag helper.

Remaining:
- [ ] ALL THE TESTS
- [x] Sort the all_tags variable in Taggable by original selection order
- [x] Uniq the tags in Taggable by case-insensitive name
- [ ] ungrouped_galleries. help what.


app/concerns/taggable.rb
```diff
@@ -1,71 +1,29 @@
 module Taggable
-  extend ActiveSupport::Concern
+  def process_tags(klass, obj_param, id_param)
+    # fetch and clean tag ids
+    ids = params.fetch(obj_param, {}).permit(id_param => [])[id_param]
```
Throne3d: would this work as `.fetch(id_param, [])` instead of `.permit(…)[…]`?

spec/models/setting_spec.rb
```diff
@@ -29,4 +29,24 @@
...
     end
   end
+
+  context "tags" do
+    it "creates no tags on invalid create" do
+      harry_potter = create(:setting, name: 'Harry Potter')
+      setting = build(:setting, name: '', parent_settings: [harry_potter])
+      expect(setting.valid?).to eq(false)
+      expect(setting.save).to eq(false)
+      expect(setting.parent_settings.count).to eq(0) # Rails clears associations on invalid save
```
Throne3d: I think if you use `.size` it doesn't, but if you use `.count` it does?

app/controllers/posts_controller.rb
```diff
@@ -139,10 +143,19 @@ def update
...
+    warnings = process_tags(ContentWarning, :post, :content_warning_ids)
+    labels = process_tags(Label, :post, :label_ids)
+
+    if params[:button_preview].present?
+      @post.association(:settings).target = []
+      @post.association(:content_warnings).target = []
+      @post.association(:labels).target = []
+      settings.each { |s| @post.association(:settings).add_to_target(s) }
+      warnings.each { |w| @post.association(:content_warnings).add_to_target(w) }
+      labels.each { |l| @post.association(:labels).add_to_target(l) }
```
Throne3d: What's this bit working around?

app/helpers/tag_helper.rb
```diff
@@ -5,4 +5,17 @@ def delete_path(tag)
...
+    collection = form.object.send(assoc)
+    ids = collection.map(&:id_for_select)
+
+    form.select(
+      attr_name,
+      options_from_collection_for_select(collection, :id_for_select, :name, ids),
+      {},
+      {multiple: true}.merge(opts)
+    )
+  end
```
Throne3d: <3

spec/controllers/characters_controller_spec.rb
```diff
@@ -405,7 +405,7 @@
       expect(character.characters_galleries.first).not_to be_added_by_group
 
       login_as(user)
-      put :update, params: { id: character.id, character: {ungrouped_gallery_ids: ['']} }
+      put :update, params: { id: character.id, character: {ungrouped_gallery_ids: [''], gallery_group_ids: [group.id]} }
```
Throne3d: What's up with this?

spec/controllers/characters_controller_spec.rb
```diff
@@ -478,14 +478,14 @@
         create(:template)
 
         login_as(user)
-        put :update, params: { id: character.id, character: {name: ''} }
+        put :update, params: { id: character.id, character: {name: '', gallery_group_ids: [group.id]} }
```
Throne3d: ditto

spec/controllers/posts_controller_spec.rb
```diff
@@ -406,6 +407,8 @@
       char2 = create(:template_character, user: user)
       expect(controller).to receive(:editor_setup).and_call_original
       expect(controller).to receive(:setup_layout_gon).and_call_original
+
+      # valid post requires a board_id
```
Throne3d: Is this a TODO?

spec/controllers/posts_controller_spec.rb
```diff
@@ -909,6 +928,11 @@
       expect(post.icon).to be_nil
       login_as(user)
 
+      # extras to not be in the array
+      create(:setting)
+      create(:content_warning)
+      create(:label)
```
Throne3d: <3

spec/controllers/posts_controller_spec.rb
```diff
@@ -1258,6 +1284,59 @@
...
+        expect(post.content_warnings.map(&:name)).to match_array([warning.name, 'warning', 'dupewarning'])
+        expect(post.labels.map(&:name)).to match_array([label.name, 'label', 'dupelabel'])
+        expect(Setting.count).to eq(3)
+        expect(ContentWarning.count).to eq(3)
+        expect(Label.count).to eq(3)
+        expect(PostTag.count).to eq(6)
+        expect(PostTag.where(post: post, tag: [setting, warning, label]).count).to eq(3)
+        expect(PostTag.where(post: post, tag: [dupes, dupew, dupel]).count).to eq(0)
+        expect(PostTag.where(post: post, tag: [reml, remw, rems]).count).to eq(3)
+      end
```
Throne3d: I love this test so much.

spec/controllers/posts_controller_spec.rb
```diff
@@ -1337,19 +1425,50 @@
...
+        expect(post.settings.map(&:name)).to match_array([setting.name, 'setting', 'dupesetting'])
+        expect(post.content_warnings.map(&:name)).to match_array([warning.name, 'warning', 'dupewarning'])
+        expect(post.labels.map(&:name)).to match_array([label.name, 'label', 'dupelabel'])
+        expect(Setting.count).to eq(4)
+        expect(ContentWarning.count).to eq(4)
+        expect(Label.count).to eq(4)
+        expect(PostTag.count).to eq(9)
+        expect(PostTag.where(post: post, tag: [setting, warning, label]).count).to eq(3)
+        expect(PostTag.where(post: post, tag: [dupes, dupew, dupel]).count).to eq(3)
+        expect(PostTag.where(post: post, tag: [reml, remw, rems]).count).to eq(0)
```
Throne3d: This test also has my love.

spec/models/setting_spec.rb
```diff
@@ -29,4 +23,26 @@
...
+
+    it "creates tags on valid create" do
+      harry_potter = create(:setting, name: 'Harry Potter')
+      setting = build(:setting, parent_settings: [harry_potter])
+      expect(setting.valid?).to eq(true)
+      expect(setting.save).to eq(true)
+      expect(setting.parent_settings.count).to eq(1)
+      expect(setting.parent_settings.size).to eq(1)
+      expect(TagTag.count).to eq(1)
+    end
```
Throne3d: <3

app/controllers/posts_controller.rb
Marri: So that they show properly in preview! this way they are in memory, whereas using @post.<name> = thing will actually create the records in the database, which is not at all the goal of preview.

spec/controllers/characters_controller_spec.rb
Marri: Oh I remember! Because the thing wasn't there, it was taking that as "this character should be assigned to have zero gallery groups", and then it wasn't persisting galleries that were tethered by gallery groups _because there were also no gallery groups_. So this is now necessary for the group to stay attached. Shouldn't matter in regular use, because this would be sent with the edit page as usual; just an edge case of not sending 100% of normal parameters in a test.

app/controllers/posts_controller.rb
Throne3d: Ahhh. Okay. I didn't know where to find documentation on these methods.

PR #599, opened 2017-11-26
Modify replies#search UI to use a div layout instead of table

Throne3d:
Also fixes large spacing around first and last paragraphs, and allows
text to flow around post info boxes, as in replies.


Throne3d:
Pending feedback from #constellation.

Throne3d:
Feedback from Constellation decided it was good to have the text wrap and move the timestamp to below the content, for consistency reasons:



They also approved of a new mobile layout, where the two columns get squished into one (placing successive elements after each other).

**Narrow but uncollapsed view**:


**Wide-ish but collapsed view**:


**Mobile-ish view**:


**Also posts#search**:


(This affects characters#search, posts#search, replies#search and users#search.)

PR #601, opened 2017-11-26
Don't require mod note from coauthors

Marri:


spec/controllers/posts_controller_spec.rb
```diff
@@ -1000,6 +1000,16 @@
...
     end
 
+    it 'does not require note from coauthors' do
+      post = create(:post, privacy: Concealable::ACCESS_LIST)
+      user = create(:user)
+      post.viewers << user
+      create(:reply, user: user, post: post)
+      login_as(user)
+      put :update, params: { id: post.id }
+      expect(flash[:error]).not_to eq('You must provide a reason for your moderator edit.')
```
Throne3d: This looks like it doesn't necessarily test what we want – can we give the post a subject, then have the coauthor edit that and we ensure it changed?

PR #603, opened 2017-11-30
Fix alignment of text boxes that have icons

### LINKED ISSUE #577: Buttons on board#edit display at weird heights in Firefox
"New Post", "New Section", "Delete" display consistently, but "Edit" and "Favorite" are at a different one – it could be related to the former ones having just text and no icons, and the latter having icons? Investigate.

Impacts UI in quite an ugly way.

teceler:
Changing this line seems to fix #577 on firefox and not significantly disrupt it on chromium; other browsers should perhaps be tested, since the results of `vertical-align` seems to vary by browser.

### DIFF
--- a/app/assets/stylesheets/global.scss
+++ b/app/assets/stylesheets/global.scss
@@ -105,6 +105,7 @@ button, .button {
   border-radius: 5px;
   font-weight: normal;
   height: 20px;
+  vertical-align: middle;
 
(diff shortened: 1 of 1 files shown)


Throne3d:
This line didn't modify Chrome for me, but broke Firefox. Adding the `vertical-align: middle;` that I did seems to make it work fine in Firefox.

PR #611, opened 2017-12-08
Upgrade to Rails 5.1

### LINKED ISSUE #589: Upgrade to Rails 5.1
It probably comes with a bunch of bug fixes and also some new useful features.

#476 should be addressed before the upgrade, though, as this likely removes the config options to retain deprecated behavior.

Throne3d:
Fixes #589.

Relevant deprecation warnings:

> DEPRECATION WARNING: Passing a class to the `class_name` is deprecated and will raise an ArgumentError in Rails 5.2. It eagerloads more classes than necessary and potentially creates circular dependencies. Please pass the class name as a string: `belongs_to :child_setting, class_name: 'Setting'` (called from <class:TagTag> at /home/travis/build/Marri/glowfic/app/models/tag_tag.rb:2)

> DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_change_to_attribute?` instead. (called from block in order_change? at /media/throne3d/Storage/Documents/glowfic/app/concerns/orderable.rb:37)

> DEPRECATION WARNING: The behavior of `attribute_was` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `attribute_before_last_save` instead. (called from reorder_others at /media/throne3d/Storage/Documents/glowfic/app/concerns/orderable.rb:15)

For more information around context and motivation, the individual commits should have that in their commit messages.

[Rails 5.1 release notes](http://guides.rubyonrails.org/5_1_release_notes.html)


app/models/post.rb
```diff
@@ -270,24 +271,19 @@ def set_last_user
...
-  def timestamp_attributes_for_update
-    # Makes Rails treat edited_at as a timestamp identical to updated_at
-    # if specific attributes are updated. Also uses tagged_at if there
-    # are no replies yet or if the status has changed.
-    # Be VERY CAREFUL editing this!
-    return super if skip_edited
-    return super + [:edited_at] if replies.exists? && !status_changed?
-    super + [:edited_at, :tagged_at]
+  # timestamps start existing between before_save and before_create/update
+  # TODO: ensure tested
```
Throne3d: fix todo before merging

Throne3d:
Still WIP because I need to figure out again if the deprecation warnings are broken for the cases they appear or if they're actually mistakes like I thought.

Throne3d:
The deprecation warnings appear to be erroneous.

They are in fact occurring inside an `after` callback, but more importantly the `after` callback is triggering a `before` callback where the data examined.

config/puma.rb
```diff
@@ -1,15 +1,15 @@
 # Puma can serve each request in a thread from an internal thread pool.
-# The `threads` method setting takes two numbers a minimum and maximum.
+# The `threads` method setting takes two numbers: a minimum and maximum.
 # Any libraries that use thread pools should be configured to match
 # the maximum value specified for Puma. Default is set to 5 threads for minimum
-# and maximum, this matches the default thread size of Active Record.
+# and maximum; this matches the default thread size of Active Record.
 #
-threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i
+threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }
```
Marri: ...why isn't .to_i necessary? It should handle weird env variables.

config/initializers/cookies_serializer.rb
```diff
@@ -2,4 +2,4 @@
 
 # Specify a serializer for the signed and encrypted cookie jars.
 # Valid options are :json, :marshal, and :hybrid.
-Rails.application.config.action_dispatch.cookies_serializer = :hybrid
+Rails.application.config.action_dispatch.cookies_serializer = :json
```
Marri: Will this log everyone out?

app/models/post.rb
Marri: still has a todo :D

config/initializers/cookies_serializer.rb
Throne3d: I don't think so – afaik hybrid deserializes either, json only deserializes a particular one. I will deploy to staging and check <3

Throne3d: Testing it, it didn't log me out, so no, I expect it won't.

app/models/post.rb
Throne3d: I think I'm confident that `Post should have the right timestamps` tests this, so I'll remove the TODO.

config/puma.rb
Throne3d: It was removed in the default configuration. Do we need to handle weird env variables?

I tried looking for where the default configuration comes from but couldn't find it in the puma gem, so I couldn't see their rationale for removing it.

config/initializers/assets.rb
```diff
@@ -3,10 +3,13 @@
 # Version of your assets, change this if you want to expire all your assets.
 Rails.application.config.assets.version = '1.0'
 
-# Add additional assets to the asset load path
+# Add additional assets to the asset load path.
 # Rails.application.config.assets.paths << Emoji.images_path
+# Add Yarn node_modules folder to the asset load path.
+Rails.application.config.assets.paths << Rails.root.join('node_modules')
```
Marri: Do we need this? Are we using Yarn?

config/environments/production.rb
```diff
@@ -14,6 +14,11 @@
   config.consider_all_requests_local       = false
   config.action_controller.perform_caching = true
 
+  # Attempt to read encrypted secrets from `config/secrets.yml.enc`.
+  # Requires an encryption key in `ENV["RAILS_MASTER_KEY"]` or
+  # `config/secrets.yml.key`.
+  config.read_encrypted_secrets = true
```
Marri: As long as this doesn't *require* us to have these set up, that's fine, since I don't think they're there yet.

config/initializers/assets.rb
Throne3d: We're using it insofar as it's now installed by default, but we're not actually using it in that we have no actual dependencies that require it. I think it doesn't damage anything by having it, though, and it's now a default component and Rails guides may in the future assume we have Yarn if we're on a modern Rails version.

config/environments/production.rb
Throne3d: The server seems to run fine without a `config/secrets.yml.key` file, so it looks like it's not required.

Throne3d:
Followup issues: #660 #661

PR #615, opened 2017-12-11
Consistent cookie treatment

Marri:


config/initializers/session_store.rb
```diff
@@ -1,7 +1,7 @@
 # Be sure to restart your server when you modify this file.
 
 options = if Rails.env.production?
-  {domain: ['glowfic.com', '.glowfic-staging.herokuapp.com'], tld_length: 2}
+  {domain: ['.glowfic.com', '.glowfic-staging.herokuapp.com'], tld_length: 2}
```
Marri: @Throne3d looking at it, this seems actually wildly wrong. Doesn't this mean sessions are shared between staging and production? Isn't that in fact SUPER not what we want?

Marri: Tentatively internet says it'll generate a cookie per .domain

PR #650, opened 2018-01-02
Post authors

Marri:
To Do:
- [ ] Add some way to remove / deactivate users? Probably with a can_post boolean and xs on the author list
- [x] The `authors_locked` checkbox should automatically check itself if you add authors on `#new` or `#edit`.
- [x] It should also be checked if the page is loaded with authors set to the board's (but presumably not just "if the post has authors").
  - [ ] When changing the board for a post to a closed board, somehow prompt the user if they want to set the board authors to be the authors? Or just do it if the list is empty?
- [ ] How do moderators interact with this all?
- [ ] Perhaps we should notify people by email when they're invited to a thread
- Unit tests for:
  - [x] `Board#writer_ids`? This is used when creating a post in a specific board.
  - [x] `Post#metadata_editable_by?` with new changes
  - [x] `Post#taggable_by?` with new changes (if not yet tested)
- [x] `Post#destroy` should destroy PostAuthors – do we want to check this?
- [x] `Reply#destroy` should un-join a user from a post if it's their only reply. (e.g. you accidentally respond to a post you shouldn't?) I think this currently isn't done
- [x] Add subheader to Owed>Hidden page to clarify why not-your-tag posts are shown (n.b. subheader is hard so just did more detailed header)
- [x] Find a better way to detect changed authors on a post for the callback
- [x] Better titles on the post#write divs
- [x] Add a way to see posts you have hidden from Tags Owed to the tags owed page
- [x] Add add / remove from Tags Owed to the post#show hamburger menu
- [x] Check: Post#create has tests that viewers are saved
- [x] Check: Post#create has tests that authors are saved
- [x] Check: Post#create has tests that settings are saved
- [x] Check: Post#create has tests that tags are saved
- [x] Check: Post#create has tests that warnings are saved
(No tests for Post#create preview associations - they can't be saved if the Post isn't saved)
- [x] Check: Post#update has tests that viewers are saved
- [x] Check: Post#update has tests that authors are saved
- [x] Check: Post#update has tests that settings are saved
- [x] Check: Post#update has tests that tags are saved
- [x] Check: Post#update has tests that warnings are saved
- [x] Check: Post#update preview has tests that viewers aren't saved
- [x] Check: Post#update preview has tests that authors aren't saved
- [x] Check: Post#update preview has tests that settings aren't saved
- [x] Check: Post#update preview has tests that tags aren't saved
- [x] Check: Post#update preview has tests that warnings aren't saved


app/assets/javascripts/posts/editor.js
```diff
@@ -38,6 +38,12 @@ function setupMetadataEditor() {
     placeholder: 'Choose user(s) to view this post'
   });
 
+  $("#post_unjoined_author_ids").select2({
+    width: '300px',
+    minimumResultsForSearch: 20,
+    placeholder: 'Choose user(s) invited to reply to this post'
```
Throne3d: "invited" → "to invite" would sound more natural?

"Choose user(s) to invite to reply to this post".

app/concerns/owable.rb
```diff
@@ -0,0 +1,46 @@
...
+    private
+
+    def add_creator_to_authors
+      return if author_ids.include?(user_id)
+      post_authors.create(user: user, joined: true, joined_at: created_at)
+    end
+
+    def update_board_cameos
+      return if board.open_to_anyone?
+      return if (new_cameos = unjoined_authors - board.writers).empty? # TODO not authors
```
Throne3d: TODO?

app/controllers/posts_controller.rb
```diff
@@ -46,9 +44,14 @@ def mark
     posts = posts.select do |post|
       post.visible_to?(current_user)
     end
+
     if params[:commit] == "Mark Read"
       posts.each { |post| post.mark_read(current_user) }
       flash[:success] = posts.size.to_s + " posts marked as read."
+    elsif params[:commit] == "Remove from Replies Owed"
+      posts.each { |post| post.opt_out_of_owed(current_user) }
+      flash[:success] = posts.size.to_s + " posts removed from replies owed."
```
Throne3d: Can we use `pluralize`? e.g. "1 post removed" would be broken here.

app/controllers/posts_controller.rb
```diff
@@ -122,6 +128,16 @@ def stats
...
   def preview
+    @post ||= Post.new(user: current_user)
+    @post.assign_attributes(post_params(false))
+    @post.board ||= Board.find_by_id(3)
+
+    @author_ids = params.fetch(:post, {}).fetch(:unjoined_author_ids, [])
+    @viewer_ids = params.fetch(:post, {}).fetch(:viewer_ids, [])
+    @settings = process_tags(Setting, :post, :setting_ids)
+    @content_warnings = process_tags(ContentWarning, :post, :content_warning_ids)
+    @labels = process_tags(Label, :post, :label_ids)
```
Throne3d: If we don't already have them, tests that viewers and authors are persisted through preview and failed save? Also that preview doesn't persist them to the database.

app/models/post.rb
```diff
@@ -250,6 +235,11 @@ def reply_count
     replies.count
   end
 
+  def user_joined(user)
+    # TODO: logic and callbacks for someone joining a post
+    # TODO put on PostAuthor
```
Throne3d: Is this TODO planned to be done before we merge?

app/models/reply.rb
```diff
@@ -72,7 +72,7 @@ def destroy_draft
   def notify_other_authors
     return if skip_notify
     return if (previous_reply || post).user_id == user_id
-    post.authors.each do |author|
+    post.tagging_authors.each do |author|
```
Throne3d: <3

app/views/posts/_write.haml
```diff
@@ -11,8 +11,22 @@
...
       = f.collection_select(:viewer_ids, User.where.not(id: post.user_id).order('username asc'),
-      :id, :username, {selected: post.viewer_ids}, {multiple: true})
-    %br
+      :id, :username, {selected: @viewer_ids.nil? ? post.viewer_ids : @viewer_ids}, {multiple: true})
+
+    - if post.persisted?
+      #current_authors
+        = f.label :joined_author_ids, 'Current authors:'
+        - authors = post.joined_authors.order('LOWER(users.username)')
+        = authors.map { |author| link_to author.username, user_path(author) }.join(', ').html_safe
```
Throne3d: Were we going to have `×` buttons to delete authors from the thread? I forget what we decided on

app/controllers/posts_controller.rb
Marri: There's at least I think tag tests, because they caught the bug when I started moving things around; I'd have to check to see if they're all there?

spec/controllers/posts_controller_spec.rb
```diff
@@ -1718,26 +1722,6 @@
...
-            id: post.id,
-            post: {
-              unjoined_author_ids: [cameo_user.id]
-            }
-          }
-        }.to change { PostAuthor.count }.by(1)
-        expect(board.cameos.count).to eq(1)
-        expect(board.cameos).to eq([cameo_user])
-      end
-
```
Throne3d: What's up with this being removed? I think we want to make sure it does this on both create and update?

Marri:
There were two of them, I think the duplicate was a couple tests down

spec/controllers/posts_controller_spec.rb
Marri: Yep, checked, line 1796 has the same test for update, it was in fact a duplicate

spec/controllers/posts_controller_spec.rb
```diff
@@ -1511,6 +1685,136 @@
...
+        main_author = post.post_authors.find_by(user: user)
+        expect(main_author.can_owe).to eq(true)
+        expect(main_author.joined).to eq(true)
+        expect(main_author.joined_at).to be_the_same_time_as(post.created_at)
+
+        # doesn't set joined time but does set invited status when inviting new user
+        new_author = post.post_authors.find_by(user: other_user)
+        expect(new_author.can_owe).to eq(true)
+        expect(new_author.joined).to eq(false)
+        expect(new_author.joined_at).to be_nil
```
Throne3d: Should check `invited_at` and `invited_by`?

spec/controllers/replies_controller_spec.rb
```diff
@@ -229,6 +229,60 @@
...
+      user = create(:user)
+      other_user = create(:user)
+      login_as(user)
+      reply_post = create(:post, user: other_user, tagging_authors: [user, other_user], authors_locked: true)
+      post :create, params: { reply: {post_id: reply_post.id, content: 'test content!'} }
+      expect(Reply.count).to eq(1)
+      other_post = create(:post, user: user, tagging_authors: [user, other_user], authors_locked: true)
+      post :create, params: { reply: {post_id: other_post.id, content: 'more test content!'} }
+      expect(Reply.count).to eq(2)
+    end
```
Throne3d: I think it's bad practice to put two actions through a controller in a single spec – can we split this into `"allows replies from coauthors in a closed post"` and `"allows replies from main author in a closed post"`?

spec/controllers/replies_controller_spec.rb
```diff
@@ -229,6 +229,60 @@
...
+
+    it "adds authors correctly when a user replies to an open thread" do
+      user = create(:user)
+      login_as(user)
+      reply_post = create(:post)
+      post :create, params: { reply: {post_id: reply_post.id, content: 'test content!'} }
+      expect(Reply.count).to eq(1)
+      expect(reply_post.tagging_authors).to include(user)
+      expect(reply_post.tagging_post_authors.count).to eq(2)
```
Throne3d: I'd suggest combining this line and the line above to do `expect(reply_post.tagging_authors).to match_array([reply_post.user, user])`

spec/controllers/replies_controller_spec.rb
```diff
@@ -229,6 +229,60 @@
...
+      post_author = reply_post.tagging_post_authors.find_by(user: user)
+      expect(post_author.user).to eq(user)
+      expect(post_author.joined).to eq(true)
```
Throne3d: Maybe check `joined_at`? And that `invited_at`/`invited_by` aren't set to anything?

spec/controllers/replies_controller_spec.rb
```diff
@@ -229,6 +229,60 @@
...
+      user = create(:user)
+      login_as(user)
+      reply_post = create(:post)
+      expect(reply_post.tagging_authors.count).to eq(1)
+      old_reply = create(:reply, post: reply_post, user: user)
+      reply_post.reload
+      expect(reply_post.tagging_authors).to include(user)
+      expect(reply_post.tagging_authors.count).to eq(2)
+      expect(reply_post.joined_authors).to include(user)
+      expect(reply_post.joined_authors.count).to eq(2)
```
Throne3d: Same as above on matching exact authors

spec/controllers/replies_controller_spec.rb
```diff
@@ -229,6 +229,60 @@
...
+    end
+
+    it "handles trying to reply to a closed thread as a non-author correctly" do
+      user = create(:user)
+      login_as(user)
+      reply_post = create(:post, authors_locked: true)
+      post :create, params: { reply: {post_id: reply_post.id, content: 'test'} }
+      expect(flash[:error][:message]).to eq("Your reply could not be saved because of the following problems:")
+      expect(flash[:error][:array]).to eq(["Post is not a valid post author"])
```
Throne3d: This error message seems weird. Note to look into fixing it? ("Post's user is not a valid post author"?)

spec/models/post_spec.rb
```diff
@@ -859,16 +859,89 @@
...
+      expect(post.authors).to eq([post.user])
+      author = post.post_authors.first
+      expect(author.joined).to eq(true)
+      expect(author.joined_at).to be_the_same_time_as(post.created_at)
+    end
+
+    it "invites coauthors on creation" do
+      invited = create(:user)
+      post = create(:post, unjoined_author_ids: [invited.id])
+      expect(post.authors.count).to eq(2)
```
Throne3d: match exact authors?

Throne3d: ditto for the rest of these tests (think it's a leftover from previous tests)

spec/models/reply_spec.rb
```diff
@@ -187,4 +192,23 @@
       include_examples 'has_edit_audits', method
     end
   end
+
+  describe "authors interactions" do
+    it "does not update can_owe upon creating a reply" do
+      post = create(:post)
+      reply = create(:reply, post: post)
+
+      expect(post.author_for(reply.user).can_owe).to be true
```
Throne3d: Style's weird here – `be(true)` would be consistent?

Throne3d:
After my comments are addressed I think there's just the following list to finish (plus what's in the PR body itself):

- [ ] Some way to remove post authors from being able to tag
- [ ] The `authors_locked` checkbox should automatically check itself if you add authors on `#new` or `#edit`.
- [ ] It should also be checked if the page is loaded with authors set to the board's (but presumably not just "if the post has authors").
  - [ ] When changing the board for a post to a closed board, somehow prompt the user if they want to set the board authors to be the authors? Or just do it if the list is empty?
- [ ] How do moderators interact with this all?
- [ ] Perhaps we should notify people by email when they're invited to a thread
- Unit tests for:
  - [ ] `Board#writer_ids`? This is used when creating a post in a specific board.
  - [ ] `Post#metadata_editable_by?` with new changes
  - [ ] `Post#taggable_by?` with new changes (if not yet tested)
- [ ] `Post#destroy` should destroy PostAuthors – do we want to check this?
- [ ] `Reply#destroy` should un-join a user from a post if it's their only reply. (e.g. you accidentally respond to a post you shouldn't?) I think this currently isn't done

app/views/posts/owed.haml
```diff
@@ -1,3 +1,22 @@
 - content_for :posts_title do
-  Threads You Owe Tags In
-= render partial: 'list', locals: {posts: @posts}
+  - if params[:view] == 'hidden'
+    Threads with Replies Owed Disabled
```
Throne3d: Capital W, since we don't seem to be lowercasing prepositions in titles (e.g. "Threads You Owe Tags/Replies In")

spec/controllers/replies_controller_spec.rb
teceler: What Throne said

spec/controllers/replies_controller_spec.rb
teceler: ...yeah that is a weird error message, seconding throne again here

app/controllers/posts_controller.rb
```diff
@@ -16,15 +16,19 @@ def index
...
-    @posts = Post.where(id: ids).where.not(status: [Post::STATUS_COMPLETE, Post::STATUS_ABANDONED]).where.not(last_user: current_user)
-    @posts = @posts.where.not(status: Post::STATUS_HIATUS).where('tagged_at > ?', 1.month.ago) if current_user.hide_hiatused_tags_owed?
-    @posts = posts_from_relation(@posts.order('tagged_at desc'))
     @show_unread = true
     @hide_quicklinks = true
     @page_title = 'Replies Owed'
+
+    if params[:view] == 'hidden'
+      ids = PostAuthor.where(user_id: current_user.id, can_owe: false).group(:post_id).pluck(:post_id)
+      @posts = posts_from_relation(Post.where(id: ids)) and return
```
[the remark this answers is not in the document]
Throne3d: I think it's intended, so you can opt yourself back in without needing to currently owe a tag.

app/views/posts/show.haml
```diff
@@ -148,9 +159,11 @@
 - if logged_in? && @unread.nil?
   <a id="unread" class="noheight"> </a>
 - if @post.taggable_by?(current_user)
-  - if @post.author_ids.include?(current_user.id) || (!@post.authors_locked? && !@reply.content.nil?)
+  - post_open = !@post.authors_locked?
+  - draft_exists = !@reply.content.nil?
+  - if (post_open && draft_exists) || @post.author_ids.include?(current_user.id)
```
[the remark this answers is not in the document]
Throne3d: No, I don't think so – `@reply.content` might be blank (`""`); this specifically checks if it's `nil`, in this context _not initialized_, to see if there's a draft.

We should maybe alter how we check if there's a draft, though.

spec/controllers/posts_controller_spec.rb
```diff
@@ -317,14 +341,15 @@
...
         # tags
-        expect(assigns(:post).settings.size).to eq(3)
-        expect(assigns(:post).content_warnings.size).to eq(3)
-        expect(assigns(:post).labels.size).to eq(3)
-        expect(assigns(:post).settings.map(&:id_for_select)).to match_array([setting1.id, setting2.id, '_other'])
-        expect(assigns(:post).content_warnings.map(&:id_for_select)).to match_array([warning1.id, warning2.id, '_other'])
-        expect(assigns(:post).labels.map(&:id_for_select)).to match_array([label1.id, label2.id, '_other'])
+        expect(assigns(:post).settings.size).to eq(0)
+        expect(assigns(:post).content_warnings.size).to eq(0)
+        expect(assigns(:post).labels.size).to eq(0)
```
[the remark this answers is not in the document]
Throne3d: Since we're no longer storing these on the post until we actually save, we don't want to amend the relations on the post, just the local variables (just below this).

spec/controllers/posts_controller_spec.rb
```diff
@@ -1415,12 +1589,12 @@
...
+        expect(assigns(:settings).map(&:id_for_select)).to match_array([setting1.id, setting2.id, '_other'])
+        expect(assigns(:content_warnings).map(&:id_for_select)).to match_array([warning1.id, warning2.id, '_other'])
+        expect(assigns(:labels).map(&:id_for_select)).to match_array([label1.id, label2.id, '_other'])
```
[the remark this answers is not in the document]
Throne3d: Yeah, I think this is the same as the above comment.

app/controllers/posts_controller.rb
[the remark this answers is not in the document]
teceler: yeah, that makes sense, just mildly unintinuitive. Might want to have the header specify that somehow? Both so that people don't think it's a bug and so it's more likely to occur to people they can do that.

app/controllers/posts_controller.rb
Marri: I do not want to delete comments On Principle but note for future selves that I did check and we do have all necessary tests for associations doing the right things on create/update preview/save.

app/models/reply.rb
```diff
@@ -118,4 +118,15 @@ def update_post_authors
     return if is_import
     post.user_joined(user)
   end
+
+  def remove_post_author
+    return if post.user == user
+    return unless post.replies.where(user: user).blank?
```
Throne3d: `return if post.replies.where(user: user).exists?` doesn't load the objects, so would be preferable.

app/models/reply.rb
```diff
@@ -118,4 +118,15 @@ def update_post_authors
...
+    post_author = post.post_authors.find_by(user: user)
+    if post.authors_locked == false
```
Throne3d: Can we do this based off `can_owe` again?

```ruby
if post_author.can_owe?
  post_author.update_attributes(joined: false)
else
  post_author.destroy
end
```

teceler: joining a thread stets can_owe to true

Throne3d: in that case, can we still flip this if-conditional:

```ruby
if post.authors_locked?
  post_author.update_attributes(joined: false)
else
  post_author.destroy
end
```

spec/controllers/posts_controller_spec.rb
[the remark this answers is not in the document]
Marri: For posterity: yes. This is now stored exclusively in an assigns rather than being pulled from the object, because it shouldn't change on the object _unless we expect it to have changed on the object_ and that is kind of explicitly not what preview is doing, so, assigns.

Throne3d:
Favorites by author should probably use the new PostAuthor system, instead of checking through all relevant replies (see <https://github.com/Marri/glowfic/blob/master/app/controllers/favorites_controller.rb#L18>).

spec/models/post_spec.rb
```diff
@@ -921,28 +921,29 @@
...
         create(:reply, user: coauthor, post: post)
-        post.author_for(coauthor).update_attributes(can_owe: false)
+        post.opt_out_of_owed(coauthor)
         expect(post).to be_taggable_by(coauthor)
       end
     end
 
     context "with open post" do
-      let(:post) { create(:post, user: poster) }
+      let(:post) { create(:post, user: poster, unjoined_authors: [poster, coauthor], authors_locked: false) }
```
Throne3d: poster has joined, only `coauthor` should be in this array (same below)

spec/controllers/posts_controller_spec.rb
```diff
@@ -1998,16 +1998,65 @@
...
+        login_as(coauthor)
+        post = create(:post, user: user, authors: [user, coauthor], authors_locked: true)
+        put :update, params: {
+          id: post.id,
+          post: {
+            content: "newtext"
+          }
+        }
+        expect(response).to redirect_to(post_url(post))
+        expect(flash[:error]).to eq("You do not have permission to modify this post.")
```
Throne3d: Needs a line to check content hasn't changed, but it's skipped anyway so this isn't major.

Throne3d:
See #703 for follow-up issue.

PR #665, opened 2018-01-08
Convert import permissions to use Permissible

Throne3d:
We'll need to manually:

```ruby
User.where(id: [1, 2, 3, 8, 19, 24, 31], role_id: nil).update_attributes(role_id: Permissible::IMPORTER)
```

This should be neater than the current hardcoding method.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -7,6 +7,7 @@ class PostsController < WritableController
   before_action :login_required, except: [:index, :show, :history, :warnings, :search, :stats]
   before_action :find_post, only: [:show, :history, :stats, :warnings, :edit, :update, :destroy]
   before_action :require_permission, only: [:edit]
+  before_action :require_import_permission, only: [:new, :create]
   before_action :editor_setup, only: [:new, :edit]
 
   def index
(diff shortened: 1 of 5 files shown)


app/controllers/posts_controller.rb
```diff
@@ -88,7 +92,10 @@ def new
   end
 
   def create
-    import_thread and return if params[:button_import].present?
+    if params[:button_import].present?
```
Marri: Why did this change?

app/controllers/posts_controller.rb
```diff
@@ -375,6 +379,13 @@ def require_permission
     end
   end
 
+  def require_import_permission
```
Marri: What if we just made this a before_action and returned unless view=import || button_import.present? Might be cleaner?

app/controllers/posts_controller.rb
Throne3d: `import_thread` stopped returning a truthy value all the time, so in some circumstances `import_thread and return` was failing to return.

I could just make this line: https://github.com/Marri/glowfic/pull/665/files#diff-ab4e9273f7df75aa5004a42f757d6f51R313 into `return true if performed?`?

app/controllers/posts_controller.rb
Throne3d: Done! Looks to solve the other comment too.

app/controllers/posts_controller.rb
Throne3d: (Solved by converting `require_import_permission` to a `before_action`.)

PR #666, opened 2018-01-09
Fix moderator note being lost on reply preview

### LINKED ISSUE #657: Previewing moderator-edited reply doesn't show moderator note
If you moderator-edit a reply, add a mod note, then preview, the mod note doesn't exist. (The whole box doesn't exist, even, so "add a mod note" is unnecessary as a step.)

teceler:
Fixes #657 and adds a test for preview on updating replies.

### DIFF
--- a/spec/controllers/replies_controller_spec.rb
+++ b/spec/controllers/replies_controller_spec.rb
@@ -563,8 +563,10 @@ RSpec.describe RepliesController do
       end
 
       it "takes correct actions for moderators" do
-        reply_post = create(:post)
-        reply = create(:reply, post: reply_post, user: reply_post.user)
+        user = create(:user)
+        reply_post = create(:post, user: user)
+        reply = create(:reply, post: reply_post, user: user)
+        char = create(:template_character, user: user)
         login_as(create(:mod_user))
 
         newcontent = reply.content + 'new'
@@ -582,6 +584,9 @@ RSpec.describe RepliesController do
         expect(assigns(:written).user).to eq(reply.user)
         expect(assigns(:written).audit_comment).to eq('note')
         expect(assigns(:written).content).to eq(newcontent)
+
+        expect(controller.gon.editor_user[:username]).to eq(user.username)
+        expect(assigns(:templates)).to eq([char.template])
       end
 
       skip
--- a/spec/controllers/replies_controller_spec.rb
+++ b/spec/controllers/replies_controller_spec.rb
@@ -562,6 +562,28 @@ RSpec.describe RepliesController do
         expect(templateless.plucked_characters).to eq([[char.id, char.name]])
       end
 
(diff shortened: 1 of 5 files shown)


spec/controllers/replies_controller_spec.rb
```diff
@@ -482,6 +482,84 @@
...
     context "preview" do
+      it "takes correct actions" do
+        user = create(:user)
+        reply_post = create(:post, user: user)
+        reply = create(:reply, post: reply_post, user: user)
+        login_as(user)
+        expect(ReplyDraft.count).to eq(0)
+
+        char = create(:character, user: user)
+        icon = create(:icon)
```
Throne3d: `, user: user`

spec/controllers/replies_controller_spec.rb
```diff
@@ -482,6 +482,84 @@
     end
 
     context "preview" do
+      it "takes correct actions" do
+        user = create(:user)
+        reply_post = create(:post, user: user)
+        reply = create(:reply, post: reply_post, user: user)
+        login_as(user)
+        expect(ReplyDraft.count).to eq(0)
```
Throne3d: This should be the case after the submission, too; can you check that then?

spec/controllers/replies_controller_spec.rb
```diff
@@ -482,6 +482,84 @@
...
+        persisted = written.reload
+        expect(persisted.user).to eq(reply_post.user)
+        expect(persisted.character).to be_nil
+        expect(persisted.icon).to be_nil
+        expect(persisted.character_alias).to be_nil
+        expect(assigns(:post)).to eq(reply_post)
+        expect(assigns(:reply)).to eq(reply)
+
+        # build_template_groups:
+        expect(controller.gon.editor_user).not_to be_nil
```
Throne3d: Would be nice to check this is the right user. I assume you copied this from somewhere? I'll amend it here for now.

spec/controllers/replies_controller_spec.rb
```diff
@@ -482,6 +482,84 @@
...
+          reply: {
+            content: newcontent,
+            audit_comment: 'note'
+          }
+        }
+
+        expect(response).to render_template(:preview)
+        expect(assigns(:written).user).to eq(reply.user)
+        expect(assigns(:written).audit_comment).to eq('note')
+        expect(assigns(:written).content).to eq(newcontent)
```
Throne3d: Would be good to check `controller.gon.editor_user` and `assigns(:templates)` here, briefly, too.

Throne3d:
@Throne3d: feature spec this sometime to make sure it works. preferably tomorrow so it can merge.

Throne3d:
Unrelated specs have been pulled out into #686 and #687, so this now contains (almost only) the stuff relating to moderator previews.

Issue #59, opened 2017-02-04
Reply reordering


Throne3d:
From the title: allow replies to be reordered

Throne3d:
Possibly related: splitting large posts into smaller sub-posts? (e.g. Room of Requirement, and also where authors forget to start a new thread in a continuity.)

It's been used a couple of times so far and could be useful if there were some actual built-in command to achieve the effect, possibly with UI somewhere.

Throne3d:
We've really had the distinction between posts and replies for too long, by this point – we have a lot of ugly code designed to work around the fact they're different, but we want to treat them the same, etc, and really we should just get around to fixing it.

Bumping the priority of this issue to critical so we have the prerequisite for #421, and separating off the "split replies" part of it (which will be easy after this is implemented).

Throne3d:
This gets a lot easier if we allow them to be reordered but _maintain the assumption that created\_at timestamps are monotonic with order_. So if we let people split replies (and have them equal in creation time)‚ or split the reply part of a post into a reply, this is safe and a lot easier.

First version of a fix to this issue should allow for that. Later, after examining more code paths, we can relax this assumption. (Unread in particular is going to be difficult around it, and we may need a Major Edit feature to accommodate upstream insertion.)

PR #667, opened 2018-01-09
Add reply_order

teceler:
Depends on #688

### DIFF
--- a/app/concerns/orderable.rb
+++ b/app/concerns/orderable.rb
@@ -6,6 +6,16 @@ module Orderable
     after_save :reorder_others_after
     after_destroy :reorder_others_before
 
+    scope :ordered_manually, -> { order('section_order asc') }
+
+    def order
+      section_order
+    end
+
+    def order=(val)
+      self.section_order = val
+    end
+
     private
 
     def reorder_others(is_after)
@@ -20,12 +30,12 @@ module Orderable
       other_where = Hash[ordered_attributes.map do |atr|
         [atr, send(is_after ? "#{atr}_before_last_save" : "#{atr}_was")]
       end]
-      others = self.class.where(other_where).order('section_order asc')
+      others = self.class.where(other_where).ordered_manually
       return unless others.present?
 
       others.each_with_index do |other, index|
-        next if other.section_order == index
-        other.section_order = index
+        next if other.order == index
+        other.order = index
         other.save
       end
     end
@@ -40,7 +50,7 @@ module Orderable
 
     def autofill_order
       return unless new_record? || order_change?(false)
-      self.section_order = ordered_items.count
+      self.order = ordered_items.count
     end
 
     def order_change?(is_after)
--- a/db/migrate/20180109003825_add_reply_order_to_replies.rb
+++ b/db/migrate/20180109003825_add_reply_order_to_replies.rb
@@ -1,0 +1,14 @@
+class AddReplyOrderToReplies < ActiveRecord::Migration[5.1]
+  def up
+    add_column :replies, :reply_order, :integer
+    Post.find_each do |post|
+      post.replies.order(id: :asc).each_with_index do |reply, i|
+        reply.update_attributes(reply_order: i)
+      end
+    end
+  end
+
+  def down
+    remove_column :replies, :reply_order, :integer
+  end
+end
--- a/db/migrate/20180109003825_add_reply_order_to_replies.rb
+++ b/db/migrate/20180109003825_add_reply_order_to_replies.rb
@@ -1,11 +1,16 @@
 class AddReplyOrderToReplies < ActiveRecord::Migration[5.1]
   def up
     add_column :replies, :reply_order, :integer
-    Post.find_each do |post|
-      post.replies.order(id: :asc).each_with_index do |reply, i|
-        reply.update_attributes(reply_order: i)
-      end
-    end
+    execute <<-SQL
+WITH v_replies AS
+(
+  SELECT ROW_NUMBER() OVER(PARTITION BY replies.post_id ORDER BY replies.id asc) AS rn, id FROM replies
+)
+UPDATE replies
+SET reply_order = v_replies.rn-1
+FROM v_replies
+WHERE replies.id = v_replies.id;
+    SQL
   end
 
   def down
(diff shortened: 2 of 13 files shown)


Throne3d:
(Addresses #59)

spec/models/reply_spec.rb
```diff
@@ -187,4 +187,22 @@
...
+      second_reply = create(:reply, post: post)
+      third_reply = create(:reply, post: post)
+      expect(post.replies.ordered).to eq([first_reply, second_reply, third_reply])
+    end
+
+    it "puts replies in the correct order even if they are created in the wrong order" do
+      first_reply = create(:reply, post: post)
+      second_reply = Timecop.freeze(first_reply.created_at - 5.seconds) { create(:reply, post: post) }
+      third_reply = Timecop.freeze(first_reply.created_at - 3.seconds) { create(:reply, post: post) }
+      expect(post.replies.ordered).to eq([first_reply, second_reply, third_reply])
```
Throne3d: Maybe compare this to `post.replies.order(:created_at)` to show the difference (and ensure it's all working as expected)?

spec/models/reply_spec.rb
```diff
@@ -187,4 +187,22 @@
...
+    end
+  end
```
Throne3d: Could we have another test for:

```ruby
it "orders replies by reply order not ID" do
  reply1 = create(:reply, post: post)
  reply2 = create(:reply, post: post)
  reply3 = create(:reply, post: post)
  reply2.update_columns(reply_order: 2)
  reply3.update_columns(reply_order: 1)
  expect(post.replies.ordered).to eq([reply1, reply3, reply2])
end
```

The second test would work on master (well, if we added a scope for `ordered`), since that orders by ID – it doesn't actually test that we're now ordering by _reply order_ instead of _ID_.

teceler: this test was originally supposed to do that and then I discovered that it wouldn't let me pass reply_order to the factory. that sounds like a way to do it!

app/views/reports/_daily.haml
```diff
@@ -32,7 +32,7 @@
         %tr
           - klass = [cycle('even', 'odd')]
           - klass << 'post-ignored' if ignored?(post)
-          - replies_today = post.replies.where(created_at: @day.beginning_of_day .. @day.end_of_day).order('created_at asc')
+          - replies_today = post.replies.where(created_at: @day.beginning_of_day .. @day.end_of_day).ordered
```
Throne3d: This is technically more of a `created_at` thing (at least we figure out how we're gonna deal with non-monotonic `created_at` timestamps), so I'd leave it as it was before. (It's used for a "last in day" link, and to say when the post was last updated, so it seems to make sense to have it be the `created_at` and appear in a later audit.)

Throne3d: … then again there are many other places where we used `id asc` assuming the same thing and so we're gonna need to fiddle with stuff there. ugh. uh. yeah. idk, leave it for now probably.

app/concerns/orderable.rb
```diff
@@ -20,12 +20,21 @@ def reorder_others(is_after)
       other_where = Hash[ordered_attributes.map do |atr|
         [atr, send(is_after ? "#{atr}_before_last_save" : "#{atr}_was")]
       end]
-      others = self.class.where(other_where).order('section_order asc')
+      if self.is_a?(Reply)
```
Marri: This would be better if we just implemented .ordered on the other Orderable classes? Or we could define it using the concern to default to 'section_order asc' and then override it for Reply. This special casing is Not Stylistic if we can avoid it.

app/concerns/orderable.rb
```diff
@@ -20,12 +20,21 @@ def reorder_others(is_after)
...
+        others = self.class.where(other_where).ordered
+      else
+        others = self.class.where(other_where).order('section_order asc')
+      end
       return unless others.present?
 
       others.each_with_index do |other, index|
-        next if other.section_order == index
-        other.section_order = index
+        if self.is_a?(Reply)
```
Marri: Also not Good Style. We could implement a section_order= method on Reply that set reply_order, or define a class method to define this, or do the same default-overriden-in-Reply pattern above, or even just rename this field to 'order' everywhere or something. Or pass the name of the method to the concern? Many options.

app/views/reports/_daily.haml
Marri: What Throne said; for report purposes, we do actually want creation time here.

db/migrate/20180109003825_add_reply_order_to_replies.rb
```diff
@@ -0,0 +1,19 @@
+class AddReplyOrderToReplies < ActiveRecord::Migration[5.1]
+  def up
+    add_column :replies, :reply_order, :integer
```
Marri: I have a strong, strong suspicion you're going to need an index on this if you're doing ordering operations on it all over the place.

db/migrate/20180109003825_add_reply_order_to_replies.rb
```diff
@@ -0,0 +1,19 @@
+class AddReplyOrderToReplies < ActiveRecord::Migration[5.1]
```
Marri: I swear, every time we say we'll Do It Differently and every time we don't, but I really need to start, so. Migrations first, and separate, please. I want to get out of the habit of having to put the site into maintenance mode every time we release something with a db change.

Throne3d: (Moved out to #688)

app/concerns/orderable.rb
Throne3d: I'm not sure what "do the same default-overriden-in-Reply pattern above" means?

I think defining `section_order` and `section_order=` methods on Reply would be the easiest route, but we might (in the long run) want to go about renaming the fields to `order`.

Throne3d:
Do we plan to standardize on `order` for these columns, to make things easier? (Alternatively to give them more accurate/descriptive names than `section_order`?) There was an idea to look into that, and I'm not sure if it's in the scope of this PR.

Throne3d:
Okay, I'm happy with this PR – it looks like we've got everything addressed properly.

PR #670, opened 2018-01-10
Add rubocop as a development dependency

Throne3d:
I'm not sure if this is necessary, but it looks like I keep forgetting to keep it (and .rubocop.yml) up to date. Having it as a development dependency may well fix that by forcing it to update every so often.


Throne3d:
Since we have a .rubocop.yml file, actually, I'm no longer on the fence about this PR, and think we should merge it. @Marri?

PR #675, opened 2018-01-10
[Very WIP] Fix/refactor scss

teceler:
Note: do _not_ merge without making sure the stuff from `add/rspec-page-regression` is cleared out first.


teceler:
In practice this has been split out into a lot of other PRs

PR #688, opened 2018-01-20
Add reply_order migration

teceler:
Migration for #667; that depends on this.


Throne3d:
I think this is gonna fall over if we migrate it too early – won't new replies formed in the time between the migration and the code release lack a valid reply_order?

Throne3d:
( @Marri )

Throne3d:
…Totally concur with past me, this'll fall over if I migrate it and there are new replies in the meantime. Gonna merge both.

Issue #218, opened 2017-05-01
Editing a post you don't own should be marked differently


Throne3d:
A different edit icon (presumably the same pencil but with some addition of some kind?) in addition to an edit button on the stats page would be good.

Suggested [by Moriwen here](http://alicorn.elcenia.com/board/viewtopic.php?f=12&t=459&p=32881#p32880).

Throne3d:
Given that there's now an edit button on the stats page, this should probably be changed similarly to how we will change the pencil on the main post (and maybe also change the text to "Edit Metadata" instead of "Edit Post").

PR #705, opened 2018-02-11
Move embedded reply in posts to a reply object

### LINKED ISSUE #421: Split the written part of a post (reply-esque) from the metadata
This will allow handling posts and replies more consistently where we want to treat all "written tag-like objects" together. Likely requires we have a manual sorting order for posts (see #59) so we can put the first parts of posts back, or a toggle on replies to make it a "top post" or something (would prefer the former for generality purposes), then we'd use the first reply in the post as the top post and paginate the rest.

teceler:
Fixes #421 

~~Depends on #667~~

Depends on #1351 and #1331


app/controllers/posts_controller.rb
```diff
@@ -91,10 +91,12 @@ def unhide
   end
 
   def new
-    @post = Post.new(character: current_user.active_character, user: current_user)
+    @post = Post.new(user: current_user)
     @post.board_id = params[:board_id]
     @post.section_id = params[:section_id]
-    @post.icon_id = (current_user.active_character ? current_user.active_character.default_icon.try(:id) : current_user.avatar_id)
+    @post.written = @post.replies.new(user: current_user, post_id: @post.id, reply_order: 0)
```
Marri: Can we not use `@post.written.build`, did we try that already? Honestly build vs new doesn't matter a ton, but also I'm pretty sure reply_order should auto fill, and if it doesn't that seems like a Thing to Fix?

app/controllers/icons_controller.rb
```diff
@@ -45,8 +45,7 @@ def show
     @page_title = @icon.keyword
     if params[:view] == 'posts'
       arel = Post.arel_table
-      post_ids = Reply.where(icon_id: @icon.id).select(:post_id).distinct.pluck(:post_id)
-      where_calc = arel[:icon_id].eq(@icon.id).or(arel[:id].in(post_ids))
+      where_calc = arel[:id].in(Reply.where(icon_id: @icon.id).select(:post_id).distinct.pluck(:post_id))
```
Marri: We can leave this if you want but it's now really just `post_ids = Reply.where(icon_id: @icon.id).select(:post_id).distinct.pluck(:post_id)` and then `Post.where(id: post_ids)`, we don't need any of this arel stuff any more.

app/controllers/icons_controller.rb
```diff
@@ -87,8 +86,7 @@ def replace
     gon.gallery = Hash[all_icons.map { |i| [i.id, {url: i.url, keyword: i.keyword}] }]
     gon.gallery[''] = {url: view_context.image_path('icons/no-icon.png'), keyword: 'No Icon'}
 
-    post_ids = Reply.where(icon_id: @icon.id).select(:post_id).distinct.pluck(:post_id)
-    all_posts = Post.where(icon_id: @icon.id) + Post.where(id: post_ids)
+    all_posts = Post.where(id: Reply.where(icon_id: @icon.id).select(:post_id).distinct.pluck(:post_id))
```
Marri: instead of `all_posts = [thing]`, remove following line and use `@posts = [thing]`, we already did distinct to dedupe for us (and Rails will dedupe duplicate ids for you anyway, `Post.where(id: [1,1,2])` returns two objects)

app/controllers/posts_controller.rb
```diff
@@ -106,24 +108,28 @@ def new
...
 
-    unless @post.save
+    begin
+      @post.save!
+      flash[:success] = "You have successfully posted."
+      redirect_to post_path(@post)
+    rescue ActiveRecord::RecordInvalid
       flash.now[:error] = {}
       flash.now[:error][:array] = @post.errors.full_messages
       flash.now[:error][:message] = "Your post could not be saved because of the following problems:"
```
Marri: Did we check this includes reply validation errors?

app/controllers/posts_controller.rb
```diff
@@ -106,24 +108,28 @@ def new
...
+
+    preview and return if params[:button_preview].present?
     @post.settings = process_tags(Setting, :post, :setting_ids)
     @post.content_warnings = process_tags(ContentWarning, :post, :content_warning_ids)
     @post.labels = process_tags(Label, :post, :label_ids)
 
-    unless @post.save
+    begin
+      @post.save!
+      flash[:success] = "You have successfully posted."
```
Marri: I'd leave this at the end; minimizing stuff in a begin/rescue is good practice unless we need to.

app/controllers/posts_controller.rb
```diff
@@ -149,15 +155,17 @@ def update
...
     preview and return if params[:button_preview].present?
 
-    @post.assign_attributes(post_params)
     @post.board ||= Board.find(3)
     settings = process_tags(Setting, :post, :setting_ids)
     warnings = process_tags(ContentWarning, :post, :content_warning_ids)
     labels = process_tags(Label, :post, :label_ids)
 
-    if current_user.id != @post.user_id && @post.audit_comment.blank? && !@post.author_ids.include?(current_user.id)
+    if current_user.id != @post.user_id && @post.written.audit_comment.blank? && @post.audit_comment.blank? && !@post.author_ids.include?(current_user.id)
```
Marri: Why do we have two of these...? Do we give two boxes for edit notes when you update a post?

app/controllers/posts_controller.rb
```diff
@@ -234,7 +243,7 @@ def search
     if params[:character_id].present?
       arel = Post.arel_table
       post_ids = Reply.where(character_id: params[:character_id]).select(:post_id).distinct.pluck(:post_id)
-      where = arel[:character_id].eq(params[:character_id]).or(arel[:id].in(post_ids))
+      where = arel[:id].in(post_ids)
```
Marri: No longer needs to be arel

app/controllers/templates_controller.rb
```diff
@@ -29,7 +29,7 @@ def show
     character_ids = @template.characters.pluck(:id)
     post_ids = Reply.where(character_id: character_ids).select(:post_id).distinct.pluck(:post_id)
     arel = Post.arel_table
-    where = arel[:character_id].in(character_ids).or(arel[:id].in(post_ids))
+    where = arel[:id].in(post_ids)
```
Marri: No longer needs arel

app/controllers/users_controller.rb
```diff
@@ -114,10 +114,9 @@ def output
     @day = calculate_day
     daystart = ActiveRecord::Base.connection.quote(@day.beginning_of_day)
     dayend = ActiveRecord::Base.connection.quote(@day.end_of_day)
-    @posts = Post.where(user: current_user).where('created_at between ? AND ?', daystart, dayend).pluck(:content)
     @replies = Reply.where(user: current_user).where('created_at between ? AND ?', daystart, dayend).pluck(:content)
 
-    @total = @posts + @replies
+    @total = @replies
```
Marri: This can probably be cleaned up at least a little now that we don't need to do array addition

app/controllers/writable_controller.rb
```diff
@@ -66,8 +67,11 @@ def show_post(cur_page=nil)
       self.page = cur_page = cur_page.to_i
     end
 
+    read_at = @post.read_time_for(@replies) if logged_in?
```
Marri: Why did this get moved? I was reasonably sure this needed to happen after paginate?

app/views/icons/show.haml
```diff
@@ -96,7 +95,7 @@
     %tbody
       %tr
         %th.sub.width-150 Times Used
-        %td.even=(posts_using.count + replies_using.count).to_s
+        %td.even=(replies_using.count).to_s
```
Marri: Probably does not require ()s anymore

app/views/icons/show.haml
```diff
@@ -96,7 +95,7 @@
...
       %tr
         %th.sub.width-150 Posts In
-        %td.even= (posts_using.pluck(:id) + replies_using.pluck('distinct post_id')).uniq.count
+        %td.even= (replies_using.pluck('distinct post_id')).uniq.count
```
Marri: `replies_using.select(:post_id).distinct.count`

db/migrate/20180209163253_split_post_text_into_replies.rb
```diff
@@ -0,0 +1,45 @@
...
+      if post.user_id == reply.user_id
+        post.character_id = reply.character_id
+        post.character_alias_id = reply.character_alias_id
+        post.icon_id = reply.icon_id
+        post.content = reply.content
+        post.save!
+      else
+        raise "Post user does not match initial reply's user"
+      end
+      # mass-shift the reply order here to avoid it doing them one at a time
```
Marri: ping

spec/models/template_spec.rb
```diff
@@ -53,7 +53,7 @@
       character1 = create(:character, template: template)
       character2 = create(:character, template: template, template_name: "nickname")
       character3 = create(:character, template: template, screenname: "screen_name")
-      expect(template.plucked_characters).to eq([[character1.id, character1.name], [character2.id, "#{character2.name} | #{character2.template_name}"], [character3.id, "#{character3.name} | #{character3.screenname}"]])
+      expect(template.plucked_characters).to match_array([[character1.id, character1.name], [character2.id, "#{character2.name} | #{character2.template_name}"], [character3.id, "#{character3.name} | #{character3.screenname}"]])
```
Marri: Is plucked_characters not a consistent order?

spec/factories.rb
```diff
@@ -41,19 +41,29 @@
...
+      character { nil }
+      character_alias { nil }
+      icon { nil }
+      content { nil }
     end
     user
     board
-    description { "" }
-    content { "test content" }
+    description ""
```
Marri: Aren't all values supposed to be in {}s? I thought that was a change we got bothered into by something?

spec/controllers/replies_controller_spec.rb
```diff
@@ -622,13 +629,13 @@
...
       reply = create(:reply, post: reply_post)
-      expect(reply.reply_order).to eq(1)
+      order = reply.reply_order
       expect(reply_post.replies.ordered.last).to eq(reply)
       create(:reply, post: reply_post)
       expect(reply_post.replies.ordered.last).not_to eq(reply)
       reply_post.mark_read(reply_post.user)
       put :update, params: { id: reply.id, reply: {content: 'new content'} }
-      expect(reply.reload.reply_order).to eq(1)
+      expect(reply  .reload.reply_order).to eq(order)
```
Marri: Why space, why new variable?

spec/controllers/replies_controller_spec.rb
```diff
@@ -334,11 +340,12 @@
     it "sets reply_order correctly on the first reply" do
       reply_post = create(:post)
       login_as(reply_post.user)
+      reply_post.mark_read(reply_post.user)
```
Marri: Why is this necessary?

teceler: Because otherwise it goes 'OH NO YOU HAVE AN UNREAD REPLY' when you try to add one >.>

app/controllers/characters_controller.rb
```diff
@@ -185,7 +185,7 @@ def replace
     end
     @alt = @alts.first
 
-    all_posts = Post.where(character_id: @character.id) + Post.where(id: Reply.where(character_id: @character.id).select(:character_id).distinct.pluck(:post_id))
+    all_posts = Post.where(id: Reply.where(character_id: @character.id).select(:character_id).distinct.pluck(:post_id))
```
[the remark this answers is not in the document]
teceler: ...I think I have another PR to fix that actually xD

app/controllers/posts_controller.rb
teceler: ...I don't know. ...I might have just assumed we'd have one tbqh

Marri:
down migration needs search indices added back in

spec/models/template_spec.rb
teceler: I seem to have lost this change in the rebase and don't have any new failures there, so I... guess... it is? idk why I thought it wasn't.

app/controllers/posts_controller.rb
teceler: it looks like it doesn't set the reply_order until it/the post actually saves, which presumably causes problems in preview?

spec/controllers/replies_controller_spec.rb
teceler: We didn't need to do that before because the code to check for unread replies doesn't trip on the _post_. I guess I could change it to ignore the first reply too.

app/controllers/posts_controller.rb
teceler: We don't currently have a way to get an invalid reply in this context, since the user will always be the same as the post user and that's all we currently validate! I'll add some code to make sure we get reply errors if any exist though.

app/controllers/characters_controller.rb
```diff
@@ -378,8 +373,7 @@ def og_data
     linked << "Setting".pluralize(settings.count) + ": " + settings.join(', ') if settings.present?
     desc = [linked.join('. ')].reject(&:blank?)
     desc << generate_short(@character.description) if @character.description.present?
-    reply_posts = Reply.where(character_id: @character.id).select(:post_id).distinct.pluck(:post_id)
-    posts_count = Post.where(character_id: @character.id).or(Post.where(id: reply_posts)).uniq.count
+    posts_count = Post.where(id: Reply.where(character_id: @character.id).select(:post_id).distinct.pluck(:post_id)).count
```
Throne3d:
```rb
Reply.where(character_id: @character.id).select(:post_id).distinct.pluck(:post_id)
```
will by itself provide a list of distinct post IDs with the character, so we can just put a count on that (and `.distinct.count` might work also, I dunno)

app/services/post_scraper.rb
```diff
@@ -152,11 +152,12 @@ def import_post_from_doc(doc)
     content = doc.at_css('.entry-content').inner_html
 
     @post = Post.new
+    @written = @post.replies.new
```
Throne3d: Do we need `@post.written = @written = @post.replies.new` here? I think we do that in the controller.

app/services/post_scraper.rb
```diff
@@ -167,12 +168,14 @@ def import_post_from_doc(doc)
     end
 
     set_from_username(@post, username)
```
Throne3d: `set_from_username` tries to set a character, which will fail on a `@post`, right? Should we `set_from_username(@written, username)` and then `@post.user = @written.user`?

Throne3d: Oh. You check if it's a Post later. I guess.

app/views/posts/_write.haml
```diff
@@ -11,18 +11,22 @@
...
+          - unless current_user.id == post.user_id
+            %b= w.label :audit_comment, 'Moderator note'
+            %br
+            = w.text_area :audit_comment, placeholder: 'Explain reason for moderation here', class: 'mod'
+            %br
+          = w.hidden_field :character_id, id: 'reply_character_id'
+          = w.hidden_field :icon_id, id: 'reply_icon_id'
+          = w.hidden_field :character_alias_id, id: 'reply_character_alias_id'
+          = submit_tag (post.new_record? ? 'Post' : 'Save'), class: 'button', id: 'submit_button', data: { disable_with: 'Saving...' }
+          = submit_tag "Preview", class: 'button', id: 'preview_button', name: 'button_preview', data: { disable_with: 'Previewing...' }
```
Throne3d: Looks like we have duplicate buttons between here and the next few lines?

app/views/posts/_write.haml
```diff
@@ -79,4 +83,4 @@
     #post-editor.padding-10= yield :form
     .subber.centered= yield :buttons
   - else
-    = render 'editor', item: post
+    = render 'editor', item: post.replies.first
```
Throne3d: `post.written`*, for consistency? Plus this might not inherently order `post.replies` so we may have issues with arbitrary database ordering.

Throne3d:
I'm not confident that post history is handled correctly:

https://github.com/Marri/glowfic/blob/c78d4df59d88355c315f22413344a8f6b7634880/app/views/writable/_history.haml#L65

Looks like we're gonna be trying to render a `@post` using `replies/single`?

app/views/replies/_single.haml
```diff
@@ -1,7 +1,8 @@
-- if reply.is_a?(Post)
-  - write = 'post-post'.freeze
-- elsif reply.is_a?(Reply)
-  - write = 'post-reply'.freeze
+- if reply.is_a?(Reply)
```
Throne3d: This should _only_ take a reply – especially given that, on line 16, we have an unguarded reference to `reply.icon_id`, which won't exist on a post. Right? (See also <https://github.com/Marri/glowfic/pull/705#issuecomment-513939874>)

app/views/replies/_single.haml
```diff
@@ -32,13 +33,20 @@
...
-          = image_tag "icons/pencil.png".freeze, title: 'Edit'.freeze, alt: 'Edit'.freeze
-      - if reply.id.present? && reply.deletable_by?(current_user)
-        = link_to path_for(reply, '%s'), method: :delete, data: { confirm: "Are you sure you wish to delete this #{reply.class.to_s.downcase}?" } do
-          = image_tag "icons/cross.png".freeze, title: 'Delete'.freeze, alt: 'Delete'.freeze
-      - if reply.id.present? && reply.is_a?(Reply) && logged_in?
+      - if reply.id.present?
+        - if reply.order == 0 && reply.post.metadata_editable_by?(current_user)
+          = link_to path_for(reply.post, 'edit_%s') do
+            = image_tag "icons/pencil.png".freeze, title: 'Edit'.freeze, alt: 'Edit'.freeze
+        - elsif reply.editable_by?(current_user)
```
Throne3d: I don't love the duplication of this "Edit" button, can we re-combine the condition? Either the reply is editable by the user, or the "reply" is the `post.written` and the post is editable by the user.

Or, I guess, we could segue this into fixing #218 sometime, using a different pencil? Maybe leave a comment in the codebase about why it's split up?

Throne3d: …Oh, I see why it's separate – different paths. That's fair. Never mind.

app/views/replies/_single.haml
```diff
@@ -32,13 +33,20 @@
...
+          = link_to path_for(reply, 'edit_%s') do
+            = image_tag "icons/pencil.png".freeze, title: 'Edit'.freeze, alt: 'Edit'.freeze
+        - if reply.reply_order == 0  && reply.post.deletable_by?(current_user)
```
Throne3d: (Super minor: There's a double space before the `&&`.)

db/migrate/20180209163253_split_post_text_into_replies.rb
```diff
@@ -0,0 +1,50 @@
+class SplitPostTextIntoReplies < ActiveRecord::Migration[5.1]
+  def up
+    Post.find_each do |post|
+      post.replies.ordered.each { |reply| reply.update_columns(reply_order: reply.reply_order+1) }
```
Throne3d: This is going to be super slow. Uhhh, since this all goes on in a transaction, put something like this outside the loop above? (I'd appreciate it being tested in practice :P)

```rb
execute <<-SQL
UPDATE replies
SET reply_order = reply_order + 1;
SQL
```

db/migrate/20180209163253_split_post_text_into_replies.rb
```diff
@@ -0,0 +1,50 @@
...
+        user_id: post.user_id,
+        character_id: post.character_id,
+        character_alias_id: post.character_alias_id,
+        icon_id: post.icon_id,
+        content: post.content,
+        created_at: post.created_at,
+        updated_at: post.updated_at,
+        skip_post_update: true
+      )
+      reply.update_columns(reply_order: 0, created_at: post.created_at)
```
Throne3d: Might want to set `updated_at` to be same as the post's `edited_at` (or whatever the right one is), so we don't spawn a ton of history changes.

db/migrate/20180209163253_split_post_text_into_replies.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+    Post.auditing_enabled = false
+    change_table :posts do |t|
+      t.remove :character_id, :character_alias_id, :icon_id, :content
+    end
+    Post.auditing_enabled = true
```
Throne3d: Does this need auditing disabled?

db/migrate/20180209163253_split_post_text_into_replies.rb
```diff
@@ -0,0 +1,50 @@
...
+        post.character_id = reply.character_id
+        post.character_alias_id = reply.character_alias_id
+        post.icon_id = reply.icon_id
+        post.content = reply.content
+        post.without_auditing { post.save! }
+      else
+        raise "Post user does not match initial reply's user"
+      end
+      post.replies.ordered.each { |reply| reply.update_columns(reply_order: reply.reply_order-1) }
+      reply.destroy!
```
Throne3d: I'd use `reply.delete` so we don't run callbacks.

I'd also execute SQL for the replies' reply_order, outside the end of this loop:

```
execute <<-SQL
UPDATE replies
SET reply_order = reply_order - 1;
SQL
```

spec/controllers/api/v1/replies_controller_spec.rb
```diff
@@ -17,7 +17,7 @@
     end
 
     it "succeeds with valid post", :show_in_doc do
-      post = create(:post, num_replies: 2, with_icon: true, with_character: true)
+      post = create(:post, num_replies: 1, with_icon: true, with_character: true)
```
Throne3d: This test suggests we're now showing `post.written` in the API, where we didn't before. Is that correct behavior?

spec/controllers/icons_controller_spec.rb
```diff
@@ -499,15 +499,15 @@
         user = create(:user)
         icon = create(:icon, user_id: user.id)
         alts = Array.new(5) { create(:icon, user_id: user.id) }
-        post = create(:post, user_id: user.id, icon_id: icon.id)
+        post = create(:post, user_id: user.id, icon: icon)
```
Throne3d: We're creating a post with a particular icon? Does this need to be changed?

spec/controllers/icons_controller_spec.rb
```diff
@@ -499,15 +499,15 @@
...
         create(:reply, post_id: post.id, user_id: user.id, icon_id: icon.id) # post reply
         reply = create(:reply, user_id: user.id, icon_id: icon.id)
 
         other_icon = create(:icon, user_id: user.id)
         gallery = create(:gallery, user_id: user.id, icon_ids: [other_icon.id])
         expect(gallery.icons).to match_array([other_icon])
-        create(:post, user_id: user.id, icon_id: other_icon.id) # other post
-        create(:reply, user_id: user.id, icon_id: other_icon.id) # other reply
+        create(:post, user_id: user.id, icon: other_icon) # other post
```
Throne3d: Same as earlier

spec/controllers/icons_controller_spec.rb
Throne3d: Never mind, I see we've amended the factory.

spec/controllers/posts_controller_spec.rb
```diff
@@ -106,12 +107,6 @@
         expect(assigns(:search_results)).to match_array([post1, post2])
       end
 
-      it "does not mix up subject with content" do
-        create(:post, subject: 'unrelated', content: 'contains stars')
```
Throne3d: Shouldn't we still have this test, just adding a reply in?

app/controllers/characters_controller.rb
teceler: oh that's clever. yeah that makes sense

spec/controllers/posts_controller_spec.rb
```diff
@@ -638,6 +661,10 @@
       expect(PostTag.count).to eq(0)
     end
 
+    it "handles invalid first reply" do
+      # TODO: Add a test for this if it becomes possible to have an invalid reply here
```
Throne3d: This should explicitly say `skip` in it, as currently it'll act as a passing test.

app/services/post_scraper.rb
teceler: I have no idea

app/services/post_scraper.rb
teceler: Huh, I wonder why the tests didn't catch that

app/views/posts/_write.haml
teceler: I think that failed on preview somehow

spec/controllers/posts_controller_spec.rb
```diff
@@ -944,7 +975,7 @@
       end
 
       it "works for unread" do
-        third_reply = post.replies.ordered.limit(3).last
+        third_reply = post.replies.ordered.limit(4).last
```
Throne3d: *squints* I think this is right? Looks a bit confusing, though, maybe we should just do:
```rb
replies = post.replies.ordered.to_a
third_reply = replies[3]
second_last_reply = replies[-2]
```
or something like that?

spec/controllers/replies_controller_spec.rb
```diff
@@ -61,13 +61,33 @@
         expect(controller.gon.editor_user[:username]).to eq(user.username)
         # templates
         templates = assigns(:templates)
-        expect(templates.length).to eq(2)
-        template = templates.first
-        expect(template).to eq(char2.template)
+        expect(templates.length).to eq(3)
+        used = templates.first
+        expect(used.name).to eq("Thread characters")
```
Throne3d: Hrrrm, do we want this? It's probably a pain to work around, and it's probably okay as is.

spec/controllers/replies_controller_spec.rb
```diff
@@ -1014,7 +1037,7 @@
         replies = Array.new(4) { create(:reply) }
         filtered_reply = replies.last
         get :search, params: { commit: true, post_id: filtered_reply.post_id }
-        expect(assigns(:search_results)).to match_array([filtered_reply])
+        expect(assigns(:search_results)).to match_array([filtered_reply.post.written, filtered_reply])
```
Throne3d: We want this, right? For searches to find the main post? This is good!

spec/features/posts/flat_spec.rb
```diff
@@ -9,7 +9,7 @@
 
     visit post_path(post, view: 'flat')
     expect(page).to have_selector('.post-container', count: 1)
-    expect(page).to have_selector('.post-post', count: 1)
+    expect(page).to have_selector('.post-reply', count: 1)
```
Throne3d: This seems like a fair change, but I want to note that it _is_ different from posts#show, and we should maybe look into renaming things appropriately.

spec/models/post_spec.rb
```diff
@@ -17,15 +17,33 @@
...
 
+      it "should update edited_at and tagged_at when status edited" do
+        Timecop.freeze(time) do
+          post.status = Post::STATUS_COMPLETE
+          post.save!
+          expect(post.tagged_at).to be > old_tagged_at
+          expect(post.edited_at).to be > old_edited_at
+        end
+      end
+
```
Throne3d: How does `edited_at` differ from `updated_at`, now? I think it doesn't. Maybe we should look into removing it in a follow-up PR.

spec/models/post_spec.rb
```diff
@@ -58,12 +76,21 @@
         old_tagged_at
       }
 
-      it "should update edited_at but not tagged_at when content edited" do
+      it "should update edited_at and but not tagged_at when subject edited" do
```
Throne3d: and not tagged_at*

spec/models/post_spec.rb
```diff
@@ -58,12 +76,21 @@
...
+          expect(post.edited_at).to be > post.created_at
           expect(post.tagged_at).to be_the_same_time_as(old_tagged_at)
-          expect(post.edited_at).to be > old_edited_at
+        end
+      end
+
+      it "should not update when content edited" do
+        Timecop.freeze(time) do
+          post.written.content = 'newer content'
+          post.save!
```
Throne3d: What's up with this saving the post? Is this meant to be intentionally different from the above "should update tagged_at but not edited_at when content edited"?

spec/models/post_spec.rb
```diff
@@ -176,28 +203,6 @@
       create(:reply, post: post, user: post.user)
       expect(post.edited_at).to eq(post.created_at)
     end
-
-    it "should update correctly when characters are edited" do
```
Throne3d: Is this stuff duplicated in `reply_spec.rb`, then?

db/migrate/20180209163253_split_post_text_into_replies.rb
teceler: I was kinda assuming someone was gonna do SQL Magic to it before it was released yeah

spec/models/post_spec.rb
Throne3d: & stuff below

app/views/posts/_write.haml
Throne3d: Can we solve that in the controller, by forcibly setting `post.written` if it's not already set there? That way it avoids some of the other pitfalls that this might here.

db/migrate/20180209163253_split_post_text_into_replies.rb
teceler: otherwise we spawn an audit for every. single. post. for those parts being deleted and moved to the reply. If you wanna re-enable it for prod that's sensible but I got fed up with it

spec/controllers/api/v1/replies_controller_spec.rb
teceler: I have _no idea_

spec/controllers/posts_controller_spec.rb
teceler: I wouldn't expect that to be a thing it could be confused about, with a reply?

db/migrate/20180209163253_split_post_text_into_replies.rb
Throne3d: Huh! I wouldn't have expected audits to act here, since it doesn't have a `current_user`. For prod, I think I'd like it enabled, so we can preserve data alright, but I'm not sure.

spec/controllers/replies_controller_spec.rb
teceler: This is the main user-facing problem with our current model, yeah!

spec/controllers/posts_controller_spec.rb
Throne3d: Probably not! Hmmmmmmmm. Okay I'm fine with this out.

spec/controllers/api/v1/replies_controller_spec.rb
Throne3d: I'm okay with this change, I think. Looks like our current docs don't show the post as being part of the replies endpoint, but it _also_ doesn't show the post content as being part of the _posts_ endpoint, and we need it somewhere.

Should let Vona know though (they maintain glowficrss.com, right?).

spec/models/post_spec.rb
teceler: I think edited_at is still updated when the metadata changes?

spec/models/post_spec.rb
teceler: saving hte post should save the content, but it should _not_ change either of those values. I should probably check it does save the content tho

teceler: apparently it doesn't

teceler: making them the same now xD

spec/models/post_spec.rb
teceler: `#has_icons` was duplicated; we don't have the entire `#edited_at` thing on replies so that isn't duplicated

app/services/post_scraper.rb
teceler: we do apparently

app/controllers/writable_controller.rb
teceler: ...looking at the intervening code it might be so it correctly includes the written in the replies to check for read status. otoh you probably... want pagination there, yes. hm.

app/controllers/writable_controller.rb
```diff
@@ -13,9 +13,7 @@ def build_template_groups(user=nil)
     @templates = templates + [templateless]
 
     if @post
-      written = @reply || @post.written
-      uniq_chars_ids = @post.replies.where(user_id: user.id).where('character_id is not null').group(:character_id).pluck(:character_id)
-      uniq_chars_ids << written.character_id unless written.character_id.nil? or uniq_chars_ids.include?(written.character_id)
+      uniq_chars_ids = @post.replies.where(user: user).select(:character_id).distinct.pluck(:character_id)
```
teceler: Although, hmm, this is probably going to get weird results on post preview

teceler:
To do:
- [ ] work out how to handle audit comments for post and post written (<https://github.com/Marri/glowfic/pull/705#discussion_r247379727>)
- [ ] work out how to properly fix show_post (<https://github.com/Marri/glowfic/pull/705#discussion_r247380201>)
- [ ] make an after_create marking post written read for post user? would avoid needing quite so many mark_reads in tests (<https://github.com/Marri/glowfic/pull/705#discussion_r247381425>)
- [ ] figure out what the _hell_ is going on with the editor buttons and how to fix that (<https://github.com/Marri/glowfic/pull/705#discussion_r306014330>)
- [ ] Test thread characters on post#preview (https://github.com/Marri/glowfic/pull/705#discussion_r359662990)

app/views/posts/_write.haml
Throne3d: Added 57464cf3938162e8b7d20ab5eb9c46103829ba16, which should hopefully fix this!

app/controllers/api/v1/posts_controller.rb
```diff
@@ -22,7 +22,7 @@ def index
   def show
     return unless (post = find_object(Post))
     access_denied and return unless post.visible_to?(current_user)
-    render json: post.as_json(include: [:character, :icon, :content])
+    render json: post.as_json
```
teceler: NTS: double check the post written shows up in replies now

app/controllers/writable_controller.rb
```diff
@@ -13,8 +13,8 @@ def build_template_groups(user=nil)
     @templates = templates + [templateless]
 
     if @post
-      uniq_chars_ids = @post.replies.where(user_id: user.id).where('character_id is not null').group(:character_id).pluck(:character_id)
-      uniq_chars_ids << @post.character_id if @post.user_id == user.id && @post.character_id.present?
+      uniq_chars_ids = @post.replies.where(user: user).select(:character_id).distinct.pluck(:character_id)
+      uniq_chars_ids << @reply.character_id if @reply.present?
```
teceler: Why were we checking user on the local reply before? was it for coauthors or mod edits?

app/models/character.rb
```diff
@@ -62,8 +62,7 @@ def self.deletable_by?(user, user_id)
   def recent_posts
     return @recent unless @recent.nil?
     reply_ids = replies.group(:post_id).pluck(:post_id)
-    post_ids = posts.select(:id).map(&:id)
-    @recent ||= Post.where(id: (post_ids + reply_ids).uniq).ordered
+    @recent ||= Post.where(id: reply_ids.uniq).ordered
```
teceler: This can probably be cleaned up more

db/migrate/20180209163253_split_post_text_into_replies.rb
```diff
@@ -0,0 +1,59 @@
...
+        user_id: post.user_id,
+        character_id: post.character_id,
+        character_alias_id: post.character_alias_id,
+        icon_id: post.icon_id,
+        content: post.content,
+        created_at: post.created_at,
+        updated_at: post.updated_at,
+        skip_post_update: true
+      )
+      reply.update_columns(reply_order: 0, created_at: post.created_at, updated_at: post.edited_at)
```
teceler: This... is going to take a long time. I don't know if there's any way to improve that.

spec/controllers/characters_controller_spec.rb
```diff
@@ -234,7 +234,7 @@
       expect(response.status).to eq(200)
       expect(assigns(:page_title)).to eq(character.name)
       expect(assigns(:posts).size).to eq(25)
-      expect(assigns(:posts)).to match_array(Post.where(character_id: character.id).ordered.limit(25))
+      expect(assigns(:posts)).to match_array(Post.where(id: Reply.where(character_id: character.id).select(:post_id).distinct.pluck(:post_id)).order('tagged_at desc').limit(25))
```
teceler: The order can probably be revered to `.ordered`

spec/controllers/replies_controller_spec.rb
```diff
@@ -654,13 +674,13 @@
...
       reply = create(:reply, post: reply_post)
-      expect(reply.reply_order).to eq(1)
+      order = reply.reply_order
       expect(reply_post.replies.ordered.last).to eq(reply)
       create(:reply, post: reply_post)
       expect(reply_post.replies.ordered.last).not_to eq(reply)
       reply_post.mark_read(reply_post.user)
       put :update, params: { id: reply.id, reply: {content: 'new content'} }
-      expect(reply.reload.reply_order).to eq(1)
+      expect(reply.reload.reply_order).to eq(order)
```
teceler: can probably do this with a not_to change?

Throne3d:
A way we might fragmentize this:

1. Make sure no code relies on `reply_order` being zero-based
2. Create a `new_reply_order` column, and when a Reply is created/updated, set it equal to `reply_order + 1`
3. In batches (by post or otherwise), make sure `new_reply_order` is filled for all replies
4. In a single transaction, do the following renames: `reply_order` -> `old_reply_order`, `new_reply_order` -> `reply_order`, `old_reply_order` -> `new_reply_order`
5. Tear out code that updates `new_reply_order`
6. Delete column `new_reply_order` (actually `old_reply_order`)
7. Make any code that currently treats Posts as a special type of Reply skip any replies with reply_order 0 (temporarily, to prevent duplicate counting). Make code that currently uses just replies use replies skipping order 0 (this is more permanent)
8. Start creating these Reply objects with reply_order 0, on new post creation + backfilling
9. Once they're all here, start reading replies with reply_order 0 instead of / in addition to posts, where relevant
10. Remove old, unused data from posts

Not sure if I got everything here, but I think this is a good chunk of the steps.

app/controllers/characters_controller.rb
redmvc, line 240:
```suggestion

```
This is no longer necessary, right?

app/controllers/icons_controller.rb
redmvc, line 115:
```suggestion

```
Here likewise.

app/controllers/posts_controller.rb
```diff
@@ -545,14 +560,9 @@ def permitted_params(include_associations=true)
...
       :subject,
       :description,
-      :content,
-      :character_id,
-      :icon_id,
-      :character_alias_id,
       :authors_locked,
       :audit_comment,
       :private_note,
-      :editor_mode,
```
redmvc, line 555: If we decided that we're keeping all of the reply params in the post for the moment, why are they removed from permitted_params?

app/controllers/users_controller.rb
```diff
@@ -178,19 +178,12 @@ def output
...
 
-    @posts, @replies = [Post, Reply].map do |klass|
-      klass.where(user: current_user).where('created_at between ? AND ?', daystart, dayend)
-    end
-    @posts = @posts.ordered_by_id.pluck(:content, :editor_mode)
-    @replies = @replies.order(post_id: :asc).ordered.pluck(:content, :editor_mode)
+    @replies = Reply.where(user: current_user)
+      .where('created_at between ? AND ?', daystart, dayend)
+      .order(post_id: :asc).ordered
+      .pluck(:content)
```
redmvc, line 185: Why did we get rid of @posts?

app/controllers/writable_controller.rb
```diff
@@ -58,12 +58,12 @@ def show_post(cur_page=nil)
         self.page = cur_page = 1
       end
     elsif cur_page == 'last'
-      self.page = cur_page = @post.replies.paginate(per_page: per, page: 1).total_pages
+      self.page = cur_page = @post.replies.where.not(reply_order: 0).paginate(per_page: per, page: 1).total_pages
```
redmvc, line 61: This is unnecessary, right? @post.replies already filters this.

app/controllers/writable_controller.rb
redmvc: Dunno, if it passes tests it seems fine by me.

app/controllers/writable_controller.rb
redmvc: Or, did that get undone from the .written code?

app/models/post.rb
```diff
@@ -53,14 +57,14 @@ class Post < ApplicationRecord
...
   NON_TAGGED_ATTRS = %w(icon_id character_alias_id character_id)
+
   audited except: NON_EDITED_ATTRS, update_with_comment_only: false
   has_associated_audits
 
   pg_search_scope(
     :search,
     against: %i(
       subject
-      content
```
redmvc, line 63: If we're not deleting content from posts, should this remain?

app/controllers/posts_controller.rb
teceler: If you're submitting it to the post at this point something has gone badly wrong so it's better for tests to fail?

app/controllers/writable_controller.rb
teceler: Yeah. I'm not sure why this one needs it or whether it *should* but it did get undone from the written code

app/controllers/users_controller.rb
teceler: They'll be duplicates of reply content

app/models/post.rb
teceler: We don't want to search posts at this point for duplication reasons, right?

app/views/posts/flat.haml
```diff
@@ -22,36 +22,4 @@
...
-          .right-align>
-            .padding-5>
-              - if @post.created_at
-                = precede 'Posted ' do
-                  %span.post-posted=pretty_time(@post.created_at, format: ApplicationHelper::TIME_FORMAT)
-              - if @post.created_at.to_i != @post.last_updated.to_i
-                = precede ' | Updated ' do
-                  %span.post-updated=pretty_time(@post.last_updated, format: ApplicationHelper::TIME_FORMAT)
-      .flat-post-replies= @post.flat_post.content.try(:html_safe)
+      = @post.flat_post.content.try(:html_safe)
```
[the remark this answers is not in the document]
teceler: I thiiink we escape this at earlier stages brakeman but thanks?

app/models/post.rb
redmvc: wdym duplication reasons? Currently posts aren't searched at all, they simply do not show up in search

app/controllers/writable_controller.rb
redmvc: This one needs it bc it's paginating, it needs to ignore the top post

redmvc (dismissed):

spec/controllers/api/v1/posts_controller_spec.rb
```diff
@@ -74,12 +74,9 @@
       get :show, params: { id: post.id }
       expect(response).to have_http_status(200)
       expect(response.parsed_body['id']).to eq(post.id)
-      expect(response.parsed_body['num_replies']).to eq(0)
+      expect(response.parsed_body['num_replies']).to eq(1)
```
redmvc, line 77: Why should this be 1? Aren't we excluding `written` from the reply count everywhere else in the code?

spec/models/daily_report_spec.rb
```diff
@@ -71,7 +71,7 @@
         create_list(:reply, 3, post: post, user: post.user)
       end
       report = DailyReport.new(now)
-      expect(report.posts.first.reply_count).to eq(5)
+      expect(report.posts.first.reply_count).to eq(6)
```
redmvc, line 74: Likewise here

spec/models/post_spec.rb
```diff
@@ -1131,7 +1060,7 @@
           created_at: post.created_at,
           tagged_at: post.tagged_at,
           status: :active,
-          num_replies: 0,
+          num_replies: 1,
```
redmvc, line 1063: And here.

spec/controllers/api/v1/posts_controller_spec.rb
teceler: ...i thought we stopped excluding it here, but i might have updated some tests incorrectly

redmvc: Huh actually you're right. It's only in application_controller's calculate_view_status that we're still excluding it, plus places that paginate, everywhere else we did stop excluding it. So should calculate_view_status instead not exclude it?

redmvc: Yeah I removed that exclusion from there, everything's ok now.

redmvc (approved):

app/models/post.rb
```diff
@@ -89,7 +93,7 @@ class Post < ApplicationRecord
   # rubocop:enable Style/TrailingCommaInArguments
 
   scope :with_reply_count, -> {
-    select('(SELECT COUNT(*) FROM replies WHERE replies.post_id = posts.id) AS reply_count')
+    select('(SELECT COUNT(*) FROM replies WHERE replies.post_id = posts.id AND replies.reply_order > 0) AS reply_count')
```
redmvc: @teceler seems like here we're filtering the written out of reply_count

Issue #708, opened 2018-02-16
Display authors in characters#search


Throne3d:
[06:51] Ezra: Hey I thought I remembered this suggestion being suggested but it didn't turn up in a GitHub search, so
[06:52] Ezra: When searching for characters, in the results table it could be good to show the name of the user whose character it is
[06:53] Ezra: E.g. in https://www.glowfic.com/characters/search?utf8=✓&author_id=&template_id=&name=Junia&search_name=true&search_screenname=true&search_nickname=true&commit=Search it could say that the cheerfulish one belongs to Kappa and the altus one belongs to Anya.

Issue #696, opened 2018-01-30
Template#new characters need linebreaks between them


Throne3d:
For some reason the checkboxes currently lack linebreaks? This is weird and is probably an artifact of when I modified the system to use `collection_check_boxes`.

PR #710, opened 2018-02-20
Various small things

Marri:
- Add Nemo to scraper authors
- Automatically strip whitespace from the beginning/end of usernames to prevent login issues
- Don't tag a post when updating character/icon/alias, especially because character replace now uses callbacks
- Add line breaks to the untemplated character list on template editor
- Add template descriptions in relevant places
- Display author of a character in the search page
- Fix misbehaving Authors Locked checkbox on post editor


Marri:
Fixes #708 and #696

PR #726, opened 2018-03-01
Dev/has dependent

teceler:
* Fixes some things not having `dependent: destroy` that probably should
* Fixes character groups not having `dependent: nullify` when they probably should
* Clarifies some dependencies being cleaned up in callbacks in comments (and gets rubocop to shush about them)

### DIFF
--- a/app/models/board.rb
+++ b/app/models/board.rb
@@ -9,7 +9,7 @@ class Board < ApplicationRecord
   has_many :favorites, as: :favorite, dependent: :destroy
   belongs_to :creator, class_name: 'User', optional: false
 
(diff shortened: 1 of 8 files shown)


Throne3d:
Again, looks straightforward. Things that changed (i.e. things that are not `dependent: false`):

- boards destroy board_authors on destruction
- characters destroy characters_galleries on destruction
- galleries destroy characters_galleries on destruction
- posts destroy flat_posts on destruction

I'm gonna check that characters_galleries have reasonable callbacks on destruction, in case these will make it duplicate a response.

PR #734, opened 2018-03-01
Use bang methods for create/update/destroy

teceler:
* Changes `.save` to `.save!`
* Changes `.destroy` to `.destroy!`
* Changes `.update_attributes` to `.update_attributes!`
* Changes `.create` to `.create!`
* Handles some tests


Marri:
I strongly prefer that we add if <save> else <fail message> for this kind of thing; using bang methods means that if it fails the user will see an Unhandled Error page rather than a useful error message, which isn't ideal. We should always prefer handling errors where we can.

spec/controllers/favorites_controller_spec.rb
```diff
@@ -237,5 +237,15 @@
...
     end
+
+    it "handles destroy failure" do
+      favorite = create(:favorite, favorite: create(:post))
+      login_as(favorite.user)
+      expect_any_instance_of(Favorite).to receive(:destroy!).and_raise(ActiveRecord::RecordNotDestroyed, 'fake error')
+      delete :destroy, params: { id: favorite.id }
+      expect(response).to redirect_to(favorites_path)
+      expect(flash[:error]).to eq({message: "Favorite could not be deleted.", array: []})
+      expect(Favorite.all.find_by(id: favorite.id)).not_to be_nil
```
Throne3d: `Favorite.find_by` instead?

spec/controllers/galleries_controller_spec.rb
```diff
@@ -413,6 +413,17 @@
...
+
+    it "handles destroy failure" do
+      gallery = create(:gallery)
+      icon = create(:icon, user: gallery.user, galleries: [gallery])
+      login_as(gallery.user)
+      expect_any_instance_of(Gallery).to receive(:destroy!).and_raise(ActiveRecord::RecordNotDestroyed, 'fake error')
+      delete :destroy, params: { id: gallery.id }
+      expect(response).to redirect_to(gallery_url(gallery))
+      expect(flash[:error]).to eq({message: "Gallery could not be deleted.", array: []})
+      expect(icon.reload.galleries.first).to eq(gallery)
```
Throne3d: `expect(icon.reload.galleries).to match_array([gallery])`? Would be more precise (though I guess not testing much differently)

spec/controllers/tags_controller_spec.rb
```diff
@@ -256,5 +256,15 @@ def create_tags
...
     end
+
+    it "handles destroy failure" do
+      tag = create(:label)
+      login_as(tag.user)
+      expect_any_instance_of(Tag).to receive(:destroy!).and_raise(ActiveRecord::RecordNotDestroyed, 'fake error')
+      delete :destroy, params: { id: tag.id }
+      expect(response).to redirect_to(tag_url(tag))
+      expect(flash[:error]).to eq({message: "Tag could not be deleted.", array: []})
+      expect(Tag.all.find_by(id: tag.id)).not_to be_nil
```
Throne3d: `Tag.find_by`

app/controllers/api/v1/characters_controller.rb
```diff
@@ -53,7 +53,7 @@ def update
     errors += @character.errors.full_messages.map { |msg| {message: msg} } unless @character.valid?
     render json: {errors: errors}, status: :unprocessable_entity and return unless errors.empty?
 
-    @character.save
+    @character.save!
     render json: @character.as_json(include: [:default])
```
Throne3d: It's probably okay, at least for now, for the API to throw an error.

app/controllers/messages_controller.rb
```diff
@@ -33,7 +33,10 @@ def create
       render action: :preview and return
     end
 
-    unless @message.valid? && flash.now[:error].nil?
+    if @message.valid? && flash.now[:error].nil? && @message.save
```
Throne3d: Should be able to condense this into `if flash.now[:error].nil? && @message.save`

app/controllers/password_resets_controller.rb
```diff
@@ -53,17 +53,21 @@ def update
...
+        @password_reset.user.save!
+        @password_reset.update!(used: true)
+      end
+      flash[:success] = "Password successfully changed."
+      redirect_to root_url
+    rescue ActiveRecord::RecordInvalid
       flash.now[:error] = {}
       flash.now[:error][:message] = "Could not update password."
       flash.now[:error][:array] = @password_reset.user.errors.full_messages
+      flash.now[:error][:array] = @password_reset.errors.full_messages
```
Throne3d: You want a `+=` here, not `=` (otherwise there's a useless assignment on the line above)

app/controllers/posts_controller.rb
```diff
@@ -243,16 +243,22 @@ def change_status
       flash[:error] = "Invalid status selected."
     else
       @post.status = new_status
-      @post.save
-      flash[:success] = "Post has been marked #{params[:status]}."
+      if @post.save
+        flash[:success] = "Post has been marked #{params[:status]}."
+      else
+        flash[:error] = "Status could not be updated."
```
Throne3d: Maybe show the post's full list of errors too? It might give a better idea of what the exact error is.

app/controllers/posts_controller.rb
```diff
@@ -243,16 +243,22 @@ def change_status
...
   end
 
   def change_authors_locked
     @post.authors_locked = (params[:authors_locked] == 'true')
-    @post.save
-    flash[:success] = "Post has been #{@post.authors_locked? ? 'locked to' : 'unlocked from'} current authors."
+    if @post.save
+      flash[:success] = "Post has been #{@post.authors_locked? ? 'locked to' : 'unlocked from'} current authors."
+    else
+      flash[:error] = "Post could not be updated."
```
Throne3d: Ditto

app/controllers/replies_controller.rb
```diff
@@ -210,9 +210,18 @@ def destroy
...
     previous_reply = @reply.send(:previous_reply)
     to_page = previous_reply.try(:post_page, per_page) || 1
-    @reply.destroy # to destroy subsequent ones, do @reply.destroy_subsequent_replies
-    flash[:success] = "Post deleted."
-    redirect_to post_path(@reply.post, page: to_page)
+
+    # to destroy subsequent replies, do @reply.destroy_subsequent_replies
+    begin
+      @reply.destroy!
+      flash[:success] = "Reply deleted."
```
Throne3d: Oh, neat, you fixed "Post deleted" being wrong. <3

app/models/reply.rb
```diff
@@ -113,7 +113,7 @@ def update_post_authors
     if post_author
       post_author.update_attributes(joined: true, joined_at: created_at)
```
Throne3d: Don't we want a bang here too?

script/character_scraper.rb
```diff
@@ -34,7 +34,7 @@
   end
 
   unless CharactersGallery.where(gallery_id: gallery.id, character_id: character.id).exists?
-    CharactersGallery.create(character_id: character.id, gallery_id: gallery.id)
+    CharactersGallery.create!(character_id: character.id, gallery_id: gallery.id)
   end
```
Throne3d: Think this thing can actually condense into:
```ruby
CharactersGallery.find_or_create_by!(character_id: character.id, gallery_id: gallery.id)
```

(But it's very old code, so it doesn't much matter.)

spec/controllers/api/v1/tags_controller_spec.rb
```diff
@@ -99,7 +99,7 @@
 
     [:setting, :label, :content_warning].each do |type|
       it "should support getting #{type} tags" do
-        tag = create(type)
+        tag = create(type) # rubocop:disable Rails/SaveBang
```
Throne3d: Vaguely don't like the look of this inline :P

It's annoying it detects this incorrectly, and I think it's a bug.

app/controllers/replies_controller.rb
teceler: I was going through them and it annoyed me!

spec/controllers/api/v1/tags_controller_spec.rb
teceler: I tried writing it as `FactoryBot.create(type)` but that didn't fix it

Throne3d: Yeah, I'm mostly just surprised that it recognizes it without a model in front of it. I'm okay leaving it as is for now.

PR #735, opened 2018-03-01
[WIP] Avoid skipping model validations

teceler:
- [x] Use `update` (or `update_attributes`, they're the same) rather than `update_attribute`
- [ ] Do something about `update_all`?
- [ ] Do something about `touch`? (only used in tests, may not be important)


Throne3d:
(Marri says some of these (most of these?) are intentional; needs checking over more thoroughly.)

PR #736, opened 2018-03-01
Use new-style validation

teceler:


app/models/user.rb
```diff
@@ -25,15 +25,15 @@ class User < ApplicationRecord
...
-  validates_presence_of :password, :password_confirmation, if: :validate_password?
+  validates :username, :crypted, presence: true
+  validates :email, presence: { on: :create }
+  validates :username, uniqueness: true
+  validates :email, uniqueness: { allow_blank: true }
+  validates :username, length: { in: MIN_USERNAME_LEN..MAX_USERNAME_LEN, allow_blank: true }
+  validates :moiety, length: { in: 3..6, allow_blank: true }
+  validates :password, length: { minimum: 6, if: :validate_password? }
+  validates :password, confirmation: { if: :validate_password? }
+  validates :password, :password_confirmation, presence: { if: :validate_password? }
```
Throne3d: Could we combine duplicate validations here? I'm pretty sure you're able to do e.g.:

```ruby
validates :password, length: { minimum: 6, if: :validate_password? }
validates :password, confirmation: { if: :validate_password? }
# turns into:
validates :password, length: { minimum: 6, if: :validate_password? }, confirmation: { if: :validate_password? }
```

teceler: oh, yup, I missed that one when I was tidying.

PR #757, opened 2018-03-10
Don't crash if the session[:user_id] is invalid

### LINKED ISSUE #752: Don't crash if client is logged in as nonexistent user
Particularly useful on staging, where we might be deleting users and yet their clients have cookies as if they're still logged in.

See 047b02bb488d44f7ffda57e09d6885daac2796ea.

Marri:
Mostly useful for staging environment, if you're logged in but we reset the data. Fixes #752

### DIFF
--- a/app/concerns/authentication.rb
+++ b/app/concerns/authentication.rb
@@ -15,7 +15,14 @@ module Authentication
     helper_method :logged_in?
 
     def current_user
-      @current_user ||= User.find_by_id(session[:user_id]) if logged_in?
+      return unless logged_in?
+      @current_user ||= User.find_by_id(session[:user_id])
+      return @current_user if @current_user
+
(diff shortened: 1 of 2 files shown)


app/concerns/authentication.rb
```diff
@@ -15,7 +15,14 @@ def logged_in?
...
 
     def current_user
-      @current_user ||= User.find_by_id(session[:user_id]) if logged_in?
+      return unless logged_in?
+      @current_user ||= User.find_by_id(session[:user_id])
+      return @current_user if @current_user
+
+      # logout - something has gone wrong, and the user id stored in session does not exist
+      reset_session
+      cookies.delete(:user_id, domain: '.glowfic.com')
```
Throne3d: Should we extract this log-out code into some separate method, in case we change it in the future? (It's also used over in sessions#destroy)

PR #780, opened 2018-03-27
Avoid attempting to include coauthor information on solo threads

### LINKED ISSUE #745: Fix new post notification with just one post author
<https://github.com/Marri/glowfic/blob/4858c6c6bafd7a64539ef81615dff0c5838f5f21/app/jobs/notify_followers_of_new_post_job.rb#L27> needs a guard around it to make it not add "with" when there's only the one author.

teceler:
Fixes #745


Throne3d:
I've also added a commit which renames `new_user` to `post_user` (to make it clearer that they're the same, in this case) and which consistently makes it use `post_user` instead of `post.user` (to prevent potential additional queries); this also converts some instances of string concatenation into interpolation. Another commit adds a test that this functions as expected.

PR #815, opened 2018-05-19
Move /users/username to the API

### LINKED ISSUE #805: Move /users/username to the API
- [ ] Since it returns JSON, and is only used in our JavaScript code, the /users/username route should probably be put into our API.
- [ ] In addition, since it's non-destructive and simply checks if a username is available, it should be a `GET` request instead of a `POST` request.

Marri:
Fixes #805

### DIFF
--- a/app/controllers/api/v1/users_controller.rb
+++ b/app/controllers/api/v1/users_controller.rb
@@ -5,10 +5,15 @@ class Api::V1::UsersController < Api::ApiController
 
   api! 'Load all the users that match the given query, results ordered by username'
   param :q, String, required: false, desc: "Query string"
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/users_controller.rb
```diff
@@ -5,10 +5,15 @@ class Api::V1::UsersController < Api::ApiController
 
   api! 'Load all the users that match the given query, results ordered by username'
   param :q, String, required: false, desc: "Query string"
+  param :match, String, required: false, desec: "If set to 'exact', requires exact username match on q instead of prefix match"
```
Throne3d: desc*

PR #825, opened 2018-06-10
Override layout-th styling in post-content tables

Throne3d:
We use `th` semantically all over the place for headers, and then style
them a particular way for consistent formatting. Unfortunately, this
messes with `th` being used in replies, since it messes up the
formatting. Unset the changes we make (as a temporary hack) so they use
a more default formatting in posts.


Throne3d:
See #826 as a follow-up to clean up the CSS later.

Issue #826, opened 2018-06-10
Improve semantic-th selector


Throne3d:
Here, we use just `th` to select all headers on the page:
https://github.com/Marri/glowfic/blob/4612a0da3e607ebe67bed8ee6af0faaa8e45c908/app/assets/stylesheets/layout.scss#L142
Since we use `th` all over the place in this exact way, this is normally fine, but it does break stuff around `.post-content`s having tables. Since we _don't_ in fact want all `th` styled this way – and we override it with `.sub` and whatnot in all sorts of places – could we make only `th` that _should_ be styled this way have a class, or otherwise mutate this selector, so we only catch the ones we want to?

Follow-up to #825, which overrides the styling just for `post-content`s.

teceler:
Resolved in #837

PR #837, opened 2018-07-03
Fix styling semantic th

teceler:
Addresses #826

- [x] Add new selector to all uses of `th` that do not have another selector overriding the css on th
- [x] Modify selectors to catch common combinations with other semantic selectors
- [x] Rename new selectors to something more descriptive
  - [x] Rename `.old_th`
  - [x] Rename `.old_th_centered`
  - [x] Rename `.old_th_width-150`
  - [x] Rename `.old_th_padding-10`
- [ ] Test with css tests

### DIFF
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -223,6 +223,7 @@ td, .table-list li { font-size: $font_size_small; }
   background-color: $bg_color_sub;
   color: $font_color_sub;
   font-size: $font_size_main;
+  font-weight: normal;
   &.width-150 { width: 170px; }
 }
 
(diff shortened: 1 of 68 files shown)


Throne3d:
I seem to recall going over this with you and being happy about it at the time – looks like it's accumulated a bunch of merge conflicts in the meantime, though.

If we could get this up to date, and do some quick wraith tests on it, I'd be happy to merge it!

Throne3d:
Have you tested this with Wraith? I no longer remember having gone over this with you, but it looks good from a quick skim, and it seems non-harmful if we do get something wrong (easy to fix). If you're confident merging this (after fixing the, like, two conflicts that seem to have appeared), go ahead!

teceler:
Pretty sure the last few commits are the result of going over it with wraith, yeah

Issue #694, opened 2018-01-29
Add OpenGraph data to more objects


Throne3d:
Boards. Characters? Templates? Settings? Galleries? Etc. See #641.

teceler:
Fixed with #856

PR #856, opened 2018-07-24
Add more open graph data

teceler:
See #694

### DIFF
--- a/app/controllers/board_sections_controller.rb
+++ b/app/controllers/board_sections_controller.rb
@@ -34,6 +34,7 @@ class BoardSectionsController < ApplicationController
   def show
     @page_title = @board_section.name
     @posts = posts_from_relation(@board_section.posts.ordered_in_section)
+    @meta_og = og_data
   end
 
   def edit
@@ -95,6 +96,20 @@ class BoardSectionsController < ApplicationController
       flash[:error] = "You do not have permission to edit this continuity."
       redirect_to boards_path and return
     end
+  end
+
+  def og_data
+    stats = []
+    stats << @board_section.board.writers.reject(&:deleted?).pluck(:username).sort_by(&:downcase).join(', ') unless @board_section.board.open_to_anyone?
+    post_count = @board_section.posts.count
+    stats << "#{post_count} " + "post".pluralize(post_count)
+    desc = [stats.join(' – ')]
+    desc << generate_short(@board_section.description) if @board_section.description.present?
+    {
+      url: board_section_url(@board_section),
+      title: "#{@board_section.board.name} » #{@board_section.name}",
+      description: desc.join("\n"),
+    }
   end
 
   def section_params
--- a/app/controllers/boards_controller.rb
+++ b/app/controllers/boards_controller.rb
@@ -63,6 +63,7 @@ class BoardsController < ApplicationController
       board_posts = board_posts.ordered
     end
     @posts = posts_from_relation(board_posts, no_tests: false)
+    @meta_og = og_data
     use_javascript('boards/show')
   end
 
@@ -155,6 +156,23 @@ class BoardsController < ApplicationController
       flash[:error] = "You do not have permission to edit that continuity."
       redirect_to board_path(@board) and return
     end
+  end
+
+  def og_data
+    metadata = []
+    metadata << @board.writers.reject(&:deleted?).pluck(:username).sort_by(&:downcase).join(', ') unless @board.open_to_anyone?
+    post_count = @board.posts.count
+    stats = "#{post_count} " + "post".pluralize(post_count)
+    section_count = @board.board_sections.count
+    stats += " in #{section_count} " + "section".pluralize(section_count) if section_count > 0
+    metadata << stats
+    desc = [metadata.join(' – ')]
+    desc << generate_short(@board.description) if @board.description.present?
+    {
+      url: board_url(@board),
+      title: @board.name,
+      description: desc.join("\n"),
+    }
   end
 
   def board_params
--- a/app/controllers/galleries_controller.rb
+++ b/app/controllers/galleries_controller.rb
@@ -86,6 +86,7 @@ class GalleriesController < UploadingController
 
       @user = @gallery.user
       @page_title = @gallery.name + ' (Gallery)'
+      @meta_og = og_data
     end
     icons = @gallery ? @gallery.icons : @user.galleryless_icons
     render :show, locals: { icons: icons }
@@ -231,6 +232,21 @@ class GalleriesController < UploadingController
   def setup_editor
     use_javascript('galleries/editor')
     gon.user_id = current_user.id
+  end
+
+  def og_data
+    icon_count = @gallery.icons.count
+    desc = ["#{icon_count} " + "icon".pluralize(icon_count)]
+    tags = @gallery.gallery_groups_data.pluck(:name)
+    tag_count = tags.count
+    desc << "Tag".pluralize(tag_count) + ": " + generate_short(tags.join(', ')) if tag_count > 0
+    title = [@gallery.name]
+    title.prepend(@gallery.user.username) unless @gallery.user.deleted?
+    {
+      url: gallery_url(@gallery),
+      title: title.join(' » '),
+      description: desc.join("\n"),
+    }
   end
 
   def gallery_params
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -20,6 +20,7 @@ class TagsController < ApplicationController
   def show
     @page_title = @tag.name.to_s
     @view = params[:view]
+    @meta_og = og_data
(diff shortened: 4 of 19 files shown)


app/controllers/boards_controller.rb
```diff
@@ -148,6 +149,22 @@ def require_permission
...
+      title: @board.name,
+    }
+    desc = []
+    desc << @board.writers.pluck(:username).sort_by(&:downcase).join(', ') unless @board.open_to_anyone?
+    post_count = @board.posts.count
+    stats = "#{post_count} " + "post".pluralize(post_count)
+    section_count = @board.board_sections.count
+    stats += " in #{section_count} " + "section".pluralize(section_count) if section_count > 0
+    desc << stats
+    data[:description] = desc.join(' – ')
```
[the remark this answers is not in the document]
teceler: I forgot boards could have descriptions! will do

app/controllers/galleries_controller.rb
```diff
@@ -228,6 +229,16 @@ def setup_editor
     gon.user_id = current_user.id
   end
 
+  def og_data
+    data = {
+      url: gallery_url(@gallery),
+      title: "#{@gallery.user.username} » #{@gallery.name}",
+    }
+    icon_count = @gallery.icons.count
+    data[:description] = "#{icon_count} " + "icon".pluralize(icon_count)
```
Throne3d: Would it be useful to include (some) gallery tags here?

app/controllers/tags_controller.rb
```diff
@@ -104,6 +105,25 @@ def build_editor
...
+    desc << generate_short(@tag.description) if @tag.description.present?
+    stats = []
+    post_count = @tag.posts.count
+    stats << "#{post_count} " + "post".pluralize(post_count) if post_count > 0
+    gallery_count = @tag.galleries.count
+    stats << "#{gallery_count} " + "gallery".pluralize(gallery_count) if gallery_count > 0
+    character_count = @tag.characters.count
+    stats << "#{character_count} " + "character".pluralize(character_count) if character_count > 0
+    desc << stats.join(', ')
+    data[:description] = desc.join("\n")
```
Throne3d: If the tag is a setting, perhaps mention who it's owned by (if it's, in fact, owned)?

Throne3d:
A more general note: I think it would be worthwhile declaring the `data` hash at the _end_ of the `meta_og` methods – like we do in the writable controller:

https://github.com/Marri/glowfic/blob/master/app/controllers/writable_controller.rb#L156-L159

This way, it's obvious at a glance which keys are set and which variables are used for them, in a way that can be traced upwards, instead of needing to look at where the various things are overridden.

app/controllers/board_sections_controller.rb
```diff
@@ -86,6 +87,19 @@ def require_permission
     end
   end
 
+  def og_data
+    data = {
+      url: board_section_url(@board_section),
+      title: "#{@board_section.board.name} » #{@board_section.name}",
```
Throne3d: To keep consistent with the style used for posts, should this instead be in the format `Board section · Board` (instead of `Board » Board section`)? Same goes for some of the other things here. It might be worth asking #constellation on it.

app/controllers/users_controller.rb
```diff
@@ -107,6 +108,29 @@ def signup_prep
...
+        height: '75',
+      }
+    end
+
+    board_ids = BoardAuthor.where(user_id: @user.id, cameo: false).pluck('distinct board_id')
+    arel = Board.arel_table
+    where = arel[:creator_id].eq(@user.id).or(arel[:id].in(board_ids))
+    boards = Board.where(where).ordered
+
+    data[:description] = generate_short(boards.pluck(:name) * ', ')
```
Throne3d: Perhaps we should say "Continuities:" before this? I'm not sure – it might be clear.

spec/controllers/board_sections_controller_spec.rb
```diff
@@ -120,6 +120,19 @@
...
+      user = create(:user, username: 'John Doe')
+      board = create(:board, name: 'board', creator: user, coauthors: [create(:user, username: 'Jane Doe')])
+      section = create(:board_section, name: 'section', board: board)
+      create(:post, subject: 'title', user: user, board: board, section: section)
+      get :show, params: { id: section.id }
+
+      meta_og = assigns(:meta_og)
+      expect(meta_og[:url]).to eq(board_section_url(section))
+      expect(meta_og[:title]).to eq('board » section')
+      expect(meta_og[:description]).to eq('Jane Doe, John Doe – 1 post')
```
Throne3d: Perhaps this test (and others) should confirm that `meta_og` doesn't have other keys set? i.e. `expect(meta_og.keys).to match_array([:url, :title, :description])`?

app/controllers/board_sections_controller.rb
teceler: Yeah I wanted to consult #constellation on a lot of the formatting tbh, and I figured it'd be easier to do with samples

spec/controllers/users_controller_spec.rb
```diff
@@ -167,6 +167,23 @@
...
+      get :show, params: { id: user.id }
+
+      meta_og = assigns(:meta_og)
+      expect(meta_og[:url]).to eq(user_url(user))
+      expect(meta_og[:title]).to eq('user')
+      expect(meta_og[:description]).to eq('Board 1, Board 2')
+      expect(meta_og[:image][:src]).to eq(user.avatar.url)
+      expect(meta_og[:image][:width]).to eq('75')
+      expect(meta_og[:image][:width]).to eq('75')
+    end
```
Throne3d: Test this works for a user without an avatar, too?

app/controllers/tags_controller.rb
teceler: Do you have a suggestion on where to put that?

Throne3d: Hmm. I would _probably_ choose to put it between the tag name and the tag type (so, like, `Proximal · Throne3d · Setting`), to match the fact we have a similar thing on posts (`Between and beyond stations · Kel, Throne3d`), but I'm not sure if that's the best place. Maybe put it there and then we ask for feedback on it later?

app/controllers/board_sections_controller.rb
Throne3d: That's totally fair.

spec/controllers/galleries_controller_spec.rb
```diff
@@ -210,6 +210,23 @@
...
+          gallery_groups: [create(:gallery_group), create(:gallery_group)],
+        )
+        create_list(:icon, 16, gallery_ids: [gallery.id])
+        get :show, params: { id: gallery.id }
+
+        meta_og = assigns(:meta_og)
+        expect(meta_og.keys).to match_array([:url, :title, :description])
+        expect(meta_og[:url]).to eq(gallery_url(gallery))
+        expect(meta_og[:title]).to eq("user » gallery")
+        expect(meta_og[:description]).to eq("16 icons\nTags: Tag1, Tag2")
```
Throne3d: This needed names set on the `create(:gallery_group)`s above, iirc (will check, am reading through commits)

Throne3d:
Alright, I'm pushing this to staging so we can get feedback! Thanks for all the hard work 😃

Throne3d:
This is currently blocked on community feedback! I forget what the result was, but I don't remember people having strong opinions. Next step is to check logs to make sure this is right, and then to either make small amendments or get it merged.

Throne3d:
> [3:34 PM] Kappa✶: a note: it was hard to tell the difference between different types of links (gallery, template, character) without looking at the url
> [3:35 PM] Kappa✶: like i remember looking at the gallery preview example and going "that's a slightly oddly formatted charact...oh"
> [3:36 PM] Throne✶: (ideas: "another '· Type' thing in the title? like 'Marri » Alli · Character'?", "a prefix instead? 'Character: Marri » Alli'?", "the first line of the description says what thing it is?")
> [3:39 PM] Unbitwise: how about Marri » Character: Alli

app/controllers/board_sections_controller.rb
```diff
@@ -86,6 +87,18 @@ def require_permission
     end
   end
 
+  def og_data
+    desc = []
+    desc << @board_section.board.writers.pluck(:username).sort_by(&:downcase).join(', ') unless @board_section.board.open_to_anyone?
+    post_count = @board_section.posts.count
+    desc << "#{post_count} " + "post".pluralize(post_count)
```
teceler: Add code for descriptions if present, since there's a migration in for that now.

Throne3d:
I think this needs to handle deleted users before being merged? After that, seems good.

Throne3d:
Apparently needs rebase & squash! Will remove ready to review tag pending that, but, like, it's ready to merge as soon as that's done basically.

PR #860, opened 2018-08-01
Fix "next/previous post" buttons around visibility statuses

teceler:


app/models/post.rb
```diff
@@ -256,12 +256,12 @@ def user_joined(user)
   end
 
   def prev_post
-    return nil if self.board.open_to_anyone? || self.section_id.nil?
+    return nil if self.board.open_to_anyone? && self.section_id.nil?
```
Throne3d: The `nil` doesn't do anything here – `return if` should work fine.

spec/controllers/posts_controller_spec.rb
```diff
@@ -1052,6 +1052,85 @@
...
+      board = create(:board, creator: user)
+      section = create(:board_section, board: board)
+      extra = create(:post, user: user, board: board, section: section)
+      prev = create(:post, user: user, board: board, section: section)
+      hidden = create(:post, board: board, section: section, privacy: Concealable::PRIVATE)
+      post = create(:post, user: user, board: board, section: section)
+      extra.update!(section_order: 0)
+      prev.update!(section_order: 1)
+      hidden.update!(section_order: 2)
+      post.update!(section_order: 3)
```
Throne3d: Shouldn't the posts be in this order anyway? Maybe just check `expect([extra, prev, hidden, post].map(&:section_order)).to eq([0, 1, 2, 3])`

spec/controllers/posts_controller_spec.rb
```diff
@@ -1052,6 +1052,85 @@
...
+
+      get :show, params: { id: post.id }
+
+      expect(assigns(:prev_post)).to eq(prev)
```
Throne3d: Maybe check `next_post` is nil, just to be more exhaustive?

spec/controllers/posts_controller_spec.rb
```diff
@@ -1052,6 +1052,85 @@
...
+      nextp = create(:post, user: user, board: board, section: section)
+      extra = create(:post, user: user, board: board, section: section)
+      post.update!(section_order: 0)
+      hidden.update!(section_order: 1)
+      nextp.update!(section_order: 2)
+      extra.update!(section_order: 3)
+
+      get :show, params: { id: post.id }
+
+      expect(assigns(:next_post)).to eq(nextp)
```
Throne3d: Ditto maybe check `prev_post` is nil

spec/controllers/posts_controller_spec.rb
```diff
@@ -1052,6 +1052,85 @@
...
+      hidden = create(:post, board: board, section: section, privacy: Concealable::PRIVATE)
+      post.update!(section_order: 0)
+      hidden.update!(section_order: 1)
+
+      get :show, params: { id: post.id }
+
+      expect(assigns(:next_post)).to be_nil
+    end
+
+    it "handles very large mostly-hidden sections correctly" do
```
Throne3d: Maybe "as expected" rather than "correctly" (since in a perfect world this would handle arbitrary numbers of hidden posts).

spec/controllers/posts_controller_spec.rb
```diff
@@ -1052,6 +1052,85 @@
       end
     end
 
+    it "gives the correct previous post" do
```
Throne3d: *with an intermediate private post

Throne3d: Do we test it without an intermediate private post, actually? I can't see a test for that – it might be worth adding one.

PR #861, opened 2018-08-01
Set up server with docker-compose

MichaelBlume:


MichaelBlume:
Running

```docker-compose up
docker-compose exec web rake db:create
docker-compose exec web rake db:migrate
docker-compose exec web rake db:seed
docker-compose exec web script/rails s```

Should get you a running glowfic server
`docker-compose exec web rake spec` will run tests

I had to make the app look for the postgres server on `postgres` rather than `localhost`, which is probably messing up travis, I'll fix this.

File: Dockerfile [hunk 1]
```diff
@@ -0,0 +1,13 @@
...
+
+ADD . /code
+
+WORKDIR /code
+
+RUN apt-get update
+RUN apt-get install -y nodejs
+
+RUN gem install bundler
+RUN bundler
```
Marri: I typically run `bundle install`; while they do the same thing, `bundler` is a wrapper around `bundle` and it might be nice to be explicit with the install?

config/database.yml
```diff
@@ -1,6 +1,8 @@
 default: &default
   adapter: postgresql
-  host: localhost
+  host: <%= ENV['GLOWFIC_DATABASE_HOST'] || 'localhost' %>
+  username: <%= ENV['GLOWFIC_DATABASE_USER'] || 'root' %>
```
Marri: I think dev servers and Travis disagree unfortunately: https://docs.travis-ci.com/user/database-setup/#postgresql might just need to use the config files that suggests

MichaelBlume: I was actually able to convince it to use environment variables so that travis would see one config and docker would see another one. Also shouldn't break anyone's existing dev environment.

Dockerfile
MichaelBlume: Done =)

Marri:
I did see the environment variables! But mentioned it cause Travis didn't
use root as a default (which I see you have since fixed) and cause I do
actually expect this to break my dev environment (I'm not sure I've set up
any users locally, which we could explicitly disallow as an anti-pattern
but then it should at least be mentioned explicitly and also probably in
the README?)

MichaelBlume:
I -- *think* I've set things up so if the environment variables aren't set no settings are changed, but I guess someone using the existing workflow would need to check out the branch and try it out to be sure.

config/database.yml
```diff
@@ -1,6 +1,8 @@
 default: &default
   adapter: postgresql
-  host: localhost
+  host: <%= ENV['GLOWFIC_DATABASE_HOST'] || 'localhost' %>
+  username: <%= ENV['GLOWFIC_DATABASE_USER'] || 'postgres' %>
```
Throne3d: I _think_ it would work better to default this to `nil` if the ENV isn't set (ie remove the or condition), which will hopefully make it act like the line isn't set. My understanding is that, by default, Rails uses `throne3d` as the database user on my developer box, because I'm logged into a UNIX account named `throne3d` (and then I have to create a database user named `throne3d` and so on).

MichaelBlume: Done =)

docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+  web:
+    build: .
+    ports:
+      - "3000:3000"
+    volumes:
+      - .:/code
+    links:
+      - redis
+      - postgres
+    command: tail -f /dev/null
```
Throne3d: What's the reason to `tail -f /dev/null` here? My guess would be that it's just to keep the container alive, so that you can perform other tasks (like actually running the web server) in different consoles, but couldn't we just have this run the web server anyway, and instead we use `docker-compose run` in the `bin-docker/` files? (Inspiration for this comes from here: <https://docs.docker.com/compose/rails/#build-the-project>.)

This may require that we migrate the database first, but I think that should be as simple as adding a line, `RUN rails db:create db:migrate` (or `RUN db:setup`) in the `Dockerfile`.

Throne3d: The reason I propose this is so that it should be as simple as `docker-compose up` to get everything running and have a live dev environment.

MichaelBlume: So the tricky thing about `RUN rails db:setup` in the Dockerfile is that it requires the postgres service to be present, so it can't just be part of the setup for the web image, it has to be part of setup for the whole cluster

Throne3d: I've looked around a bit, and can't actually find how to get around this issue, which is annoying. I would expect this to be a problem other people had encountered, but it's not obvious from the keywords I searched with. I tried just using `bundler exec rails db:migrate && bundler exec rails s`, but wasn't able to properly test it. (I'm encountering issues where the `/code/` directory of the container is empty, but this only happened after I started making changes. I'm not sure if there's some kind of caching issue going on with how docker's creating the images, or if it's something to do with me using Windows Subsystem for Linux in the same directory, but reverting to your version of the branch doesn't seem to fix it to how it was before — even when I use `docker-compose build` or `docker-compose up --build`.)

Throne3d:
This no longer seems to interfere with my development environment (with some cursory testing), which is good. If other people can get this to work – my device is having trouble with it, as described in https://github.com/Marri/glowfic/pull/861#discussion_r207697701 – then it seems worth merging.

MichaelBlume:
@Throne3d the web image doesn't have all the code, I actually set that deliberately so that when the code changed the image would not be rebuilt. The web image only hase the Gemfile which indicates which ruby gems need to be installed. When the web container is run it mounts the code from the user's filesystem. I would've assumed that'd happen before the command runs though.

MichaelBlume:
(A previous iteration of this PR copied all the code into the image)

Throne3d:
Yeah – I did in fact try using `COPY` to get all the files over, in case that made a difference – but the command complains of being unable to find `Gemfile` or a .bundle folder, and when I instead tell it to execute `ls` (or give me a bash prompt so I can do the same), the /code/ directory seems to be completely empty. I'm honestly not sure what's up with it, and I was thinking about returning to it later in case I'm just doing something stupid. (Using `docker exec [etc]` also produces the error about Gemfile / .bundle, whereas I recall previously getting this to work.)

Throne3d:
(This is especially weird given that the `bundle install` step of the build process works just fine and seems to install the right gems. /shrug)

README.md
```diff
@@ -12,13 +12,9 @@
...
-This README mostly focuses on how to get started developing this project with Ubuntu – commands including `apt` or `apt-get` use the Debian/Ubuntu package manager.
-If you are using another flavor of UNIX (another Linux, such as Arch Linux, or macOS), you may find that a lot of the commands are similar but specifics will likely vary – the links to the individual dependency manuals may help you in this case.
-
-If you're using Windows, you may find your best bet – for a variety of reasons – is to set up a virtual machine running Ubuntu where you can do your development.
-This is not completely trivial, but if you choose to do so it would be worth looking into [Oracle VirtualBox](https://www.virtualbox.org/).
-
-It should be possible to develop this project otherwise on a Windows machine, but the steps to install all the dependencies on Windows are quite different, and a lot of the steps will differ from those listed here.
+This README mostly focuses on how to get started developing this project with
+Docker -- in any environment with Docker installed you should be able to have
+the Glowfic server up and running.
```
Throne3d: Nitpick for here and other places: I prefer linebreaks to occur at the end of sentences in documentation, so rephrasing sentences won't cause diffs for the whole paragraph (only the lines modified). Could you modify the hard wrapping accordingly?

README.md
```diff
@@ -41,144 +37,58 @@ It should be possible to develop this project otherwise on a Windows machine, bu
...
+Then run `docker-compose up` within the glowfic directory. This will set up a
+Postgres server, a Redis server, and a Glowfic app server which can talk to
+both of them.
 
-*   `gem install bundler`
-*   `bundler`
+There are some scripts for interacting with the Glowfic servir in bin-docker.
+You may want to add them to your PATH. If you want, you can use
+[direnv](https://direnv.net/) to ensure that they're on your path only when
+you're in the glowfic server. In any case, you'll use the `rake ` and `rails`
```
Throne3d: Nitpick: erroneous space after `rake`

README.md
MichaelBlume: Done =)

README.md
MichaelBlume: Done =)

README.md
```diff
@@ -41,144 +35,55 @@ It should be possible to develop this project otherwise on a Windows machine, bu
...
+If you haven't already, [start by installing Docker](https://docs.docker.com/install/).
 
-Once you've installed Ruby, you need to install the necessary dependencies.
-Enter the appropriate folder (`cd ~/Documents/glowfic`) and then execute:
+Then run `docker-compose up` within the glowfic directory.
+This will set up a Postgres server, a Redis server, and a Glowfic app server which can talk to both of them.
 
-*   `gem install bundler`
-*   `bundler`
+There are some scripts for interacting with the Glowfic servir in bin-docker.
```
Throne3d: Glowfic server*

README.md
```diff
@@ -41,144 +35,55 @@ It should be possible to develop this project otherwise on a Windows machine, bu
...
+You may want to add them to your PATH.
+If you want, you can use [direnv](https://direnv.net/) to ensure that they're on your path only when you're in the glowfic server.
+In any case, you'll use the `rake` and `rails` commands in that directory to set up and start the server.
 
-In addition, the 'execjs' library requires a 'JavaScript runtime'.
-For a full list, look at the [ExecJS README](https://github.com/rails/execjs).
-Otherwise (and it may be overkill), you can install [NodeJS](https://nodejs.org/en/download/package-manager/).
-On Ubuntu this can be done with `sudo apt-get install nodejs`.
+If you have added these scripts to your PATH you can leave off "bin-docker/" in any commands that follow
```
Throne3d: Should have a trailing period

MichaelBlume:
Done! Thanks =)

PR #873, opened 2018-09-27
Protect Anya pending block system

Marri:


Marri:
Things seem to have tentatively settled. This PR can be easily reopened if necessary, but for now I don't think it is.

PR #874, opened 2018-09-28
Make font sizes relative

teceler:


Throne3d:
From #constellation-dev, I think it'd be good if we used rem with a px fallback: <https://snook.ca/archives/html_and_css/font-size-with-rem> (see "Sizing with rem").

teceler:
All of our tentatively supported browsers support `rem`, I don't think a fallback is needed.

PR #875, opened 2018-09-29
Add blocks CRUD

teceler:


app/controllers/blocks_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+    @block.blocked_user_id = params.fetch(:block, {})[:blocked_user_id]
+
+    if @block.save
+      flash[:success] = "User blocked!"
+      redirect_to blocks_path
+    else
+      flash.now[:error] = {}
+      flash.now[:error][:message] = "User could not be blocked."
+      flash.now[:error][:array] = @block.errors.full_messages
+      @users = [User.find_by(id: params.fetch(:block, {})[:blocked_user_id])]
```
Marri:
```suggestion
      @users = [@block.blocked_user]
```

app/controllers/blocks_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+      @page_title = 'Block User'
+      render :action => :new
```
Marri: You need to call editor_setup here as well probably

app/controllers/blocks_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+  def update
+    if @block.update(permitted_params)
+      flash[:success] = "Block updated!"
+      redirect_to blocks_path
+    else
+      flash.now[:error] = {}
+      flash.now[:error][:message] = "Block could not be saved."
+      flash.now[:error][:array] = @block.errors.full_messages
+      @page_title = 'Edit Block: ' + @block.blocked_user.username
+      render :action => :edit
```
Marri: editor_setup again

app/controllers/blocks_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+      flash[:error][:message] = "User could not be unblocked."
+      flash[:error][:array] = @block.errors.full_messages
+    end
+    redirect_to blocks_path
+  end
+
+  private
+
+  def permitted_params
+    params.fetch(:block, {}).permit(:block_interactions, :hide_me, :hide_them)
```
Marri: Any reason not to include blocking_user_id here? Just because of edit or?

app/models/block.rb
```diff
@@ -0,0 +1,42 @@
+class Block < ApplicationRecord
+  belongs_to :blocking_user, class_name: 'User', optional: false, inverse_of: :blocks
+  belongs_to :blocked_user, class_name: 'User', optional: false
+
+  validates :blocking_user_id, uniqueness: { scope: :blocked_user_id }
+  validates :hide_them, :hide_me, inclusion: { in: 0..2 }
+  validate :not_blocking_self
```
Marri: Note to selves: need another validation to make sure at least one of hide them / hide me / interact is set

app/views/blocks/index.haml
```diff
@@ -0,0 +1,51 @@
+%span.time-loaded= pretty_time(DateTime.now.in_time_zone)
+
+%table
+  %thead
+    %tr
+      %th{colspan: 5}
+        = @page_title
+        - if logged_in?
```
Marri: This can come out, you need to be logged in to see the page

app/views/blocks/index.haml
```diff
@@ -0,0 +1,51 @@
...
+            None
+        %td.padding-5{class: klass}
+          - if block.hide_my_content?
+            All
+          - elsif block.hide_my_posts?
+            Posts
+          - else
+            None
+        %td.width-70.right-align{class: klass}
+          - unless local_assigns[:hide_buttons]
```
Marri: Bad copy/paste maybe?

app/models/block.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  scope :ordered, -> { includes(:blocked_user).sort_by { |block| [block.blocked_user.username.downcase] } }
+
+  NONE = 0
+  POSTS = 1
+  ALL = 2
+
+  def editable_by?(user)
+    return false unless user
+    self.blocking_user.id == user.id
```
Marri:
```suggestion
    self.blocking_user_id == user.id
```

app/controllers/api/v1/users_controller.rb
```diff
@@ -7,13 +7,18 @@ class Api::V1::UsersController < Api::ApiController
...
   param :page, :number, required: false, desc: 'Page in results (25 per page)'
+  param :hide_unblockable, ['true', 'false'], required: false, desc: "If set to 'true', elimiates users who cannot be blocked from the result"
   error 422, "Invalid parameters provided"
   def index
     queryset = if params[:match] == 'exact'
       User.where(username: params[:q])
     else
       User.where("username LIKE ?", params[:q].to_s + '%').ordered
     end
+    if params[:hide_unblockable].present? && params[:hide_unblockable] == 'true'
```
Marri: technically might require a logged_in? check but this is not urgent cause this flow shouldn't happen except for logged in users so it's just error hardening the API against bad users. Historical note that including API examples of this wouldn't be helpful because one of APIPie's current gaps is that it doesn't show the _request_ format.

app/controllers/blocks_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+  before_action :editor_setup, only: [:new, :edit]
+
+  def index
+    @page_title = "Blocked Users"
+    @blocks = Block.where(blocking_user: current_user)
+  end
+
+  def new
+    @page_title = "Block User"
+    @block = Block.new
```
Marri:
```suggestion
    @block = Block.new(blocking_user: current_user)
```

app/controllers/blocks_controller.rb
teceler: because we don't want to allow that on edit, yeah

spec/controllers/api/v1/users_controller_spec.rb
```diff
@@ -38,5 +38,14 @@ def create_search_users
...
     end
+
+    it "handles hiding unblockable users" do
+      user = create(:user)
+      login_as(user)
+      create_list(:block, 2, blocking_user: user)
+      create_list(:user, 3)
+      get :index, params: { hide_unblockable: true }
+      expect(response.json['results'].count).to eq(3)
+    end
```
Throne3d: Can you add a test that it doesn't hide them unless the parameter is sent?

spec/controllers/blocks_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
+require 'rails_helper'
+
+RSpec.describe BlocksController, type: :controller do
+  describe "GET index" do
+    it "requires login" do
+      get :index
+      expect(response).to redirect_to(root_url)
```
Throne3d: Does this get a flash? Should we test that it has that flash?

spec/controllers/blocks_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+    end
+
+    it "succeeds" do
+      login
+      get :index
+      expect(response.status).to eq(200)
+    end
```
Throne3d: Can we test this succeeds _also_ with extant blocks?

Ideally we'd feature spec this, too, so we load the views up.

spec/controllers/blocks_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+      post :create
+      expect(response).to redirect_to(root_url)
+    end
+
+    it "requires a blocked user" do
+      login
+      post :create
+      expect(response).to render_template('new')
+      expect(flash[:error][:message]).to eq("User could not be blocked.")
+      expect(flash[:error][:array]).to be_present
```
Throne3d: Could you require a specific error array here?

Throne3d: And in the tests below, too

spec/controllers/blocks_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+    it "requires an existing blocked user" do
+      login
+      post :create, params: { block: { blocked_user_id: ([1] + User.order(id: :desc).limit(1).pluck(:id)).max + 1 } }
+      expect(response).to render_template('new')
+      expect(flash[:error][:message]).to eq("User could not be blocked.")
+      expect(flash[:error][:array]).to be_present
+      expect(assigns[:block].blocked_user).to be_nil
+    end
+
+    it "requires users not be the same" do
```
Throne3d: does not let the user block themself*

spec/controllers/blocks_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+
+    it "requires valid variables" do
+      login
+      post :create, params: { block: { blocked_user_id: create(:user).id, hide_me: -1 } }
+      expect(response).to render_template('new')
+      expect(flash[:error][:message]).to eq("User could not be blocked.")
+      expect(flash[:error][:array]).to be_present
+    end
+
+    it "requires an existing blocked user" do
```
Throne3d: requires an existing user to block*

spec/controllers/blocks_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+      expect(response).to render_template('new')
+      expect(flash[:error][:message]).to eq("User could not be blocked.")
+      expect(flash[:error][:array]).to be_present
+      expect(flash[:error][:array]).to include("User cannot block themself")
+    end
+
+    it "succeeds" do
+      login
+      post :create, params: { block: { blocked_user_id: create(:user).id } }
+      expect(response).to redirect_to(blocks_url)
```
Throne3d: Can we check the flash, if we give one, and also make sure it persists the new block in the database? `expect { the_post_create_thing }.to change { Blocks.count }.by(1)` or such

Throne3d: Or are you doing that below? (See further comment there)

spec/controllers/blocks_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+    end
+
+    it "sets expected variables" do
```
Throne3d: Is this on a failure? If so, what's the failure? We should be verifying that it fails, if it does.

spec/controllers/blocks_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+    end
+  end
+
+  describe "GET edit" do
+    it "requires permission" do
+      block = create(:block)
+      login
+      get :edit, params: { id: block.id }
+      expect(response).to redirect_to(blocks_url)
+      expect(flash[:error]).to eq("You do not have permission to edit that block.")
```
Throne3d: Might be worthwhile using the same error as below, for this, so people can't get data about how many blocks exist. (That is, respond "Block could not be found." if the user doesn't have permission to edit it.)

spec/controllers/blocks_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+    end
+  end
+
+  describe "PUT update" do
+    it "requires permission" do
+      block = create(:block)
+      login
+      put :update, params: { id: block.id }
+      expect(response).to redirect_to(blocks_url)
+      expect(flash[:error]).to eq("You do not have permission to edit that block.")
```
Throne3d: Same as in `GET edit` – respond that we don't know what block they're talking about, instead of that they don't have permission to edit this one.

spec/controllers/blocks_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+    it "requires valid variables" do
+      block = create(:block)
+      login_as(block.blocking_user)
+      put :update, params: { id: block.id, block: { hide_them: -1 } }
+      expect(response).to render_template('edit')
+      expect(flash[:error][:message]).to eq("Block could not be saved.")
+      expect(flash[:error][:array]).to be_present
+    end
+
+    it "succeeds" do
```
Throne3d: Is this split from the test below intentionally? I'd prefer if they could be merged, if they're testing the same thing (that successful changes do particular things), since as is, this test isn't doing much – in the real world, nobody will submit an update with no specified data.

spec/controllers/blocks_controller_spec.rb
```diff
@@ -0,0 +1,198 @@
...
+    end
+  end
+
+  describe "DELETE destroy" do
+    it "requires permission" do
+      block = create(:block)
+      login
+      delete :destroy, params: { id: block.id }
+      expect(response).to redirect_to(blocks_url)
+      expect(flash[:error]).to eq("You do not have permission to edit that block.")
```
Throne3d: (Same as `GET edit`)

Throne3d:
Migration for blocks is here: https://github.com/Marri/glowfic/blob/master/db/migrate/20180928230642_create_blocks.rb

spec/models/block_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+    expect(block.hide_their_content?).to be(true)
+  end
+
+  it "should return the correct values for posts only" do
+    block = create(:block, hide_me: Block::POSTS, hide_them: Block::POSTS)
+    expect(block.hide_my_posts?).to be(true)
+    expect(block.hide_my_content?).to be(false)
+    expect(block.hide_their_posts?).to be(true)
+    expect(block.hide_their_content?).to be(false)
+  end
```
Throne3d: Might as well test "for interactions only" for completeness purposes?

spec/controllers/blocks_controller_spec.rb
teceler: The flash is a temporary 'this is incomplete' notice, so no.

spec/models/block_spec.rb
Throne3d: (Gonna test for the other combinations too.)

PR #877, opened 2018-09-30
Mark a bunch of bare SQL string literals as "safe"

MichaelBlume:


app/controllers/api/v1/board_sections_controller.rb
```diff
@@ -22,7 +22,7 @@ def reorder
       render json: {errors: [error]}, status: :not_found and return
     end
 
-    boards = Board.where(id: sections.pluck('distinct board_id'))
+    boards = Board.where(id: sections.pluck(Arel.sql('distinct board_id')))
```
Marri: Clever abstract: sections.select(:board_id).distinct.pluck(:board_id). There's probably an even prettier way to do it but I'd have to go through the rest of the method and make sure I didn't break anything.

PR #879, opened 2018-09-30
Upgrade to Rails 5.2

MichaelBlume:
Fails


MichaelBlume:
No longer fails! Should work. Prints a *lot* of deprecation warnings but those will be 100% cleaned up by #877

MichaelBlume:
This probably contains a lot of unrelated upgrades -- I deleted and regenerated the Gemfile.lock

MichaelBlume:
Ok I should probably disclaim that all I know about this branch is that it passes tests and if there is more QA to be done on a change of this scope (in particular if there are library dependencies whose responsibilities aren't covered by tests) I probably don't know how to do it yet

PR #883, opened 2018-10-02
[WIP] Avoid chaining arrays

teceler:


app/models/character.rb
```diff
@@ -48,7 +48,7 @@ def deletable_by?(user)
   def recent_posts
     return @recent unless @recent.nil?
     reply_ids = replies.group(:post_id).pluck(:post_id)
-    post_ids = posts.select(:id).map(&:id)
+    post_ids = posts.select(:id).pluck(:id)
```
Throne3d: This `select` bit should be unnecessary too.

Throne3d:
What was WIP about this again?

Throne3d:
can we rebase this? but I think a lot of this might not be necessary anymore, since I think we enabled this lint already

PR #905, opened 2018-10-29
Refactor the post import flow to use a service

### LINKED ISSUE #711: AlreadyImportedErrors should halt at the same time as looking for screennames, not later
Currently the UI proceeds through and then emails to mention that the thread has already been imported; it should notify at the earlier stage, when looking for missing screennames, instead.

Marri:
So this is an adventure in questions.
- Do we like this style of services? We've also talked about a PostList service iirc. (We do)
- Do we want to also make app/services contain some kind of subfolder for this kind of scraping logic and move PostScraper into it? (We do not)
- How do we name PostImporter vs PostScraper more clearly? (We leave this problem for future!us)
- Is all of this overkill given scraping is, hopefully, not permanent long term? (We're stuck with it long enough to care)
- Do we like this style of exceptions? I'm not huge on exception flow in general, and yet, here we are (we do)
- Why have I not yet written tests for the new service? (Now I have)
- Why does any of this matter literally at all except to fix #711 in a MASSIVELY overbuilt fashion? (It does not)
- I mean I guess code quality and CodeClimate??? (Code quality!)

Fixes #711


app/services/post_importer.rb
```diff
@@ -0,0 +1,69 @@
+class PostImporter < Object
+  def initialize(url)
+    @url = url
+  end
+
+  def import(board_id, section_id, status, threaded, importer_id)
```
Throne3d: Should we use keyword params here, for clarity?

spec/support/spec_test_helper.rb
```diff
@@ -7,3 +7,9 @@ def login
     login_as(create(:user))
   end
 end
+
+def stub_fixture(url, filename)
+  url = url.gsub(/\#cmt\d+$/, '')
```
Throne3d: Maybe we should parse the URL here using `URI` and then remove the whole fragment, to make it more general?

app/services/post_importer.rb
```diff
@@ -0,0 +1,69 @@
...
+  def validate_url!
+    raise InvalidDreamwidthURL.new('Invalid URL provided.') unless self.class.valid_dreamwidth_url?(@url)
+  end
+
+  def validate_duplicate!(board_id)
+    subject = dreamwidth_doc.at_css('.entry .entry-title').text.strip
+    subj_post = Post.where(subject: subject, board_id: board_id).first
+    return unless subj_post
+    raise AlreadyImported.new("This thread has already been imported! " + ScrapePostJob.view_post(subj_post.id))
+  end
```
Throne3d: Given this code is identical to https://github.com/Marri/glowfic/blob/9389578a83fd7b2056d59920669e31189be165b8/app/services/post_scraper.rb#L162-L166, should we try to deduplicate this somehow?

app/controllers/posts_controller.rb
```diff
@@ -347,48 +347,15 @@ def editor_setup
...
-    return false unless parsed_url.host
-    parsed_url.host.ends_with?('dreamwidth.org')
-  rescue URI::InvalidURIError
-    false
+  rescue PostImportError => e
+    flash.now[:error] = e.api_error
+    params[:view] = 'import'
+    editor_setup
+    render action: :new
   end
```
Throne3d: This `import_thread` method looks a lot cleaner now! :D (Yes, I realize that's the point, but still!)

PR #906, opened 2018-10-31
Fix displayed icon count

### LINKED ISSUE #901: Icon usage count includes non-visible threads
Potential solution: filter to replies from posts that are public or constellation user only (as opposed to access listed or private)?

teceler:
* Moves calculating the times an icon is used and the number of posts it's used in from the view to the controller
* Limits the posts the count is determined from to public posts, and constellation-only posts for logged in users who are not the icon owner (the count is should be unaffected for the icon owner) (Fixes #901)

### DIFF
--- a/app/controllers/icons_controller.rb
+++ b/app/controllers/icons_controller.rb
@@ -53,6 +53,15 @@ class IconsController < UploadingController
     else
       posts_using = Post.where(icon_id: @icon.id)
       replies_using = Reply.where(icon_id: @icon.id)
+      if logged_in?
+        unless @icon.user == current_user
+          posts_using = posts_using.where(privacy: [Concealable::PUBLIC, Concealable::REGISTERED])
+          replies_using = replies_using.where(posts: {privacy: [Concealable::PUBLIC, Concealable::REGISTERED]}).joins(:post)
+        end
+      else
+        posts_using = posts_using.where(privacy: Concealable::PUBLIC)
+        replies_using = replies_using.where(posts: {privacy: Concealable::PUBLIC}).joins(:post)
+      end
       @times_used = (posts_using.count + replies_using.count)
       @posts_used = (posts_using.pluck(:id) + replies_using.pluck('distinct post_id')).uniq.count
     end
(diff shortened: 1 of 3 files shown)


app/controllers/icons_controller.rb
```diff
@@ -49,6 +49,20 @@ def show
       @posts = posts_from_relation(Post.where(where_calc).ordered)
     elsif params[:view] == 'galleries'
       use_javascript('galleries/expander_old')
+    else
+      posts_using = Post.where(icon_id: @icon.id)
+      replies_using = Reply.where(icon_id: @icon.id)
+      if logged_in?
+        unless @icon.user == current_user
+          posts_using = posts_using.where(privacy: [Concealable::PUBLIC, Concealable::REGISTERED])
+          replies_using = replies_using.where(posts: {privacy: [Concealable::PUBLIC, Concealable::REGISTERED]}).joins(:post)
```
Throne3d: Can we put the `joins` before the `where`? Makes it easier for me to logic about it.

app/controllers/icons_controller.rb
```diff
@@ -49,6 +49,20 @@ def show
       @posts = posts_from_relation(Post.where(where_calc).ordered)
     elsif params[:view] == 'galleries'
       use_javascript('galleries/expander_old')
+    else
```
Throne3d: Can you put an explanatory comment for this chain of the if statement? Something like:

```rb
# Approximate icon counts for non-owner
# Check only more obviously visible posts for icon counts
# (public posts, and constellation-only posts if logged in)
```

spec/controllers/icons_controller_spec.rb
```diff
@@ -235,6 +235,43 @@
...
+        expect(assigns(:posts_used)).to eq(3)
+      end
+
+      it "fetches corect counts when logged in" do
+        get :show, params: { id: icon.id }
+        expect(response).to have_http_status(200)
+        expect(assigns(:times_used)).to eq(3)
+        expect(assigns(:posts_used)).to eq(2)
+      end
+    end
```
Throne3d: Nice!

PR #907, opened 2018-11-01
Blocks: Add functionality to block interaction

teceler:
(depends on #875)

* Prevents messaging blocked or blocking users
* Prevents adding as coauthors blocked or blocking users

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -18,7 +18,7 @@ class MessagesController < ApplicationController
   def new
     @message = Message.new
     recipient = User.find_by_id(params[:recipient_id])
-    @message.recipient = recipient unless recipient&.has_interaction_blocked?(current_user)
+    @message.recipient = recipient unless recipient && current_user.has_interaction_blocked?(recipient)
     @page_title = 'Compose Message'
   end
 
@@ -116,7 +116,7 @@ class MessagesController < ApplicationController
     use_javascript('messages')
     unless @message.try(:parent)
       recent_ids = Message.where(sender_id: current_user.id).order(Arel.sql('MAX(id) desc')).limit(5).group(:recipient_id).pluck(:recipient_id)
-      base_users = User.where.not(id: [current_user.id] + current_user.blocked_interaction_user_ids(receiver_direction: 'blocking'))
+      base_users = User.where.not(id: [current_user.id] + current_user.user_ids_blocked_interaction)
       recents = base_users.where(id: recent_ids).pluck(:username, :id).sort_by{|x| recent_ids.index(x[1]) }
       users = base_users.ordered.pluck(:username, :id)
       @select_items = if recents.present?
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -116,7 +116,7 @@ class MessagesController < ApplicationController
     use_javascript('messages')
     unless @message.try(:parent)
       recent_ids = Message.where(sender_id: current_user.id).order(Arel.sql('MAX(id) desc')).limit(5).group(:recipient_id).pluck(:recipient_id)
-      base_users = User.where.not(id: [current_user.id] + current_user.blocked_interaction_user_ids(reciever_direction: 'blocking'))
+      base_users = User.where.not(id: [current_user.id] + current_user.blocked_interaction_user_ids(receiver_direction: 'blocking'))
       recents = base_users.where(id: recent_ids).pluck(:username, :id).sort_by{|x| recent_ids.index(x[1]) }
       users = base_users.ordered.pluck(:username, :id)
       @select_items = if recents.present?
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -116,7 +116,7 @@ class MessagesController < ApplicationController
     use_javascript('messages')
     unless @message.try(:parent)
       recent_ids = Message.where(sender_id: current_user.id).order(Arel.sql('MAX(id) desc')).limit(5).group(:recipient_id).pluck(:recipient_id)
-      base_users = User.where.not(id: [current_user.id] + current_user.block_interaction_users)
+      base_users = User.where.not(id: [current_user.id] + current_user.blocked_interaction_users(reciever_direction: 'either'))
       recents = base_users.where(id: recent_ids).pluck(:username, :id).sort_by{|x| recent_ids.index(x[1]) }
       users = base_users.ordered.pluck(:username, :id)
       @select_items = if recents.present?
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -17,7 +17,8 @@ class MessagesController < ApplicationController
 
   def new
     @message = Message.new
-    @message.recipient = User.find_by_id(params[:recipient_id])
+    recipient = User.find_by_id(params[:recipient_id])
+    @message.recipient = recipient if recipient&.can_interact_with?(current_user)
     @page_title = 'Compose Message'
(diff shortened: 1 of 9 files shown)


app/models/message.rb
```diff
@@ -6,8 +6,10 @@ class Message < ApplicationRecord
   belongs_to :first_thread, class_name: 'Message', foreign_key: :thread_id, inverse_of: false, optional: false
 
   validates :sender, presence: { if: Proc.new { |m| m.sender_id != 0 } }
+  validate :unblocked_recipient
 
   before_validation :set_thread_id
+  before_create :validate_recipient
```
Marri: Let's not call this validate if it's a before_create

app/models/post.rb
```diff
@@ -281,6 +281,18 @@ def valid_board_section
     errors.add(:section, "must be in the post's board")
   end
 
+  def valid_coauthors
```
Marri: You're not actually guaranteed that the person editing is post.user unfortunately, this might need to happen controller-side

app/controllers/messages_controller.rb
```diff
@@ -17,7 +17,8 @@ def index
 
   def new
     @message = Message.new
-    @message.recipient = User.find_by_id(params[:recipient_id])
+    recipient = User.find_by_id(params[:recipient_id])
+    @message.recipient = recipient if recipient&.can_interact_with?(current_user)
```
Marri: I'm finding following all the different method names to be surprisingly hard; that said, I don't think we should strip users unless the current_user has blocked the recipient, and I think you strip them both ways?

app/models/post.rb
teceler: I think what I did was check against the post.user and joined_coauthors

app/models/post.rb
```diff
@@ -281,6 +281,18 @@ def valid_board_section
...
+    return if self.unjoined_authors.empty?
+    return if self.user.blocked_interaction_users(reciever_direction: 'either').empty? && self.joined_authors.empty?
+    self.unjoined_authors.each do |unjoined|
+      next if unjoined.blocked_interaction_users(reciever_direction: 'either').empty?
```
Marri: I am going to be That Person cause `receiver` is spelled wrong :/

app/models/post.rb
```diff
@@ -281,6 +281,18 @@ def valid_board_section
...
+      errors.add(:post_author, "cannot be added") unless self.user.can_interact_with?(unjoined)
```
Marri: I have not mentally parsed this entire loop but I am confident we need a break somewhere

app/models/user.rb
```diff
@@ -72,6 +72,29 @@ def default_view
...
+    !blocked_interaction_users(reciever_direction: 'either').include?(user.id)
+  end
+
+  def has_interaction_blocked?(user)
+    blocked_interaction_users(reciever_direction: 'blocking').include?(user.id)
+  end
+
+  def blocked_interaction_users(reciever_direction:)
+    unless ['blocked', 'blocking', 'either'].include?(reciever_direction)
+      throw ArgumentError("Must pass one of 'blocked', blocking', 'either'")
```
Marri: Nonblocking: curious why throw not raise

app/controllers/messages_controller.rb
Marri: This is still both ways

teceler:
TODO: modify flash warning on blocks index

spec/models/post_spec.rb
```diff
@@ -419,6 +419,22 @@
       post = create(:post, content: '')
       expect(post.id).not_to be_nil
     end
+
+    it "should not allowed blocking coauthor" do
```
Throne3d:
```suggestion
    it "should not allow adding a coauthor who's blocked you" do
```

spec/models/post_spec.rb
```diff
@@ -419,6 +419,22 @@
...
+      author = create(:user)
+      coauthor = create(:user)
+      create(:block, blocking_user: coauthor, blocked_user: author, block_interactions: true)
+      post = build(:post, user: author, unjoined_authors: [coauthor])
+      expect(post).not_to be_valid
+    end
+
+    it "should not allow blocked coauthor" do
```
Throne3d:
```suggestion
    it "should not allow adding a coauthor you've blocked" do
```

app/models/post.rb
```diff
@@ -281,6 +281,21 @@ def valid_board_section
...
+      next if unjoined.user_ids_uninteractable.empty?
+      joined_authors = self.joined_authors + [self.user]
+      joined_authors.each do |author|
+        unless author.can_interact_with?(unjoined)
+          errors.add(:post_author, "cannot be added")
+          break
+        end
+      end
+    end
+  end
```
Throne3d: It looks like we can add two people to a post where one has blocked the other, and then if either of them tries to join the thread, this will get an error? I'm not sure if that's intentional, but I think it's the case.

spec/controllers/messages_controller_spec.rb
```diff
@@ -60,6 +60,34 @@
...
+      login_as(block.blocked_user)
+      get :new, params: { recipient_id: block.blocking_user_id }
+      expect(response.status).to eq(200)
+      expect(assigns(:message).recipient_id).to be(block.blocking_user_id)
+    end
+
+    it "hides blocked users" do
+      user = create(:user)
+      login_as(user)
+      blocking = create_list(:block, 2, blocked_user: user).pluck(:blocking_user_id)
```
Throne3d: You could simply map this to `:blocking_user`, and then you don't need to do `User.find(blocking)` below.

spec/controllers/messages_controller_spec.rb
```diff
@@ -60,6 +60,34 @@
...
+      create_list(:block, 2, blocking_user: user)
+      other_users = create_list(:user, 2) + User.find(blocking)
+      other_users = other_users.sort_by!(&:username).pluck(:username, :id)
+      get :new
+      expect(response.status).to eq(200)
+      expect(assigns(:select_items)).to match_array(Users: other_users)
```
Throne3d: This should be `eq`, not `match_array`, right? Since they're both hashes.

spec/models/message_spec.rb
```diff
@@ -31,5 +31,24 @@
...
+  end
+
+  it "validates recipient" do
+    sender = create(:user)
+    recipient = create(:user)
+    create(:block, blocking_user: recipient, blocked_user: sender, block_interactions: true)
+    message = build(:message, sender: sender, recipient: recipient)
+    expect(message).to be_valid
+    message.save!
+    expect(message.visible_inbox).to eq(false)
```
Throne3d: What's happening here? This doesn't seem to be testing anything different from the test above. We also don't seem to have a test that you can't send a message to someone you've got blocked, which I would have assumed this would test.

spec/models/post_spec.rb
```diff
@@ -419,6 +419,22 @@
...
+      expect(post).not_to be_valid
+    end
+
+    it "should not allow adding a coauthor you've blocked" do
+      author = create(:user)
+      coauthor = create(:user)
+      create(:block, blocking_user: author, blocked_user: coauthor, block_interactions: true)
+      post = build(:post, user: author, unjoined_authors: [coauthor])
+      expect(post).not_to be_valid
+    end
```
Throne3d: Can we also test this prevents you adding coauthors who blocked / are blocked by some other coauthor in the post? That is, create a post, create a reply by a new author, create some block between this new author and a new user, then try amending the post to have the new user as an unjoined author.

teceler:
Closing in favor of #963

PR #908, opened 2018-11-01
Blocks: Add functionality to hide posts

teceler:
a. Allows hiding the posts of people the user has blocked
b. Allows hiding the user's posts from people they have blocked
c. Allows temporarily disabling a, but _not_ b with a url param (manually added only, currently)
d. Prevents notifications being sent about posts that would be hidden under a or b.

### DIFF
--- a/app/concerns/blockable.rb
+++ b/app/concerns/blockable.rb
@@ -13,6 +13,11 @@ module Blockable
     user_ids_blocked_interaction.include?(user.id)
   end
 
+  def author_blocking?(post, author_ids)
+    return false unless post.authors_locked
+    Block.where(blocking_user_id: author_ids, blocked_user: self).where("hide_me >= ?", Block::POSTS).exists?
+  end
+
   def user_ids_blocked_interaction
     Block.where(block_interactions: true, blocking_user: self).pluck(:blocked_user_id)
   end
@@ -23,5 +28,13 @@ module Blockable
 
   def user_ids_uninteractable
     (user_ids_blocking_interaction + user_ids_blocked_interaction).uniq
+  end
+
+  def hidden_post_users
+    (blocking_post_users + Block.where(blocking_user: self).where("hide_them >= ?", Block::POSTS).pluck(:blocked_user_id)).uniq
+  end
+
+  def blocking_post_users
+    Block.where(blocked_user: self).where("hide_me >= ?", Block::POSTS).pluck(:blocking_user_id)
   end
 end
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -9,7 +9,7 @@ class PostsController < WritableController
   before_action :editor_setup, only: [:new, :edit]
 
   def index
-    @posts = posts_from_relation(Post.ordered)
+    @posts = posts_from_relation(Post.ordered, show_blocked: !!params[:show_blocked])
     @page_title = 'Recent Threads'
   end
 
@@ -59,7 +59,7 @@ class PostsController < WritableController
       .or(with_post_view.where("date_trunc('second', post_views.read_at) < date_trunc('second', posts.tagged_at)"))
 
     @posts = with_post_view.or(no_post_view)
-    @posts = posts_from_relation(@posts.ordered, with_unread: true)
+    @posts = posts_from_relation(@posts.ordered, with_unread: true, show_blocked: !!params[:show_blocked])
 
     @hide_quicklinks = true
     @page_title = @started ? 'Opened Threads' : 'Unread Threads'
@@ -280,7 +280,7 @@ class PostsController < WritableController
       post_ids = Reply.where(character_id: params[:character_id]).select(:post_id).distinct.pluck(:post_id)
       @search_results = @search_results.where(character_id: params[:character_id]).or(@search_results.where(id: post_ids))
     end
-    @search_results = posts_from_relation(@search_results)
+    @search_results = posts_from_relation(@search_results, show_blocked: !!params[:show_blocked])
   end
 
   def warnings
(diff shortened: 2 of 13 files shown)


app/controllers/posts_controller.rb
```diff
@@ -48,9 +48,7 @@ def unread
 
   def mark
     posts = Post.where(id: params[:marked_ids])
-    posts = posts.select do |post|
-      post.visible_to?(current_user)
-    end
+    posts = posts.select { |post| post.visible_to?(current_user, true) }
```
Marri: why is this one true instead of param based?

app/models/post.rb
```diff
@@ -77,7 +77,9 @@ class Post < ApplicationRecord
     select('(SELECT COUNT(*) FROM replies WHERE replies.post_id = posts.id) AS reply_count')
   }
 
-  def visible_to?(user)
+  def visible_to?(user, show_blocked=false)
```
Marri: I can't tell if you're using named args for this or not?

app/models/post.rb
```diff
@@ -77,7 +77,9 @@ class Post < ApplicationRecord
...
+    return false if !show_blocked && user&.author_hidden?(self, author_ids)
```
Marri: authorship (and authors_locked below) seems distinct from if the post is viewable at all; this seems like two mixed concepts

app/controllers/posts_controller.rb
```diff
@@ -48,9 +49,7 @@ def unread
 
   def mark
     posts = Post.where(id: params[:marked_ids])
-    posts = posts.select do |post|
-      post.visible_to?(current_user)
-    end
+    posts = posts.select { |post| post.visible_to?(current_user) }
```
Marri: Purely informationally: `posts` here is indeed a Queryset not an Array so the issue with `select!` was probably Querysets doing nonsense to differentiate from the SQL select function.

teceler:
TODO: modify flash warning on blocks index

app/concerns/blockable.rb
```diff
@@ -13,6 +13,16 @@ def has_interaction_blocked?(user)
     user_ids_blocked_interaction.include?(user.id)
   end
 
+  def author_hidden?(post, author_ids)
```
Marri: What's this for? I don't see it used?

teceler:
Currently blocked on scope issues

app/concerns/blockable.rb
```diff
@@ -13,6 +13,11 @@ def has_interaction_blocked?(user)
     user_ids_blocked_interaction.include?(user.id)
   end
 
+  def author_blocking?(post, author_ids)
+    return false unless post.authors_locked
+    author_ids.any?{ |author| self.blocking_post_users.include?(author) }
```
teceler, line 18: I don't remember if I use this anywhere but if I do it should probably be 
 ```Block.where(blocking_user_id: author_ids, blocked_user: self).where("hide_me >= ?", Block::POSTS).exists?``` or similar

Marri: I do not presently disagree but worth mentioning this flavor is more future-cache-friendly

teceler: Hmm. Would intersection be equally friendly? I think it would work here?

app/models/block.rb
```diff
@@ -56,4 +63,10 @@ def mark_messages_read
       message.save
     end
   end
+
+  def invalidate_caches
+    return unless created_at == updated_at || saved_change_to_hide_me? || saved_change_to_hide_them?
```
Marri, line 68: can we not use something like new_record? here, I worry about the created/updated check. for that matter, do new records automatically trigger the saved_change_to thing?

teceler: We can't use new_record? because it's... already saved... because we're using after_commit. I think this is the closest I found. And I don't think so but I can check I suppose? I remember getting this working was an Adventure.

app/jobs/notify_followers_of_new_post_job.rb
```diff
@@ -61,4 +63,12 @@ def view_post(post_id)
     url = Rails.application.routes.url_helpers.post_url(post_id, host: host, protocol: 'https')
     "<a href='#{url}'>View it here</a>."
   end
+
+  def blocked_user_ids(post)
+    blocked = Block.where(blocked_user_id: post.author_ids).where("hide_them >= ?", Block::POSTS)
+    blocked = blocked.select(:blocking_user_id).distinct.pluck(:blocking_user_id)
+    blocking = Block.where(blocking_user_id: post.author_ids).where("hide_me >= ?", Block::POSTS)
+    blocking = blocking.select(:blocked_user_id).distinct.pluck(:blocked_user_id)
+    (blocked + blocking).uniq
```
teceler, line 72: Would it be better to check the blocked post caches of favorited users here? At the point where this is going out they probably haven't regenerated yet, which means a bunch of seperate db queries where this just queries twice -- and actually could probably query only once -- but it means we're not skipping the cache.

app/views/blocks/index.haml
```diff
@@ -3,7 +3,7 @@
 - content_for :flashes do
   .flash.error
     = image_tag "icons/exclamation.png".freeze, class: 'vmid'.freeze
-    Warning: blocking is not currently implemented. This page should remember any blocks you have created, but they will not take effect yet.
+    Warning: content blocking is not yet implemented, and will function the same as simply blocking posts
```
Marri, line 6: We should clarify somewhere on this page if the post block setting isn't going to hide non-locked posts

teceler: Yes, good point. Hmm... Words are hard.

PR #911, opened 2018-11-01
Enforce uniquness for board names

### LINKED ISSUE #678: Boards should be unique by name
Currently they don't seem to. This seems wrong.

teceler:
Fixes #678 (also contains a few unrelated validation tests for boards)

(Check we don't have any existing identically named boards before releasing!)


Marri:
If you say fixes #678 (or various other keywords besides fixes which I forget, closes is probably one) instead of per, then it will automagically close the issue once the PR is merged :)

PR #912, opened 2018-11-03
Stop auditing Reply.reply_order

teceler:


Marri:
```
right_changes = ["reply_order"]
ids = []
Audited::Audit.where(auditable_type: 'Reply', action: 'update').find_each do |audit|
    next unless audit.audited_changes.keys == right_changes
    ids << audit.id
end
Audited::Audit.where(id: ids).delete_all
```
maybe?

Marri:
...in retrospect this should have been a migration and I may yet commit it as one.

PR #914, opened 2018-11-05
Tests

Marri:
Bump up our code coverage some more. Fine with all of it except that I'm vaguely worried about the check_domain test, please opine.

### DIFF
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -265,6 +265,17 @@ RSpec.describe ApplicationController do
     end
   end
 
+  describe "#require_glowfic_domain" do
+    it "redirects on valid requests" do
+      ENV['DOMAIN_NAME'] ||= 'domaintest.host'
+      expect(controller.request.method).to eq('GET')
+      expect(controller.request.xhr?).to be_nil
+      expect(controller.request.host).not_to include('glowfic')
+      expect(controller).to receive(:redirect_to).with("https://domaintest.host", status: :moved_permanently)
+      controller.send(:check_domain)
+    end
+  end
+
   describe "#page_view" do
     context "when logged out" do
       it "works by default" do
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -1,65 +1,77 @@
 require "spec_helper"
 
 RSpec.describe ApplicationController do
+  controller do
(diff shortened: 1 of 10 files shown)


spec/controllers/application_controller_spec.rb
```diff
@@ -265,6 +265,17 @@
...
 
+  describe "#require_glowfic_domain" do
+    it "redirects on valid requests" do
+      ENV['DOMAIN_NAME'] ||= 'domaintest.host'
+      expect(controller.request.method).to eq('GET')
+      expect(controller.request.xhr?).to be_nil
+      expect(controller.request.host).not_to include('glowfic')
+      expect(controller).to receive(:redirect_to).with("https://domaintest.host", status: :moved_permanently)
+      controller.send(:check_domain)
+    end
```
Throne3d: I had a suggestion in the Discord around this:

> for the check_domain test, it might be worthwhile trying to fake the `Host` parameter of the request (<https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Request_fields>)
> which you should be able to do by specifying `headers: {Host: '…'}` on the get thing: <https://api.rubyonrails.org/v5.2.1/classes/ActionDispatch/Integration/Session.html#method-i-process>
```

That way we can check it happens through a proper flow when we load the site (e.g. by putting it on whatever serves the index page) on the wrong URL, and doesn't happen when we do it on the right URL?

It seems worthwhile to also have this test, though I think it'd make more sense to be labeled `describe "#check_domain"` instead?

File: spec/controllers/posts_controller_spec.rb [hunk 2]
```diff
@@ -2316,6 +2316,60 @@
...
+      end
+    end
+
+    context "with hidden" do
+      let(:user) { create(:user) }
+
+      before(:each) do
+        login_as(user)
+
+        @unhidden_post = create(:post, user: user)
```
Throne3d: Couldn't we use `let`s for these, instead of instance variables? I think that's more RSpec-y, and it should load them just fine so long as we still reference them in this `before(:each)`.

Marri:
I can kind of Github if I use emails! I think your suggestion assumes that
I am actually using request flow; I here am not (because this function
suffers from the same problem as the TOS check in galleries_controller, we
may need a Better Approach) and therefore have no get thing to pass the
Host header to. I did fix the let() thing but can't push just yet.

PR #919, opened 2018-11-06
Terms of Service and Privacy Policy pages

Marri:
Done:
- [x] Require logged in users to accept the new TOS
- [x] Require new users to accept the new TOS
- [x] Require logged out users to accept the new TOS
- [x] DCMA page
- [x] Contact Us page
- [x] Site menu at the bottom of the layout

Remaining:
- [x] AO3 policy 4 - policies for actually handling reported violations
- [x] Review the privacy policy? Is it actually comprehensive?
- [x] Squash the many, many commits

Later issues:
- make a way to make content Not Visible without a user having the ability to undo this for DMCA reasons; deletion is fine for audited stuff but some things are not audited. Or possibly just install more audits, or both.
- make a way for an account to be Suspended (can't create anything text/icon/etc but can read)
- make (finish) a way for an account to be Deleted (see #810)
- I'd like a cookie policy as well but I think it can be a separate project
- GDPR review at this point should probably be separate.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 class SessionsController < ApplicationController
-  before_action :logout_required, only: [:new, :create]
+  before_action :logout_required, only: [:new, :create, :confirm_tos]
   before_action :login_required, only: [:destroy]
 
   def index
@@ -31,6 +31,11 @@ class SessionsController < ApplicationController
     else
       flash[:error] = "You have entered an incorrect password."
     end
+    redirect_to session[:previous_url] || root_url
+  end
+
+  def confirm_tos
+    cookies.permanent[:accepted_tos] = cookie_hash(User::CURRENT_TOS_VERSION)
     redirect_to session[:previous_url] || root_url
   end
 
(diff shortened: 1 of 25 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -34,6 +34,11 @@ def create
     redirect_to session[:previous_url] || root_url
   end
 
+  def update
```
Throne3d: I'm not sure I like the name of this route – updating a session constitutes accepting the ToS? It might be more descriptive if we forced the route to have a different name instead (like `accept_tos`).

app/views/about/contact.haml
```diff
@@ -0,0 +1,12 @@
+#contact-content.odd{style: 'padding: 15px;'}
+  %h1 Contact the Constellation
+
+  %p
+    Currently, to contact the Constellation for help or to report a violation of our
+    = succeed ',' do
+      = link_to 'Terms of Service', tos_path
+    you may visit us in the #constellation channel of the
+    = link_to "Glowfic Discord", 'https://discord.gg/Ss6fRFj', target: '_blank'
```
Throne3d: Is this discord link the same as we have on the homepage? If so, we might want to extract it somewhere so it's easier to amend in the future if/when we need to.

PR #926, opened 2018-11-13
User#edit cleanup

Marri:
In sane commits and everything! Things covered here:

- For sanity's sake, remove the :pass option in flash. I believe it was added because a) we used to have HTML tags in there that needs html_safe and that needed special handling and b) we wanted it not to conflict with other error tags, but we have few enough password-broken users now (and c) I think flashes changed from IndifferentAccess so it was actually incorrectly displaying as flash.pass not flash.error) that I went with the consolidated code for clarity

- The user#edit page no longer specifies url or method for the user edit form, as Rails can detect that itself. (Change password needs specifying still because they're non standard)

- The user#edit page sets the id of the change password form to #change_password_form explicitly, as otherwise it autogenerates an ID which conflicts with the user edit form

- User#update now renders edit on failure instead of redirecting, as this allows us to use the flash message/array syntax to actually explain to the user what they did wrong. (Just updating the flash object wouldn't work, as flash.now objects have :keys but flash objects have "keys".)

- We now support images with the .flash-dismiss class within links that will dismiss their parent div when you click them. This has been installed in our ephemeral error and success divs (but not warnings or inbox) to allow users to dismiss notices without reloading the page.

- Database migration for subcontinuity descriptions in preparation for actually adding them to the code.

### DIFF
--- a/app/views/users/edit.haml
+++ b/app/views/users/edit.haml
@@ -1,4 +1,4 @@
-= form_for current_user, :url => user_path(current_user), :method => :put do |f|
+= form_for current_user do |f|
   %table.form-table
     %tr
       %th.centered{colspan: 2} Account Settings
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/global.js
```diff
@@ -7,6 +7,12 @@ $(document).ready(function() {
     minimumResultsForSearch: 10,
   });
 
+  $(".flash-dismiss").click(function() {
+    // assumes the .flash-dismiss image is within an <a> tag within the flash div
+    $(this).parent().parent().hide(200);
```
Throne3d: `$(this).closest('.flash').hide(200);` should work

PR #939, opened 2018-11-23
Traceroute

Marri:
Traceroute is a route cleaning tool. This adds the tool, adds a config file so we don't run it on Apipie or Resque Web (both of which are mounted from third party gems and not under our control, so no reason to analyze them) and fixes any extant errors found by the tool.

### DIFF
--- a/.traceroute.yml
+++ b/.traceroute.yml
@@ -1,0 +1,7 @@
+ignore_unused_routes:
+- ^apipie\/
+- ^resque_web\/
+ignore_unreachable_actions:
+- ^apipie\/
+- ^api.*#apipie_validations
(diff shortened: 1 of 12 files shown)


Throne3d:
What's up with galleries#search and templates#search? I would have expected them to completely fail without the empty methods, before, but apparently they didn't. (It might be as simple as "route don't actually *need* corresponding methods" – but if so, I didn't realize this.)

This looks good!

Marri:
Afaik, route does not in fact actually need corresponding methods if there's a haml file to fall back on, but adding these doesn't bother me much cause we'll need them once we remove the Under Construction anyway.

PR #940, opened 2018-11-23
Bootstrap

Marri:
Someday I wanna do cool things with this? Like have a News page with release notes and badge the updates like so: 
screenshot from 2018-11-24 00-44-07

But I rabbithole whenever I try adding actual UI things, so have JUST a "ability to use Bootstrap" PR. No rabbit holing. Bad Marri. Baaaad.


Marri:
For posterity: We are using 8 in production.

PR #947, opened 2018-11-26
A couple more Bootstrap tweaks

Marri:
So very, very many things affected by line heights and box sizings.


app/assets/stylesheets/bootstrap-ui.scss
```diff
@@ -1 +1,32 @@
...
+@import 'bootstrap/progress';
+@import 'bootstrap/media';
+@import 'bootstrap/list-group';
+@import 'bootstrap/close';
+@import 'bootstrap/modal';
+@import 'bootstrap/tooltip';
+@import 'bootstrap/popover';
+@import 'bootstrap/carousel';
+@import 'bootstrap/utilities';
+@import 'bootstrap/print';
```
Throne3d: Does this reduce the weird changes bootstrap makes?

app/assets/stylesheets/global.scss
```diff
@@ -3,11 +3,19 @@
 body, html { height: 100%; }
 body {
     margin: 0 0 0 0;
-    font-family: Helvetica, Verdana, sans-serif;
+    font-family: FreeSans;
```
Throne3d: This surprises me! Won't this change the site font? We might want to use this as a fallback, instead of making it the primary, since people had quite strong opinions on fonts in the past.

Marri: I would guess this was a relic of running the site diff images, I certainly wouldn't plan to merge it in - @teceler for input?

app/assets/stylesheets/bootstrap-ui.scss
Marri: @teceler did we get the changeset small enough that we can put this back to @bootstrap instead?

app/assets/stylesheets/global.scss
teceler: whoops that got accidentally committed, yeah

app/assets/stylesheets/bootstrap-ui.scss
teceler: I feel like it's much cleaner to not pull in reboot and code and type, which do the things we'd have to override and have Strong Feelings About Things, but I can be swayed possibly.

Marri: As long as we manually check new PRs with bootstrap components until this is fixed, this'll do for an MVP

PR #948, opened 2018-11-27
Automatically convert Dreamwidth and Imgur URLs to HTTPS

Marri:
We know Dreamwidth and Imgur support HTTPS cleanly; we should force HTTPS at least for icons from those sites, to reduce mixed content issues.


db/migrate/20181127010456_add_https_to_icons.rb
```diff
@@ -0,0 +1,16 @@
+class AddHttpsToIcons < ActiveRecord::Migration[5.2]
+  def up
+    Icon.where("url LIKE ''").where("url LIKE ''").find_each do |icon|
```
Throne3d: Uh, did you forget to fill this line in?

db/migrate/20181127010456_add_https_to_icons.rb
```diff
@@ -0,0 +1,16 @@
...
+      icon.url = icon.url.gsub("http:", "https:")
+      icon.save
+    end
+
+    Icon.where("url like '%dreamwidth%'").where("url like 'http:%'").find_each do |icon|
```
Throne3d: Can we check against `dreamwidth.org` instead of just `dreamwidth`? Probably not an issue, but just to be consistent with the Ruby code.

db/migrate/20181127010456_add_https_to_icons.rb
Marri: ...I am mystified because I thought I copied the _second_ bit from the _first_ bit after copying all of it from my console and changing URLs but WHATEVER I'M FIXING IT

app/models/icon.rb
```diff
@@ -43,6 +44,13 @@ def use_icon_host
     self.url = ENV['ICON_HOST'] + url[(url.index(S3_DOMAIN).to_i + S3_DOMAIN.length)..-1]
   end
 
+  def use_https
+    return if uploaded?
+    return unless url.starts_with?('http://')
+    return unless url.include?("imgur.com") || url.include?("dreamwidth.org")
+    self.url = url.gsub('http:', 'https:')
```
Throne3d: (super super minor: the `g` in this is unnecessary)

db/migrate/20181127010456_add_https_to_icons.rb
```diff
@@ -0,0 +1,16 @@
+class AddHttpsToIcons < ActiveRecord::Migration[5.2]
+  def up
+    Icon.where("url LIKE 'imgur'").where("url LIKE 'http:'").find_each do |icon|
```
Throne3d: missing the end % symbol on `'http:'`?

db/migrate/20181127010456_add_https_to_icons.rb
```diff
@@ -0,0 +1,16 @@
...
+      icon.url = icon.url.gsub("http:", "https:")
+      icon.save
```
Throne3d: should we use bangs so this blows up on failure?

db/migrate/20181127010456_add_https_to_icons.rb
Throne3d: against imgur.com too?

PR #949, opened 2018-11-27
Copy output to clipboard

Marri:
Fancy JavaScript option to copy your output for the day to clipboard. Needs to be tested against HTML input probably?

### DIFF
--- a/app/views/users/output.haml
+++ b/app/views/users/output.haml
@@ -10,4 +10,3 @@
     %p= sanitize_written_content(r).html_safe
   - if @posts.empty? && @replies.empty?
     .centered — Nothing written today —
-= text_area_tag :_hiddenCopyText_, nil
--- a/app/views/users/output.haml
+++ b/app/views/users/output.haml
@@ -1,9 +1,13 @@
-.content-header Your Daily Ouput: #{@total} words
+.content-header
+  Your Daily Ouput: #{@total} words
+  = link_to '#' do
+    %button.float-right#select-output Copy All
 .sub= render partial: 'reports/date_nav'
-.even
+.even#copy-output
   - @posts.each do |post|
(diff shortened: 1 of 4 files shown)


app/views/users/output.haml
```diff
@@ -1,9 +1,13 @@
...
 .sub= render partial: 'reports/date_nav'
-.even
+.even#copy-output
   - @posts.each do |post|
     %p= sanitize_written_content(post).html_safe
   - @replies.each do |r|
     %p= sanitize_written_content(r).html_safe
   - if @posts.empty? && @replies.empty?
     .centered — Nothing written today —
+= text_area_tag :_hiddenCopyText_, nil
```
Throne3d: Is there a reason to put this in the HAML, as opposed to constructing it and appending it to the page in the JavaScript as necessary? I wonder because it seems more general (and more widely usable) if we generate it in the JavaScript, too, as it doesn't rely on so much pairing between server HTML and JavaScript.

Marri: It actually was originally in the JavaScript, but it took quite a bit more JavaScript to get it in and out than this one line of HAML did, so I went with this. I'm fine updating it back, but I'm also now having an exstiential crisis over the fact that people apparently copy/paste more than once in the same day so I dunno how useful this still is... I shall consult #constellation I guess?

Throne3d: Huh! I would expect the JavaScript to be along the lines of:

```js
var target = $("<textarea>");
target.attr('id', '_hiddenCopyText_');
$("body").append(target);
```

and then using `target` instead of `document.getElementById("_hiddenCopyText_");`. Then you can clean it up at the end with `target.delete();` or something.

(Should only be one more line, overall.)

PR #951, opened 2018-11-30
Best practices

Marri:
Ran https://github.com/flyerhzm/rails_best_practices, took a crack at some of the output. (Some of it was lies. We ignore the lies.)


spec/controllers/characters_controller_spec.rb
```diff
@@ -1174,16 +1174,15 @@
         [Template.count, Gallery.count, Icon.count, Reply.count, Post.count, Tag.count]
       }.and change { Character.count }.by(1).and change { CharactersGallery.count }.by(3).and change { CharacterTag.count }.by(2)
 
-      dup = assigns(:dup)
-      dup.reload
+      dupe = Character.last
```
Marri: dup is reserved in Ruby, so it's now dupe

app/views/indexes/show.haml
```diff
@@ -25,7 +25,7 @@
       %tr
         %td.odd.written-content{colspan: 6}= sanitize_written_content(@index.description).html_safe
   %tbody
-    = render partial: 'index_sections/single', collection:@index.index_sections.ordered, as: :section
+    = render partial: 'index_sections/single', collection: @index.index_sections.ordered, as: :section
```
Marri: Anything using collection or as arguments can't have the partial keyword removed or it assumes the arguments are locals not arguments for render. There's plenty of places I left alone, this one just showed up cause I added the missing space, so figured I'd call out the case.

PR #956, opened 2018-12-07
More bootstrap tweaks

Marri:
- checkboxes in the post list view are 3px too tall
- table spacers are 2px too short
- search box in search pages is 20px too slim
- headers in post/reply history pages are 20px too slim

### DIFF
--- a/app/assets/stylesheets/replies.scss
+++ b/app/assets/stylesheets/replies.scss
@@ -425,6 +425,11 @@ div.post-subheader { width: 100%; }
   text-align: center;
   width: 50px;
 }
+.post-check-box {
+  width: 30px;
+  padding-right: 10px;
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/replies.scss
```diff
@@ -425,6 +425,11 @@ div.post-subheader { width: 100%; }
   text-align: center;
   width: 50px;
 }
+.post-check-box {
+  width: 30px;
+  padding-right: 10px;
+  input.checkbox { height: 13px; margin: 0px; vertical-align: middle; } /* Keeps unread table from padding excessively for the checkboxes */
```
Throne3d: Does this really need so much specification? It might, it just surprises me.

Throne3d: Per #constellation-dev, it did need weird amounts of specification! Maybe this can be neatened up somehow, though.

PR #964, opened 2018-12-09
Filter messages from blocked users from in/outboxes

Marri:
If user A has blocked user B, user A should not see messages (including old messages) to or from user B in ther inbox or outbox. User B should continue to see messages as normal. This will handle blocks with threading.

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -4,12 +4,15 @@ class MessagesController < ApplicationController
   before_action :editor_setup, only: :new
 
   def index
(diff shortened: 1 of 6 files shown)


app/views/messages/index.haml
```diff
@@ -3,7 +3,7 @@
     %thead
       = render partial: @view.dup
     %tbody
-      - if @messages.empty?
+      - if @messages.reject { |m| m.blocked?(current_user) }.empty?
```
teceler: Maybe filter this on the model? Avoids having to do the each/next too. Or is that intended to save memory? And then you can also test whether it correctly filters, which seems useful

PR #985, opened 2019-01-05
Add a bin-docker/runner file so I can stop making more files

Marri:
Was using it as `runner rubocop` but presumably works just as well for e.g. `runner wraith`


Marri:
Historical note wraith will get its own container and should not use runner. Derp. Runner works with other things though!

PR #987, opened 2019-01-05
Make posts_from_relation a service

teceler:


teceler:
Closed b/c we've dramatically reworked posts_from_relation since I made this iirc

PR #995, opened 2019-01-10
Clean up daily report code

### LINKED ISSUE #979: Hide site testing from daily reports

teceler:
* Makes the daily report use posts_from_relation instead of It's Own Thing (incidentally fixes #979)
* Moves the daily report calculation to the controller
* Gives posts_from_relation an option to use max() on boards.name and users.username so the daily report call doesn't panic postgresql
* makes Report and DailyReport into services


app/models/daily_report.rb
```diff
@@ -17,15 +17,10 @@ def posts(sort='', page=1, per_page=25)
           then posts.created_at
           else coalesce(min(replies_today.created_at), posts.created_at)
           end as first_updated_at")
-      .joins(:board)
```
Marri: You're getting errors because you're including this join in the posts_from_relation call in the ApplicationController now, but the rspec test is just testing DailyReport.posts so it looks for this, fails to find it because we haven't called posts_from_relation, and crashes.

PR #996, opened 2019-01-11
Clean up rubocop file

teceler:
* Removes disabling of cops that register no offenses (it's possible some of these were disabled for being useless on our codebase, or while-they-are-fine-now-we-disagree-with-them, or similar, unfortunately I can't tell that from the results)
* Removes enabling of enabled-by-default cops
* Configures some cops to work with our coding style
* Cleans up a few cop complaints


.rubocop.yml
```diff
@@ -16,87 +16,53 @@ Rails:
...
   Enabled: false
 
 Layout/MultilineMethodCallIndentation:
-  Enabled: false
-
-Layout/MultilineMethodDefinitionBraceLayout:
-  Enabled: false
+  EnforcedStyle: 'indented'
+  Exclude:
+    - 'app/models/post.rb'
```
Throne3d: What's up with this exclusion?

.rubocop.yml
```diff
@@ -136,7 +96,8 @@ Lint/HandleExceptions:
     - 'spec/jobs/generate_flat_post_job_spec.rb'
 
 Lint/ParenthesesAsGroupedExpression:
-  Enabled: false
+  Exclude:
+    - 'spec/**/*'
```
Throne3d: How many issues do we get if we don't exclude this? I'm not sure where we'd be breaching it here, and we probably shouldn't breach it.

.rubocop.yml
```diff
@@ -233,35 +188,23 @@ Rails/Output:
...
 Style/AndOr:
-  Enabled: false
-
-Style/ArrayJoin:
-  Enabled: false
+  EnforcedStyle: 'conditionals'
 
 Style/AsciiComments:
-  Enabled: false
+  AllowedChars: ['–', '…']
```
Throne3d: I don't particularly like this cop – it's very anglocentric, the guideline behind it, and not very unicode friendly (as we see here). Can we keep it disabled?

app/services/post_scraper.rb
```diff
@@ -1,23 +1,23 @@
...
+    'erinflight'         => 'ErinFlight',
+    'andaisq'            => 'andaisq',
+    'rockeye-stonetoe'   => 'Rockeye',
+    'rockeye_stonetoe'   => 'Rockeye',
     'maggie-of-the-owls' => 'MaggieoftheOwls',
     'maggie_of_the_owls' => 'MaggieoftheOwls', # have both - and _ versions cause Dreamwidth supports both
-    'nemoconsequentiae' => 'Nemo',
-    'armokgob' => 'Armok',
+    'nemoconsequentiae'  => 'Nemo',
+    'armokgob'           => 'Armok',
```
Throne3d: This style might make it a bit of a nightmare to maintain. Are we able to disable it for certain areas? Or maybe completely, actually, I'm not sure we care all that much about keeping `=>` aligned like this.

.rubocop.yml
teceler: It had Opinions about the visible_to scope and I disagreed with them

.rubocop.yml
teceler: I think it was mis-firing (at least I _think_ that was mis-firing) on some expectation stuff. So... a lot.

.rubocop.yml
teceler: Yeah, that's fine. The main useful thing I would expect it to do is 'prevent files getting horribly corrupted if they're loaded in another encoding' but that of course requires you're only using ascii in your _code_ too...

app/services/post_scraper.rb
teceler: it's _so much easier to parse_ tho, at least to me. (And the cop does auto-correction so I don't think it'd be hard to maintain, but I do see your point)

Throne3d: Hmmm, okay. I don't mind too much.

.rubocop.yml
Throne3d: What were the opinions? This is the only thing I'm confused about, I think.

teceler:
```
app/models/post.rb:83:7: C: Layout/MultilineMethodCallIndentation: Use 2 (not 0) spaces for indenting an expression spanning multiple lines.
      .or(where(privacy: Concealable::REGISTERED))
      ^^^
app/models/post.rb:84:7: C: Layout/MultilineMethodCallIndentation: Use 2 (not 0) spaces for indenting an expression spanning multiple lines.
      .or(where(privacy: Concealable::ACCESS_LIST, user_id: user.id))
      ^^^
app/models/post.rb:85:7: C: Layout/MultilineMethodCallIndentation: Use 2 (not 0) spaces for indenting an expression spanning multiple lines.
      .or(where(privacy: Concealable::ACCESS_LIST, id: PostViewer.where(user_id: user.id).select(:post_id)))
      ^^^
app/models/post.rb:86:7: C: Layout/MultilineMethodCallIndentation: Use 2 (not 0) spaces for indenting an expression spanning multiple lines.
      .or(where(privacy: Concealable::PRIVATE, user_id: user.id))
      ^^^
```

Throne3d: I think I agree with it here that these should be indented. After merging #1111, can we do this?

teceler: FINE :P

PR #999, opened 2019-01-13
Turn Icon#delete_multiple into a service

teceler:


teceler:
Coverage failure is on specifically https://github.com/glowfic-constellation/glowfic/blob/7872e0079e108740c71013d0adac5156793e6823/app/controllers/icons_controller.rb#L147-L148 for reasons I do not understand

PR #1000, opened 2019-01-14
Improve post tests

teceler:
* Adds editable_by? and deletable_by? to post model
* Checks post is saved when testing if author can write in it (this is irrelevant atm but matters for the post-reply split)
* Checks a flat post is generated
* Uses timecop on some post#mark_read tests
* Marks post read on some reply tests
* Adds post content to post#new and post#edit feature specs
* Tests fields are preserved on failed post#create and post#update
* Feature specs preview on post#new
* Fixes a minor niggle on template model tests

### DIFF
--- a/spec/controllers/replies_controller_spec.rb
+++ b/spec/controllers/replies_controller_spec.rb
@@ -275,6 +275,7 @@ RSpec.describe RepliesController do
       other_user = create(:user)
       login_as(user)
       reply_post = create(:post, user: other_user, tagging_authors: [user, other_user], authors_locked: true)
+      reply_post.mark_read(user)
       post :create, params: { reply: {post_id: reply_post.id, content: 'test content!'} }
       expect(Reply.count).to eq(1)
     end
@@ -284,6 +285,7 @@ RSpec.describe RepliesController do
       other_user = create(:user)
       login_as(user)
       other_post = create(:post, user: user, tagging_authors: [user, other_user], authors_locked: true)
+      other_post.mark_read(user)
       post :create, params: { reply: {post_id: other_post.id, content: 'more test content!'} }
       expect(Reply.count).to eq(1)
     end
@@ -292,6 +294,7 @@ RSpec.describe RepliesController do
       user = create(:user)
       login_as(user)
       reply_post = create(:post)
+      reply_post.mark_read(user)
       Timecop.freeze(Time.zone.now) do
         post :create, params: { reply: {post_id: reply_post.id, content: 'test content!'} }
       end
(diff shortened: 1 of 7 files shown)


spec/controllers/replies_controller_spec.rb
```diff
@@ -275,6 +275,7 @@
       other_user = create(:user)
       login_as(user)
       reply_post = create(:post, user: other_user, tagging_authors: [user, other_user], authors_locked: true)
+      reply_post.mark_read(user)
```
Marri: What's the goal here? It makes me nervous that we need this side effect.

spec/models/template_spec.rb
```diff
@@ -53,7 +53,7 @@
       character1 = create(:character, template: template)
       character2 = create(:character, template: template, template_name: "nickname")
       character3 = create(:character, template: template, screenname: "screen_name")
-      expect(template.plucked_characters).to eq([[character1.id, character1.name], [character2.id, "#{character2.name} | #{character2.template_name}"], [character3.id, "#{character3.name} | #{character3.screenname}"]])
+      expect(template.plucked_characters).to match_array([[character1.id, character1.name], [character2.id, "#{character2.name} | #{character2.template_name}"], [character3.id, "#{character3.name} | #{character3.screenname}"]])
```
Marri: I think this is Most Correctly fixed by specifying alphabetical names for the characters and then expecting the ordering; the issue comes from letting the factories define the names and then having e.g. char100 come before char99.

spec/controllers/replies_controller_spec.rb
teceler: The post/reply split means the post-reply isn't marked read. For some reason. Possibly it _should_ be? But I think the current marking-read occurs b/c of the page being loaded after?
Anyway if this shouldn't be fixed elsewhere this can safely be merged before that, which was the critera I was pulling things out from that branch by. It can have it's own tho I suppose?

app/models/post.rb
```diff
@@ -278,6 +278,18 @@ def next_post
     Post.where(board_id: self.board_id, section_id: self.section_id).find_by(section_order: self.section_order + 1)
   end
 
+  def editable_by?(editor)
+    return false unless editor
+    return true if editor.id == user_id
+    editor.has_permission?(:edit_replies)
+  end
+
+  def deletable_by?(editor)
```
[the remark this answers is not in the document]
teceler: aha. yeah I must have done that after I stopped including writable in that PR, that makes sense as why that happened. I can pull this and the other thing into a branch for specifically changes for posts-to-replies that can be merged seperately? idk

PR #1007, opened 2019-01-21
Replies owed improvements

### LINKED ISSUE #704: Have posts with saved drafts show up in replies owed (even if last reply is user's)
[11:14 AM] Kappa✶: hmm, does this relate in any way to my dream of one day having posts where I have a draft saved show up in my replies owed?

### LINKED ISSUE #747: Keep self-threads / solo threads in Replies Owed
Related to #32, but probably easier to implement.

if there is only one author for a post it can be kept in replies owed until removed / completed?

Alternatively, a way to 'pin' posts in replies owed.

teceler:
* Adds a 'hiatused posts' view
* Shows posts with reply drafts on Replies Owed even if they would not show otherwise (fixes #704)
* Shows solo posts on Replies Owed (fixes #747)
* Improves code flow

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -20,13 +20,17 @@ class PostsController < WritableController
     @hide_quicklinks = true
     @page_title = 'Replies Owed'
 
+    ids = PostAuthor.where(user_id: current_user.id)
+
     if params[:view] == 'hidden'
-      ids = PostAuthor.where(user_id: current_user.id, can_owe: false).group(:post_id).pluck(:post_id)
-      @posts = posts_from_relation(Post.where(id: ids).ordered) and return
+      ids = ids.where(can_owe: false)
+    else
+      ids = ids.where(can_owe: true)
     end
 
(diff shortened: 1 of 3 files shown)


app/controllers/posts_controller.rb
```diff
@@ -20,14 +20,31 @@ def owed
...
+      ids = ids.where(can_owe: false)
+    else
+      ids = ids.where(can_owe: true)
+    end
+
+    ids = ids.group(:post_id).pluck(:post_id)
+    @posts = Post.where(id: ids)
+    unless params[:view] == 'hidden'
+      drafts = ReplyDraft.where(post_id: @posts.select(:id))
+      solo = PostAuthor.where(post_id: ids).group(:post_id).having('count(post_id) < 2').count.keys
```
Marri: Can we do pluck(:post_id) here since we're not using the select for anything else?

teceler: On the ReplyDraft one or? Not sure how to on the PostAuthor one

PR #1008, opened 2019-01-21
Clean up various rubocop lints

teceler:
* Switches to use of save! and update! in tests
* Changes to dir expand_path syntax in rails_helper
* Does some whitespace improvement
* Cleans up assignments of objects created in tests that were not used
* Fixes an unless blank to if present
* Fixes pluralization of hours and days
* Cleans up tag options (also fixed in Marri's new tags view branch)
* Disables Style/NumericLiterals so it stops pinging on the ToS version

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -317,6 +317,9 @@ Style/MultilineMemoization:
 Style/MutableConstant:
   Enabled: false
 
+Style/NumericLiterals:
+  Enabled: false
+
 Style/NumericPredicate:
   Enabled: false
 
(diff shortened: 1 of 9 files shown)


.rubocop.yml
```diff
@@ -317,6 +317,9 @@ Style/MultilineMemoization:
 Style/MutableConstant:
   Enabled: false
 
+Style/NumericLiterals:
+  Enabled: false
```
Marri: Note for future: reasonably sure this is because we're using a numerical literal to represent a date for CURRENT_TOS_VERSION and 20181109 should not in fact be represented as 20_181_109 because it actually represents 2018_11_09.

PR #1015, opened 2019-01-23
Add an additional select2 selector

Marri:
Currently does not catch the text inputs where you enter text that will be used to dynamically populate a search dropdown, in this case specifically the Setting etc fields on Post, as reported by Moriwen. Naively tested on Starry Dark and Dark and is now much more legible.

EDIT: `.select2-container--default .select2-search--dropdown .select2-search__field` doesn't catch it because it looks like this might be `.select2-container--default .select2-search--inline .select2-search__field`


teceler:
If it's `.select2-container--default .select2-search--inline .select2-search__field` maybe use that selector? Tho this works so idk. The code looks fine, haven't tested it but I don't see any obvious way for this to go horribly wrong.

PR #1019, opened 2019-01-26
Move all color definitions to their own SCSS partial file

Marri:
Keeps colors separate from semantic variable definition which are (in theory) already separate from colorizing selectors. For future proper inheritance attempts.


teceler:
This should be safe

PR #1024, opened 2019-02-01
Filter for draft by the current user on Replies Owed

teceler:
Because, in fact, you should only see posts where _you_ have a draft on your replies owed, not where your coauthor does, oops.


Throne3d:
(Could we add some quick tests for this? If not, they can always come later – it would just be cool to have them.)

PR #1028, opened 2019-02-22
Tweaks

Marri:
Gets us from 65 haml-lint errors to 57 by:

- Changing #saveconf to .saveconf so we don't duplicate IDs
- Removing inline CSS for .alert divs and incidentally renaming them .user-alert to avoid Bootstrap conflict
- Removing inline CSS from the Accept Terms partial

### DIFF
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -325,6 +325,7 @@ td, .table-list li { font-size: $font_size_small; }
   .sub { width: 90px; }
   .user-alert { margin: 7px 0 2px 0; }
   .msg { vertical-align: middle; }
+  #terms td { padding: 15px; }
 }
 .continuity-spacer { height: 20px; }
 .post-ignored { opacity: 0.3; }
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -321,7 +321,11 @@ td, .table-list li { font-size: $font_size_small; }
 
 /* Not yet deserving of its own CSS file */
 #stats ul { margin-top: 0; margin-bottom: 0; padding-left: 15px; }
-#signup .sub { width: 90px; }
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/layout.scss
```diff
@@ -321,7 +321,12 @@ td, .table-list li { font-size: $font_size_small; }
 
 /* Not yet deserving of its own CSS file */
 #stats ul { margin-top: 0; margin-bottom: 0; padding-left: 15px; }
-#signup .sub { width: 90px; }
+#signup {
+  .sub { width: 90px; }
+  .user-alert { margin: 7px 0 2px 0; }
+  .msg { vertical-align: middle; }
+  #terms td { padding: 15px; }
```
Marri: @teceler this thing

app/assets/stylesheets/layout.scss
```diff
@@ -396,6 +399,7 @@ The copy box can't be invisible/hidden or copy fails but opacity works
   height: 20px;
 }
 #_hiddenCopyText_ { height: 0px; width: 0px; opacity: 0; padding: 0px; margin: 0px; }
+
```
Marri: The two styles above this new line break are the ones that the comment is referring to about:
```
Styles for the user output page to handle copy/pasting text
The copy box can't be invisible/hidden or copy fails but opacity works
```
Since the subsequent styles have nothing to do with the user output page I wanted them separated by a break

app/assets/stylesheets/layout.scss
```diff
@@ -368,9 +373,7 @@ td, .table-list li { font-size: $font_size_small; }
   }
 }
 
-.user-moiety {
-  font-size: 80%;
-}
+.user-moiety { font-size: 80%; }
```
Marri: ...while I was in here?

PR #1033, opened 2019-04-01
News

Marri:
- [x] News#create
- [x] News#edit
- [x] News#update
- [x] Maybe a News#show redirect?
- [x] Fix db/structure.sql
- [x] Finish tests (what if your view is a deleted news post?)
- [ ] Consolidate views like post/message/news?
- [x] Rebase

### DIFF
--- a/app/models/news.rb
+++ b/app/models/news.rb
@@ -33,6 +33,6 @@ class News < ApplicationRecord
     view = NewsView.find_by(user_id: user.id)
     return News.count unless view
 
-    News.where('id > ?', view.news_id).order(id: :asc).count
+    News.where('id > ?', view.news_id).count
   end
 end
--- a/app/models/news.rb
+++ b/app/models/news.rb
@@ -1,0 +1,38 @@
+class News < ApplicationRecord
+  belongs_to :user, inverse_of: :news, optional: false
+
+  validates :content, presence: true
+
+  def editable_by?(user)
+    return false unless user
+    return true if user.id == user_id
+    user.has_permission?(:edit_news)
+  end
+
+  def deletable_by?(user)
+    return false unless user
+    return true if user.id == user_id
+    user.has_permission?(:delete_news)
+  end
+
+  def mark_read(user)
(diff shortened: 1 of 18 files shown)


app/models/news.rb
```diff
@@ -0,0 +1,38 @@
...
+    view.update(news: self)
+  end
+
+  def self.num_unread_for(user)
+    return 0 unless user
+
+    view = NewsView.find_by(user_id: user.id)
+    return News.count unless view
+
+    News.where('id > ?', view.news_id).order(id: :asc).count
```
Throne3d: Why bother ordering here?

app/views/layouts/application.haml
```diff
@@ -148,6 +148,12 @@
           = link_to 'Contact Us', contact_path
           &bull;
           = link_to 'Contribute', contribute_path
+          &bull;
+          - news_count = News.num_unread_for(current_user)
+          = link_to news_index_path(page: (news_count > 1 ? news_count : nil)) do
```
Throne3d: Should we store this away in a News method? News.latest_path_for(user) or something?

spec/controllers/news_controller_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+      expect {
+        post :create, params: {news: {content: 'staff made this'}}
+      }.to change { News.count }.by(1)
+      expect(response).to redirect_to(news_index_url)
+      expect(News.last.content).to eq('staff made this')
+    end
+  end
+
+  describe "GET show" do
+    it "errors without news post" do
```
Throne3d: I think this is in the wrong section (PATCH update)

app/models/news.rb
Marri: tl;dr because it used to not have a count on it, I can take it out

spec/controllers/news_controller_spec.rb
Marri: I think it's a bad copy paste not the wrong section but yes

PR #1041, opened 2019-05-05
Record editor mode on replies and posts

### LINKED ISSUE #764: Mark in the database what editor mode was used when editing something
Currently, it guesses heuristically, but this should be fixed to save it properly so we don't guess incorrectly or such.

teceler:
Fixes #764.

~~This does *not* currently record the editor mode for posts, because that *hopefully* should be irrelevant soon but means it weakly requires #705~~ This now handles replies, replydrafts, and posts


app/views/replies/_write.haml
```diff
@@ -11,8 +12,9 @@
     = f.hidden_field :character_id
     = f.hidden_field :icon_id
     = f.hidden_field :character_alias_id
+    = f.hidden_field :editor_mode, value: editor
     = hidden_field_tag :allow_dupe, @allow_dupe
-    = hidden_field_tag :editor_mode, params[:editor_mode] || current_user.default_editor
+    = hidden_field_tag :editor_mode, editor
```
teceler, line 17: hey past me why does that have both a hidden field tag and a form hidden_field

teceler:
Note: the test failures are happening because the feature spec is loading it up without js but default-editor set to rtf, and then entering text in the html field (because no js) but it gets parsed as rtf (because that's what's set).

app/helpers/writable_helper.rb
```diff
@@ -68,8 +68,8 @@ def simple_format_largebreak(text, options={})
   BR = '<br>'.freeze
 
   # specific blockquote handling is due to simple_format wanting to wrap a blockquote in a paragraph
-  def sanitize_written_content(content)
-    unless content[P_TAG] || content[BR_TAG]
+  def sanitize_written_content(content, editor_mode='html')
+    unless editor_mode == 'rtf' || content[P_TAG] || content[BR_TAG]
```
Throne3d, line 72: My brain is hurting with this, but shouldn't it be

```suggestion
    unless editor_mode == 'html' || content[P_TAG] || content[BR_TAG]
```
? The situations where the content contains P_TAG or BR_TAG are exactly the situations when we should declare editor_mode = html, I believe

app/views/replies/_write.haml
Throne3d: A good question - I guess we can remove this second one?

app/helpers/writable_helper.rb
Throne3d: Waaaait, no, HTML is the fake HTML that doesn't contain actual markup, I seeeeeeee

Okay this makes sense, I was just getting confused by the editor mode meaning

spec/controllers/replies/create_spec.rb
```diff
@@ -397,7 +499,15 @@
     login_as(user)
     reply_post = create(:post, authors_locked: true)
     reply_post.mark_read(user)
-    post :create, params: { reply: { post_id: reply_post.id, content: 'test' } }
+
+    post :create, params: {
+      reply: {
+        post_id: reply_post.id,
+        content: 'test',
+        editor_mdoe: 'html',
```
Throne3d, line 507:
```suggestion
        editor_mode: 'html',
```

app/controllers/posts_controller.rb
```diff
@@ -478,6 +478,7 @@ def permitted_params(include_associations=true)
       :authors_locked,
       :audit_comment,
       :private_note,
+      :editor_mode,
```
Throne3d, line 481: Do we use the root params[:editor_mode] anywhere, or just params[:post][:editor_mode]? That's probably related to why we have the two hidden fields below, so maybe grepping for editor_mode would be useful

app/views/posts/_write.haml
```diff
@@ -22,6 +23,7 @@
         = f.hidden_field :character_id, id: 'reply_character_id'
         = f.hidden_field :icon_id, id: 'reply_icon_id'
         = f.hidden_field :character_alias_id, id: 'reply_character_alias_id'
+        = f.hidden_field :editor_mode, value: editor_mode, id: 'editor_mode'
```
Throne3d, line 26: Should probably do this in the controller - `@post.editor_mode = current_user.default_editor` and then here just let it handle everything magically

script/add_editor_mode.rb
```diff
@@ -0,0 +1,15 @@
+def collect_written
+  replies = Reply.where(editor_mode: nil) + ReplyDraft.where(editor_mode: nil)
```
Throne3d, line 2:
```suggestion
  replies = Reply.where(editor_mode: nil) + ReplyDraft.where(editor_mode: nil) + Post.where(editor_mode: nil)
```

Throne3d (approved):

app/controllers/posts_controller.rb
```diff
@@ -328,6 +328,7 @@ def preview
     @post ||= Post.new(user: current_user)
     @post.assign_attributes(permitted_params(false))
     @post.board ||= Board.find_by_id(3)
+    @post.editor_mode ||= current_user.default_editor
```
teceler, line 331: Shouldn't this be set by permitted_params? Am I missing something there?

teceler: Oh it's ||=, I'm not sur ehow you'd wind up with it not set other than it the immediate move over but fair enough

spec/factories.rb
```diff
@@ -50,6 +50,7 @@
     user
     board
     description { "" }
+    editor_mode { 'rtf' }
```
teceler, line 53: Is this the right default for factories?

app/controllers/posts_controller.rb
Throne3d: Yeah! Just for the immediate move over, exactly - I think after that we should be fine to remove it here.

spec/factories.rb
Throne3d: I think so! I don't think it matters much either way, actually, since we usually set the content directly when we need it (and then this only affects the rendering of the writtens, not the models).

app/controllers/posts_controller.rb
Throne3d: Could remove it already tbh, it's not super important

```suggestion
```

PR #1047, opened 2019-05-15
Color-strip the Scss

teceler:
* Strips colors from layout files to _variables.scss
* Creates some new palette colors and substitutes them in
* Expands some hex codes to the long form
* Removes an unused selector


teceler:
Passes testing with wraith

PR #1051, opened 2019-05-21
Use ActiveStorage

teceler:


teceler:
Closing because it's probably easier to try this from scratch when we get to rails 6.1

PR #1052, opened 2019-05-23
Consistently use begin/rescue/else syntax on create/update/destroy

teceler:
* Selects begin/rescue/else syntax of one of the 3/4 different ones we were using because the ones that use transaction need it, and the services will need it
* Changes existing instances to use that syntax
* Expands Template, Gallery, and User failure error messages to include model errors
* Uses hash syntax on flash[:error] calls that have array and message components

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -344,6 +344,9 @@ Style/Proc:
 Style/RaiseArgs:
   Enabled: false
 
+Style/RedundantBegin:
+  Enabled: false
+
 Style/RedundantException:
   Enabled: false
 
--- a/app/controllers/characters_controller.rb
+++ b/app/controllers/characters_controller.rb
@@ -42,17 +42,20 @@ class CharactersController < ApplicationController
     @character.gallery_groups = process_tags(GalleryGroup, :character, :gallery_group_ids)
     build_template
 
-    if @character.save
+    begin
+      @character.save!
+    rescue ActiveRecord::RecordInvalid
+      @page_title = "New Character"
+      flash.now[:error] = {
+        message: "Your character could not be saved.",
+        array: @character.errors.full_messages
+      }
+      build_editor
+      render :new
+    else
       flash[:success] = "Character saved successfully."
-      redirect_to character_path(@character) and return
+      redirect_to character_path(@character)
     end
-
-    @page_title = "New Character"
-    flash.now[:error] = {}
-    flash.now[:error][:message] = "Your character could not be saved."
-    flash.now[:error][:array] = @character.errors.full_messages
-    build_editor
-    render :new
   end
 
   def show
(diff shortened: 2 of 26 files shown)


.rubocop.yml
```diff
@@ -344,6 +344,9 @@ Style/Proc:
 Style/RaiseArgs:
   Enabled: false
 
+Style/RedundantBegin:
+  Enabled: false
```
Throne3d: How come? I'd expect this to be good.

teceler: Because it does weird stuff to the indentation if you _don't_ use redundant begins and if you _do_, as I did here, this linter complains. If you really think I should do it the other way I can refactor for that tho I guess

Throne3d: I think I prefer the weird indentation thing, because it's Ruby-er – is that alright?

app/controllers/characters_controller.rb
```diff
@@ -83,35 +86,44 @@ def update
...
+          dupalias.save!
+        end
+        dupe.save!
       end
-      dupe.save!
+    rescue ActiveRecord::RecordInvalid
+      flash[:error] = {
+        message: "Character could not be duplicated.",
+        array: dup.errors.full_messages
+      }
```
Throne3d: I don't think `character#duplicate` has a view associated, and we don't redirect here, so this path should fail. I'm also not confident that `ActiveRecord::RecordInvalid` errors will get spit out of a transaction block. Could we leave this as it was – that way it'll send us an error email, if it does ever happen (and we can code something appropriately then)?

app/controllers/galleries_controller.rb
```diff
@@ -38,15 +38,20 @@ def create
...
 
-    unless @gallery.save
-      flash.now[:error] = "Your gallery could not be saved."
+    begin
+      @gallery.save!
+    rescue ActiveRecord::RecordInvalid
+      flash.now[:error] = {
+        message: "Your gallery could not be saved because of the following problems:",
+        array: @gallery.errors.full_messages
+      }
```
Throne3d: <3

app/controllers/characters_controller.rb
teceler: I could do that, or I could fake up a test, just a sec

app/controllers/indexes_controller.rb
```diff
@@ -48,27 +51,34 @@ def edit
...
-    flash[:success] = "Index deleted."
-    redirect_to indexes_path
-  rescue ActiveRecord::RecordNotDestroyed
-    flash[:error] = {}
-    flash[:error][:message] = "Index could not be deleted."
-    flash[:error][:array] = @index.errors.full_messages
-    redirect_to index_path(@index)
+    begin
+      @index.destroy!
+      redirect_to indexes_path
```
Throne3d: Should be in the else?

app/controllers/templates_controller.rb
```diff
@@ -13,14 +13,19 @@ def new
...
-      redirect_to template_path(@template)
-    else
-      flash.now[:error] = "Your template could not be saved."
+    begin
+      @template.save!
+    rescue ActiveRecord::RecordInvalid
+      flash.now[:error] = {
+        message: "Your template could not be saved because of the following problems:",
+        array: @template.errors.full_messages
+      }
```
Throne3d: <3

app/views/layouts/application.haml
```diff
@@ -124,8 +124,8 @@
             %ul
               - flash[:error][:array].each do |error|
                 %li= error
-        - flash.delete(:error)
       - flash.keys.each do |key|
+        - next if key == "error" && flash[key].is_a?(Hash)
```
Throne3d: What's up with this change here? Do we use `flash['error']` sometimes…?

spec/controllers/galleries_controller_spec.rb
```diff
@@ -71,7 +71,7 @@
         expect(response.status).to eq(200)
         expect(response).to render_template(:new)
         expect(assigns(:page_title)).to eq('New Gallery')
-        expect(flash[:error]).to eq('Your gallery could not be saved.')
+        expect(flash[:error][:message]).to eq('Your gallery could not be saved because of the following problems:')
```
Throne3d: Can we check `[:array]` here too?

spec/controllers/galleries_controller_spec.rb
```diff
@@ -83,7 +83,7 @@
       post :create, params: { gallery: {icon_ids: [icon.id]} }
       expect(response).to have_http_status(200)
       expect(assigns(:page_title)).to eq('New Gallery')
-      expect(flash[:error]).to eq('Your gallery could not be saved.')
+      expect(flash[:error][:message]).to eq('Your gallery could not be saved because of the following problems:')
```
Throne3d: Can we check `[:array]` here too?

spec/controllers/templates_controller_spec.rb
```diff
@@ -28,7 +28,7 @@
       login
       post :create
       expect(response).to render_template(:new)
-      expect(flash[:error]).to eq("Your template could not be saved.")
+      expect(flash[:error][:message]).to eq("Your template could not be saved because of the following problems:")
```
Throne3d: Can we check `[:array]` here too?

spec/controllers/templates_controller_spec.rb
```diff
@@ -149,7 +149,7 @@
       put :update, params: { id: template.id, template: {name: ''} }
       expect(assigns(:template)).not_to be_valid
       expect(response).to render_template(:edit)
-      expect(flash[:error]).to eq("Your template could not be saved.")
+      expect(flash[:error][:message]).to eq("Your template could not be saved because of the following problems:")
     end
 
     it "works" do
```
Throne3d: Can we check `[:array]` here too?

Throne3d: …wrong location, but, a couple lines above

Throne3d: ok just anywhere else we add `[:message]`

spec/controllers/users_controller_spec.rb
```diff
@@ -310,7 +310,7 @@
         user.update_columns(username: 'a') # too short to validate
         put :update, params: { id: user.id, tos_check: true }
         expect(user.reload.tos_version).to be_nil
-        expect(flash[:error]).to eq('There was an error saving your changes. Please try again.')
+        expect(flash[:error][:message]).to eq('There was an error saving your changes. Please try again.')
```
Throne3d: (so here too)

app/views/layouts/application.haml
Throne3d: Oh, it's to avoid breaking `flash[:error]` in tests with `render_views`! Neat!

teceler: That was breaking one of the galleries tests after I changed those to use a hash on failure!

app/controllers/characters_controller.rb
Throne3d: Looks good!

app/views/layouts/application.haml
Throne3d: Sorted out with #1103!

PR #1054, opened 2019-05-26
Prevent the code from trying to add users to board cameos multiple times

teceler:
(Also includes a test demonstrating the problem)

### DIFF
--- a/app/concerns/owable.rb
+++ b/app/concerns/owable.rb
@@ -48,7 +48,7 @@ module Owable
 
       # adjust for the fact that the associations are managed separately
       all_authors = authors + unjoined_authors + joined_authors + tagging_authors
-      new_cameos = all_authors.uniq - board.writers
+      new_cameos = all_authors.uniq - (board.writers + board.cameos)
       return if new_cameos.empty?
       board.cameos += new_cameos
     end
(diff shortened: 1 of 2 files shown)


app/concerns/owable.rb
```diff
@@ -48,7 +48,7 @@ def update_board_cameos
 
       # adjust for the fact that the associations are managed separately
       all_authors = authors + unjoined_authors + joined_authors + tagging_authors
-      new_cameos = all_authors.uniq - board.writers
+      new_cameos = all_authors.uniq - (board.writers + board.cameos)
```
Marri: Does board.authors include board.creator? Because if it does we can use that and then we don't need to do the addition.

teceler: board.authors seems to not exist, actually! it's just board.coauthors and board.cameos

Marri: sorry, I meant board.board_authors

teceler: Ah. Having poked at Effulgence in my dev env, the creator doesn't seem to have a BoardAuthor object.

Marri: Blah okay, we'll fix it if we ever get that sorted

PR #1056, opened 2019-05-26
Merge _galleryless.haml into _single.haml

### LINKED ISSUE #479: Fix galleries#_galleryless to look more like galleries#show
Currently it's quite obviously different, which is weird.

teceler:
Fixes #479

### DIFF
--- a/app/views/galleries/_single.haml
+++ b/app/views/galleries/_single.haml
@@ -35,7 +35,7 @@
   - if !gallery && @user.id == current_user.try(:id)
     %tr.gallery-subheader
       %th.sub Unsorted icons without a gallery will appear here. They can still be individually assigned to a character with no galleries.
-  - icons = gallery ? gallery.icons : @user.galleryless_icons
+  - icons ||= gallery ? gallery.icons : @user.galleryless_icons
   %tr.gallery-icons
     %td.icons-box
       .gallery{id: "gallery#{gallery ? gallery.id : 0}"}
--- a/app/views/galleries/_single.haml
+++ b/app/views/galleries/_single.haml
@@ -3,24 +3,26 @@
 %thead
   %tr.gallery-header{**attrs}
     %th.padding-10{class: (klass if defined? klass)}
-      = link_to gallery.name, gallery_path(gallery), class: 'gallery-title'
-      - if gallery.user_id == current_user.try(:id) && !defined? skip_forms
-        = link_to add_gallery_path(gallery), class: 'gallery-add' do
+      - link = gallery ? gallery_path(gallery) : user_gallery_path(id: 0, user_id: @user.id)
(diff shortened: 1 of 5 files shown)


app/views/galleries/_single.haml
```diff
@@ -2,23 +2,25 @@
...
+          .link-box.action-delete x Delete Gallery
     - unless local_assigns[:hide_minmax]
       .gallery-box.float-right.gallery-minmax{data: {id: gallery.id}} -
-    - if local_assigns[:character_gallery] && (gallery.user_id == current_user.try(:id) || current_user.try(:admin?))
+    - if local_assigns[:character_gallery] && (is_owner || current_user.try(:admin?))
       .float-right
         = image_tag "icons/arrow_up.png", class: "section-up disabled-arrow"
         = image_tag "icons/arrow_down.png", class: "section-down disabled-arrow"
-- groups = gallery.gallery_groups_data
+- groups = gallery.gallery_groups_data if gallery
```
Throne3d: I'm surprised this and the next line don't cause an error on galleryless

app/views/galleries/_single.haml
```diff
@@ -2,23 +2,25 @@
 - attrs = {class: 'section-ordered', data: {id: character_gallery.id, order: character_gallery.section_order}} if local_assigns[:character_gallery]
 %tr.gallery-header{**attrs}
   %th.padding-10{class: (klass if defined? klass)}
-    = link_to gallery.name, gallery_path(gallery), class: 'gallery-title'
-    - if gallery.user_id == current_user.try(:id) && !defined? skip_forms
-      = link_to add_gallery_path(gallery), class: 'gallery-add' do
+    - link = gallery ? gallery_path(gallery) : user_gallery_path(id: 0, user_id: @user.id)
+    = link_to gallery ? gallery.name : 'Galleryless icons', link, class: 'gallery-title'
+    - if is_owner && !skip_forms
```
Throne3d: Do we always define `skip_forms`? Might be worth using `local_assigns[:skip_forms]`

app/views/galleries/_single.haml
teceler: Did a quick test on my dev env in case the tests were missing something, but it seems to load galleryless fine
`present?` might do something Fancy?

app/views/galleries/_single.haml
teceler: I recall going through and making it so we _did_ define it everywhere this got called from so I could get rid of the `!defined`

app/views/galleries/_single.haml
Throne3d: Shouldn't! Oh well.

app/views/galleries/_single.haml
Throne3d: Ah, I see – cool!

PR #1059, opened 2019-06-04
[WIP] Merge color-setting into a colorize file

teceler:


teceler:
Pretty happy with this now, although I might still generate the full set of images for monochrome, river, and dark, and check against those

app/assets/stylesheets/layouts/default.scss
```diff
@@ -0,0 +1 @@
+@import 'colorize'
```
Throne3d: Missing semicolon?

Throne3d:
I haven't checked over this at all thoroughly, and also it has a lot of conflicts, but if you check over it with Wraith I'd be pretty happy with it!

app/assets/stylesheets/layouts/default.scss
teceler: I think we don't _technically_ need one since that's the only line in the file, but also, yes.

app/views/posts/flat.haml
```diff
@@ -7,6 +7,8 @@
     = stylesheet_link_tag 'application'
     - if current_user.try(:layout).present?
```
Marri: `= stylesheet_link_tag "layouts/#{current_user&.layout || 'default'}"` maybe?

teceler: Would that work? I guess that would work

Throne3d (commented):
Can we rebase this and check it against backstop?

PR #1060, opened 2019-06-04
Don't use HAML internal tag, use @tag

Marri:
@teceler


teceler:
Looks good (I think that one was my fault, oops)

PR #1067, opened 2019-06-20
Clean up prefixed linear-gradient functions

teceler:
All of our supported browsers have supported plain linear-gradient for a *while*, and we have a fallback anyway.


Throne3d:
I'm surprised this is the only place! I'm happy with it, though.

PR #1071, opened 2019-06-22
Avoid n+1 query in author_links

teceler:


app/helpers/application_helper.rb
```diff
@@ -230,8 +230,8 @@ def user_mem_link(user_id, username, deleted)
   end
 
   def author_links(post, linked: true, colored: false)
-    authors = post.authors.active.order(Arel.sql('lower(username) asc'))
-    num_deleted = post.authors.where(deleted: true).count
+    authors = post.authors.reject(&:deleted?).sort_by{|a| a.username.downcase}
+    num_deleted = post.authors.select(&:deleted?).size
```
Marri: we could maybe use post.authors.size - authors.size to avoid another select but I don't know that it matters *that* much?

teceler: oh, hmm, that would probably work. and slightly computationally easier, probably?

PR #1094, opened 2019-07-05
Various small css lints

teceler:
- [ ] Wraith run?

### DIFF
--- a/app/assets/stylesheets/_variables.scss
+++ b/app/assets/stylesheets/_variables.scss
@@ -37,8 +37,6 @@ $bg_color_flash_error: #FFDDDD !default; // error
 $font_color_flash_error: $colors_black !default;
 $bg_color_flash_success: #DDEEDD !default; // success
 $font_color_flash_success: $colors_black !default;
-$bg_color_flash_inbox: #DDDDEE !default; // new message
-$font_color_flash_inbox: $colors_black !default;
 $bg_color_flash_breadcrumbs: $colors_greybrown !default; // breadcrumbs ("Continuities » Effulgence » make a wish » one enormous bruise")
 $font_color_flash_breadcrumbs: $colors_black !default;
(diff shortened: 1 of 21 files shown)


app/assets/stylesheets/replies.scss
```diff
@@ -73,7 +74,6 @@
     position: relative;
     width: 130px;
     word-wrap: break-word;
-    display: inline-block;
```
Throne3d: Are we sure that this can be removed without changing anything?

app/assets/stylesheets/_variables.scss
```diff
@@ -37,8 +37,6 @@ $bg_color_flash_error: #FFDDDD !default; // error
 $font_color_flash_error: $colors_black !default;
 $bg_color_flash_success: #DDEEDD !default; // success
 $font_color_flash_success: $colors_black !default;
-$bg_color_flash_inbox: #DDDDEE !default; // new message
-$font_color_flash_inbox: $colors_black !default;
```
Throne3d: What's up with `.flash.inbox` being removed? (Here and everywhere else?)

teceler: We don't use it anymore? It was for the new message notifications we used to get in the flash format but now have the button thingy for

app/assets/stylesheets/replies.scss
teceler: Yes because -- if you expand it's also there five lines up.

app/assets/stylesheets/_variables.scss
Throne3d: OH. Right, yes, we don't use it any more, cool.

app/assets/stylesheets/replies.scss
Throne3d: Oh, cool! I didn't think it'd be duplicated.

teceler:
Wraith run looks good

PR #1096, opened 2019-07-13
Properly split tables with headers into %thead and %tbody

teceler:
(and put form-table-enders in %tfoot where possible)

### DIFF
--- a/app/views/users/show.haml
+++ b/app/views/users/show.haml
@@ -13,6 +13,7 @@
         %td.icons-box#user-avatar
           = link_to icon_path(@user.avatar) do
             = icon_tag @user.avatar
+  %tbody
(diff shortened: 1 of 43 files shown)


app/views/users/show.haml
```diff
@@ -7,12 +7,12 @@
           .user-moiety
             = color_block(@user)
             = @user.moiety_name
-  %tbody
```
Throne3d: I don't see the avatar as part of the header, really, though I grant it's not part of the regular body either. It's in a `%td` at the moment, though, so I'd err on putting it in the `%tbody`.

teceler: (put it in a seperate `%tbody`)

PR #1099, opened 2019-07-19
Fix invalid html

teceler:


app/views/users/new.haml
```diff
@@ -4,7 +4,7 @@
       %tr
         %th{colspan: 2} Sign Up
     %tbody
-      %tr#username
+      %tr#username-row
```
Marri: I am reasonably confident despite not having checked that this will break the user validation javascript

Throne3d: Can we call these like `#signup-username`, `#signup-password`, `#signup-secret` btw? That way we describe their uniqueness better. Just before we go changing the javascript.

teceler: Can do

app/views/posts/_list_item.haml
```diff
@@ -53,7 +53,7 @@
     by
     = user_mem_link(post.last_user_id, post.last_user_name, post.last_user_deleted?)
   - if local_assigns[:check_box_name]
-    %td.post-check-box{class: klass}= check_box_tag check_box_name, post.id, false, class: 'checkbox'
+    %td.post-check-box{class: klass}= check_box_tag check_box_name, post.id, false, class: 'checkbox', id: "#{check_box_name.delete('[]')}-#{post.id}"
```
Throne3d: Do we use this ID? It's a little weird that we're naming it, for example, `marked_ids_9`, if we don't actually need an ID, so maybe it should have `id: nil`?

app/views/messages/index.haml
```diff
@@ -28,7 +28,7 @@
               - else
                 = user_link(message.recipient)
             %td{class: klass}= pretty_time(message.last_in_thread.created_at)
-            %td{class: klass}= check_box_tag :"marked_ids[]", message.id, false, class: 'select-all-box'
+            %td{class: klass}= check_box_tag :"marked_ids[]", message.id, false, class: 'select-all-box', id: "marked_ids_#{message.id}"
```
Throne3d: Again wondering if we need IDs here

Throne3d: Applies to the other checkboxes too, actually.

app/views/password_resets/new.haml
```diff
@@ -4,7 +4,7 @@
       %th.centered{colspan: 2} Request Password Reset
     %tr
       %th.sub Username
-      %td.even= text_field_tag :username, params[:username], placeholder: "Username", class: 'text'
+      %td.even= text_field_tag :username, params[:username], placeholder: "Username", class: 'text', id: 'user_username'
     %tr
       %th.sub Email
       %td.odd= text_field_tag :email, params[:email], placeholder: "Email address", class: 'text'
```
Throne3d: For consistency, give this `id: 'user_email'`?

app/views/users/new.haml
Throne3d: Yep, user validation javascript will need tweaking: <https://github.com/Marri/glowfic/blob/9692cda6d8ecc3b53f1d7058d57d5cbc5714e812/app/assets/javascripts/users/new.js#L3>

Specifically, `#username`, `#email`, `#password`, and also most uses of `addAlertAfter`.

Throne3d:
Needs the javascript change before further review.

app/views/galleries/_single.haml
```diff
@@ -47,7 +47,7 @@
                 %br
                 %span.icon-keyword= keyword
               - if is_owner && !skip_forms
-                .select-button= check_box_tag :"marked_ids[]", icon_id
+                .select-button= check_box_tag :"marked_ids[]", icon_id, true, id: nil
```
Throne3d: This should default to false – currently defaults to making all icon check boxes ticked on gallery#show

Throne3d:
- gallery select all buttons still work on add_existing
- character replace post_id checkboxes still work
- messages select all button still works
- "no icon" displays fine
- marking icons in a gallery works fine (for deleting), but see comment
- submits `return_tag` when removing icons from a gallery in a tag#show?view=galleries expandable
- hides posts fine (`marked_ids` there)
- switching per_page works
- mobile pagination shows `&raquo;` fine (it did before, too, though, which was weird)
- username, email submission on password reset is fine
- javascript on signup form works, submitting sends params

PR #1100, opened 2019-07-19
Add HtmlProofer

teceler:
Depends on #1099 (to avoid tests and dev throwing errors about invalid html)


Throne3d:
Note that this may cause dev to throw errors if we look at pages we don't currently have automated tests rendering the HTML of.

PR #1102, opened 2019-07-19
Catch and handle InvalidAuthenticityToken errors

Marri:
I am so very, very tired of receiving these emails and I'm sure users are equally tired of losing the replies they're writing.

- Does not presently handle literally anything other than replies. Should probably someday be extended to handle posts, but I vote that waits on the post-to-replies stuff.

### DIFF
--- a/spec/controllers/posts_controller_spec.rb
+++ b/spec/controllers/posts_controller_spec.rb
@@ -1007,7 +1007,7 @@ RSpec.describe PostsController do
 
         login_as(user)
         expect(post).to be_taggable_by(user)
-        expect(post).to receive(:build_new_reply_for).with(user).and_call_original
+        expect(post).to receive(:build_new_reply_for).with(user, {}).and_call_original
         expect(controller).to receive(:setup_layout_gon).and_call_original
 
         get :show, params: { id: post.id }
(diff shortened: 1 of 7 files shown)


Throne3d:
Note: will have trouble with two tabs of replies being submitted at the same time if both are invalid. Shouldn't happen often. User can use "back" to get their reply back, as they always could before, so this should be a pretty clear improvement.

(Thank you :D Finally getting rid of these!)

spec/controllers/posts_controller_spec.rb
```diff
@@ -1037,7 +1037,7 @@
 
         login_as(user)
         expect(post).to be_taggable_by(user)
-        expect(post).to receive(:build_new_reply_for).with(user).and_call_original
+        expect(post).to receive(:build_new_reply_for).with(user, {}).and_call_original
```
teceler: I don't _think_ you need these, since you gave it a default value?

Marri: I do actually! I don't for generic calls to post.build_new_reply_for() because they can just use the default value, but this is checking what arguments the controller calls the method with, and the controller does in fact call it with both arguments.

teceler: Aha, okay

PR #1103, opened 2019-07-19
Unify error flashes with others

Throne3d:
Pick between formats by whether the value is a hash or not.

This is to make the special-casing of #1052 a little less ugly. Plus, it might be useful for notifications in the future (or condensing the "you have mail!" notifications a bit or something idk).


Throne3d:
Passes manual testing for:
- log in fail (`flash[:error]` with just a simple message)
- log in success (`flash[:success]` with just a simple message)
- content warnings (`flash[:error]` with a hash with an image)
- post failure (`flash[:error]` with a hash without image)

So I'm happy with it! Good to merge?

PR #1106, opened 2019-07-20
Create gallery list view

### LINKED ISSUE #56: Gallery stats
Icon stats but seen for all icons in a gallery.

    <Teceler> I wanted to see what icons in a gallery were used more without looking at each invidual icon yesterday
    <kappabeta> gallery stats would be so good

teceler:
Fixes #56

### DIFF
--- a/app/views/galleries/_list_section.haml
+++ b/app/views/galleries/_list_section.haml
@@ -1,0 +1,31 @@
+%tr.gallery{id: "gallery#{gallery ? gallery.id : 0}"}
+  %td
+    = form_tag delete_multiple_icons_path, method: :delete do
+      %table.gallery-table
+        %tr
+          - klass = cycle('even', 'odd')
+          %th.subber{class: klass} Keyword
+          %th.subber{class: klass} Credit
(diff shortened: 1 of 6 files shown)


app/views/galleries/_list_section.haml
```diff
@@ -0,0 +1,31 @@
...
+          %tr
+            %td{class: klass}
+              = link_to keyword, icon_path(id: icon_id)
+            %td{class: klass}= credit
+            %td{class: klass}= @times_used[icon_id]
+            %td{class: klass}= @posts_used[icon_id]
+
+            %td.width-70.right-align{class: klass}
+              - if is_owner && !skip_forms
+                = check_box_tag :"marked_ids[]", icon_id
```
Throne3d: Should this be wrapped in `.select-button`? Also should it have `, false, id: nil` on the end? `_icon_view.haml` seems to do that, and I'm not sure if they're necessary.

teceler: ...yes it should have the false id:nil. not sure about the .select-button

PR #1119, opened 2019-07-23
Move last_user, last_reply, and tagged_at logic into the post model

Throne3d:
The `post.set_last_caches` method should be useful for fixing messed up posts, this way (ones that get stuck in unread), especially when we're reordering the replies in a post or restoring a reply or such.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -33,7 +33,7 @@ class Post < ApplicationRecord
   has_many :indexes, inverse_of: :posts, through: :index_posts
   has_many :index_sections, inverse_of: :posts, through: :index_posts
 
-  attr_accessor :is_import
+  attr_accessor :is_import, :postpone_save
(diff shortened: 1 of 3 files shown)


Throne3d:
Looks like it might be taking a performance penalty:

Travis gave:
`Finished in 2 minutes 40.4 seconds (files took 5.72 seconds to load)` [master]
`Finished in 2 minutes 56.7 seconds (files took 5.86 seconds to load)` [fix/move-post-logic]

Marri got:
`Finished in 2 minutes 26 seconds (files took 4.94 seconds to load)` [master]
`Finished in 2 minutes 38.5 seconds (files took 5.23 seconds to load)` [fix/move-post-logic]

Not as big a penalty as I worried, but it's still 15 seconds on an almost-three-minute run, so it's pretty hefty.

PR #1120, opened 2019-07-24
Make methods with several arguments use keyword arguments

### LINKED ISSUE #895: Review functions for keyword arguments
@teceler fixed up the posts_from_relation method in #880, but we should do a full review to make sure we are using them where relevant more generally.

teceler:
Fixes #895

### DIFF
--- a/app/controllers/writable_controller.rb
+++ b/app/controllers/writable_controller.rb
@@ -95,7 +95,7 @@ class WritableController < ApplicationController
     @meta_canonical = post_url(@post, canon_params)
 
     # show <meta property="og:..." content="..."> – for embed data
-    @meta_og = og_data_for_post(@post, self.page, @replies.total_pages, per)
+    @meta_og = og_data_for_post(@post, page: self.page, total_pages: @replies.total_pages, per_page: per)
     @meta_og[:url] = @meta_canonical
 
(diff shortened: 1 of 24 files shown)


app/controllers/writable_controller.rb
```diff
@@ -155,7 +155,7 @@ def setup_layout_gon
     gon.no_icon_path = view_context.image_path('icons/no-icon.png')
   end
 
-  def og_data_for_post(post, page, total_pages, per_page)
+  def og_data_for_post(post, page: 1, total_pages:, per_page: 25)
```
Throne3d: Are we sure we want to allow defaults here?

spec/controllers/writable_controller_spec.rb
```diff
@@ -57,7 +57,7 @@
       user = create(:user, username: 'Tester')
       post = create(:post, subject: 'Temp', description: 'With an <a href="/characters/1">Alli</a>.', board: board, user: user)
 
-      data = controller.send(:og_data_for_post, post, 1, 5, 25)
+      data = controller.send(:og_data_for_post, post, {total_pages: 5})
```
Throne3d: Do these not work using keyword parameter syntax? I think they should…

PR #1122, opened 2019-07-24
Use dependent: :nullify and :destroy correctly

Marri:
Have Icon use dependent: :nullify instead of custom callbacks where possible. I have sanity checked and it does in fact use an update_all equivalent (which wouldn't work for posts or replies because we want audit history but is fine for avatars and drafts).

Also attach dependent: :destroy to has_many :through relationships, since otherwise callbacks will not be run when modifying the contents of the through association.

### DIFF
--- a/app/concerns/owable.rb
+++ b/app/concerns/owable.rb
@@ -3,19 +3,19 @@ module Owable
 
   included do
     has_many :post_authors, inverse_of: :post, dependent: :destroy
-    has_many :authors, class_name: 'User', through: :post_authors, source: :user
+    has_many :authors, class_name: 'User', through: :post_authors, source: :user, dependent: :destroy
 
     # list of users who owe replies on this post, whether or not they have posted yet
     has_many :tagging_post_authors, -> { where(can_owe: true) }, class_name: 'PostAuthor', inverse_of: :post
-    has_many :tagging_authors, class_name: 'User', through: :tagging_post_authors, source: :user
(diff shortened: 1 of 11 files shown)


app/concerns/owable.rb
```diff
@@ -3,19 +3,19 @@ module Owable
 
   included do
     has_many :post_authors, inverse_of: :post, dependent: :destroy
-    has_many :authors, class_name: 'User', through: :post_authors, source: :user
+    has_many :authors, class_name: 'User', through: :post_authors, source: :user, dependent: :destroy
 
     # list of users who owe replies on this post, whether or not they have posted yet
     has_many :tagging_post_authors, -> { where(can_owe: true) }, class_name: 'PostAuthor', inverse_of: :post
```
Throne3d: Does this also want `dependent: :destroy`, along with the other similar `_post_` things below?

Throne3d: There are other places that do this, too, and I get the feeling we want more `dependent: :destroy`s there (board coauthors)?

Marri: I think it does not matter here because it's scoped on the `has_many :post_authors`, which does; since this is one way (deleting a PostAuthor does not delete a Post, but having the unscoped declaration means deleting a Post will always delete all PostAuthors, therefore including scoped PostAuthors) we should be fine?

Throne3d: Makes sense! It would be super weird to do `.tagging_post_authors = [list, of, post_authors]` anyway, and that's where this comes into play.

PR #1133, opened 2019-07-29
Add more linters to our Code Climate checks

Marri:


teceler:
Honestly most of those complaints are 'don't use inline style attributes', which are Extremely Valid

PR #1134, opened 2019-07-30
Add to seeds

teceler:
- Reorders Boards
- Adds BoardSections
- Adds Posts and Replies
- Adds ContentWarnings
- Adds Audits
- Adds Messages
- Adds Teceler to seed users


db/seeds/post.rb
```diff
@@ -0,0 +1,622 @@
...
+    tagged_at: "2015-07-23 00:59:00",
+    last_reply_id: 164,
+    authors_locked: true
+  },
+])
+
+puts "Assigning users to threads..."
+
+ActiveRecord::Base.connection.execute("TRUNCATE TABLE post_authors RESTART IDENTITY")
+PostAuthor.create!([
```
Marri: Does creating replies after these exist muck with the timestamps etc at all?

db/seeds/post.rb
```diff
@@ -0,0 +1,622 @@
...
+  { post_id: 1, user_id: 5, read_at: "2018-12-06 00:05:13" },
+  { post_id: 1, user_id: 1, read_at: "2018-12-06 00:05:13" },
+  { post_id: 1, user_id: 3, read_at: "2019-05-06 23:07:00" },
+  { post_id: 2, user_id: 3, read_at: "2019-06-22 07:40:11" },
+  { post_id: 4, user_id: 3 },
+  { post_id: 3, user_id: 3, read_at: "2019-06-06 01:53:29" },
+  { post_id: 32, user_id: 3, read_at: "2015-07-23 00:57:00" },
+])
+
+FlatPost.regenerate_all
```
Marri: This depends heavily on background processing existing; not saying it's the wrong thing to do, but we might want it to be :bug: called out more obviously that this, unlike everything else, might be delayed in appearing in the SQL dataset.

db/seeds/post.rb
teceler: Not that I recall noticing when I repeatedly dumped them, which is weird, but it might be going with the replies created_at to check if they need updated

db/seeds/post.rb
teceler: I can put a puts there about queueing Flat Plot generation?

teceler: Does that work?

script/seed_dumper.rb
```diff
@@ -0,0 +1,121 @@
...
+MODELS.each do |model|
+  next if model.count == 0
+  puts "#{model.count} #{model.name.titleize.pluralize(model.count)}"
+  file = dump(model)
+  expand = true if [Character, Post, Reply, Audited::Audit].include?(model)
+  clean(file, expand)
+end
+
+FILES.each do |key, value|
+  next if value.count == 1
```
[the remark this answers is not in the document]
teceler: Easier to tell at a glance where things go if they're there? idk

PR #1138, opened 2019-08-02
Upgrade Audited

teceler:
* Updates audited from 4.8.0 to 4.9.0 ([Changelog](https://github.com/collectiveidea/audited/blob/master/CHANGELOG.md#490-2019-07-17))
* Applies old audited migration
* Pins Audited at minor version and adds a note to hopefully prevent this recurring


teceler:
New thing looks good!

PR #1142, opened 2019-08-08
Character API additions

Marri:
* You can now filter by template ID (with 0 for templateless)
* You can now provide additional JSON keys you wish included in the results
* Documentation updated to reflect changes

### DIFF
--- a/app/controllers/api/v1/characters_controller.rb
+++ b/app/controllers/api/v1/characters_controller.rb
@@ -3,6 +3,7 @@ class Api::V1::CharactersController < Api::ApiController
   before_action :find_character, except: [:index, :reorder]
   before_action :find_post, except: [:update, :reorder]
   before_action :find_template, only: :index
+  before_action :find_user, only: :index
   before_action :require_permission, only: :update
 
   resource_description do
@@ -13,14 +14,15 @@ class Api::V1::CharactersController < Api::ApiController
   param :q, String, required: false, desc: "If provided, will return only characters where q is present as a substring anywhere in any of a character's name, screenname or template nickname."
   param :page, :number, required: false, desc: 'Page in results (25 per page)'
   param :post_id, :number, required: false, desc: 'If provided, will return only characters that appear in the provided post'
+  param :user_id, :number, required: false, desc: 'If provided, will return only characters that belong to the provided user'
   param :template_id, :number, required: false, desc: 'If provided, will return only characters that belong to the provided template. Use 0 to find only characters with no template.'
-  param :includes, Array, in: ["default", "aliases", "template_name"], of: String, required: false, desc: 'Specify additional fields to return in JSON'
+  param :includes, Array, in: ["default_icon", "aliases", "template_name"], of: String, required: false, desc: 'Specify additional fields to return in JSON'
   error 403, "Post is not visible to the user"
   error 422, "Invalid parameters provided"
   def index
     queryset = Character.ordered
     queryset = queryset.with_name(params[:q].to_s) if params[:q].present?
-    queryset = queryset.where(user_id: current_user.id) if params[:user_id].present?
+    queryset = queryset.where(user_id: @user.id) if params[:user_id].present?
     queryset = queryset.where(template_id: @template&.id) if params[:template_id].present?
 
     if @post
@@ -40,7 +42,7 @@ class Api::V1::CharactersController < Api::ApiController
   error 404, "Character not found"
   error 422, "Invalid parameters provided"
   def show
-    render json: @character.as_json(include: [:galleries, :default, :aliases], post_for_alias: @post)
+    render json: @character.as_json(include: [:galleries, :default_icon, :aliases], post_for_alias: @post)
   end
 
   api :PATCH, '/characters/:id', 'Update a given character'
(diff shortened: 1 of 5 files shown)


app/controllers/api/v1/characters_controller.rb
```diff
@@ -2,27 +2,35 @@ class Api::V1::CharactersController < Api::ApiController
...
+  before_action :find_template, only: :index
   before_action :require_permission, only: :update
 
   resource_description do
     description 'Viewing and editing characters'
   end
 
   api :GET, '/characters', 'Load all the characters that match the given query, results ordered by name'
-  param :q, String, required: false, desc: "Query string"
+  param :q, String, required: false, desc: "If provided, will return only characters where q is present as a substring anywhere in any of a character's name, screenname or template nickname."
```
teceler: Improved description! Is good.

app/controllers/api/v1/characters_controller.rb
```diff
@@ -112,6 +120,12 @@ def find_post
     access_denied unless @post.visible_to?(current_user)
   end
 
+  def find_template
+    return unless params[:template_id].present?
+    return if params[:template_id] == '0' # used to filter for templateless characters
```
teceler: Shouldn't this do... something...?

Marri: Up in #index it checks if `params[:template_id].present?` (which it would be) and then filters on `template_id: @template&.id` which means it will filter on `template_id: nil` since we quit out of this method without setting `@template`

teceler: Looks good then!

teceler: (Might not be the _clearest_ code, but I assume making it clearer would also make it much more convoluted, so.)

Marri: I can give the comment a few more words without much drama I think :D

app/controllers/api/v1/characters_controller.rb
```diff
@@ -2,27 +2,35 @@ class Api::V1::CharactersController < Api::ApiController
...
   param :post_id, :number, required: false, desc: 'If provided, will return only characters that appear in the provided post'
+  param :template_id, :number, required: false, desc: 'If provided, will return only characters that belong to the provided template. Use 0 to find only characters with no template.'
+  param :includes, Array, in: ["default", "aliases", "template_name"], of: String, required: false, desc: 'Specify additional fields to return in JSON'
   error 403, "Post is not visible to the user"
   error 422, "Invalid parameters provided"
   def index
-    queryset = Character.with_name(params[:q].to_s).ordered
+    queryset = Character.ordered
+    queryset = queryset.with_name(params[:q].to_s) if params[:q].present?
+    queryset = queryset.where(user_id: current_user.id) if params[:user_id].present?
```
Throne3d: Do we need to specify `param :user_id` above? (Or is it inherited somehow? I haven't touched API code in forever.)

app/controllers/api/v1/characters_controller.rb
```diff
@@ -2,27 +2,35 @@ class Api::V1::CharactersController < Api::ApiController
...
     description 'Viewing and editing characters'
   end
 
   api :GET, '/characters', 'Load all the characters that match the given query, results ordered by name'
-  param :q, String, required: false, desc: "Query string"
+  param :q, String, required: false, desc: "If provided, will return only characters where q is present as a substring anywhere in any of a character's name, screenname or template nickname."
   param :page, :number, required: false, desc: 'Page in results (25 per page)'
   param :post_id, :number, required: false, desc: 'If provided, will return only characters that appear in the provided post'
+  param :template_id, :number, required: false, desc: 'If provided, will return only characters that belong to the provided template. Use 0 to find only characters with no template.'
+  param :includes, Array, in: ["default", "aliases", "template_name"], of: String, required: false, desc: 'Specify additional fields to return in JSON'
```
Throne3d: (It confuses me that `default` indicates `default_icon` but it looks like that's existing practice.)

spec/controllers/api/v1/characters_controller_spec.rb
```diff
@@ -46,6 +63,26 @@
...
+
+      it "filters by templateless" do
+        template = create(:template)
+        create(:character, template: template)
+        char = create(:character)
+        get :index, params: { template_id: '0' }
+        expect(response).to have_http_status(200)
+        expect(response.json).to have_key('results')
+        expect(response.json['results']).to contain_exactly(char.as_json(include: [:selector_name]).stringify_keys)
+      end
```
Throne3d: Should we also test specific output for a specially constructed example? Like, create an icon and character and template with specific names and then make sure the output is a specific thing? (If we have good tests for the character presenter, this might not be necessary, but I notice we didn't add a test for `template_name` producing correct results.)

Throne3d: Oh. I was confused by `template_name` and thinking it was literally the name of the template association. Never mind, since it's the template nickname and is definitely a property of character, I'm not concerned. (I am concerned by the fact we still have the confusing `character.template_name` in our code though XD)

app/controllers/api/v1/characters_controller.rb
Marri: Oops this looks like a relic of the charpicker, I'll update it yes.

app/controllers/api/v1/characters_controller.rb
Marri: We can probably rename it without issue tbh, I agree it's confusing, it was probably just a non-user-visible length thing

PR #1146, opened 2019-08-10
Test JS with selenium-webdriver and capybara

Throne3d:
Currently uses chromedriver. May switch to geckodriver. Hopefully Travis will work with it, and hopefully it won't be too slow!

For docker users, you will need to rebuild your images to successfully run tests here.


Throne3d:
Follow-up to #1145. Ensures no regression.

PR #1147, opened 2019-08-10
Test JS with selenium-webdriver and capybara (and geckodriver)

Throne3d:
Modification of #1146. Uses geckodriver and firefox. The Dockerfile uses firefox-esr because buster (the Debian release) doesn't have the other one directly, but I figure we should test on the latest so I stuck Travis at latest. `allowed_sites` is a bit ugly on this one – the download has some redirects.


Throne3d:
(I will probably use the chrome one if there's no compelling reason to use this one – the geckodriver seems to be using an unstable semver at present.)

Throne3d:
Closing in favor of #1146.

PR #1159, opened 2019-08-30
Avoid using exception flow where we don't need transactions

teceler:
Courtesy of the discussion about this


app/controllers/galleries_controller.rb
```diff
@@ -101,12 +99,10 @@ def edit
   def update
     @gallery.assign_attributes(gallery_params)
 
-    begin
-      Gallery.transaction do
-        @gallery.gallery_groups = process_tags(GalleryGroup, :gallery, :gallery_group_ids)
-        @gallery.save!
-      end
-    rescue ActiveRecord::RecordInvalid
+    if @gallery.update(gallery_groups: process_tags(GalleryGroup, :gallery, :gallery_group_ids))
```
Throne3d: This is harder to read like this, imo, since `.update(...)` looks like it's going to update _only_ the things you put there (not things you might have also changed earlier). I think it'll work, but I'd prefer to either do the explicit transaction here or else something like:

```rb
updates = gallery_params.merge(gallery_groups: process_tags(GalleryGroup, :gallery, :gallery_group_ids))
if @gallery.update(updates)
```

app/controllers/posts_controller.rb
```diff
@@ -187,23 +185,16 @@ def update
...
 
-    begin
-      Post.transaction do
-        @post.settings = settings
-        @post.content_warnings = warnings
-        @post.labels = labels
-        @post.save!
-      end
-    rescue ActiveRecord::RecordInvalid
+    if @post.update(settings: settings, content_warnings: warnings, labels: labels)
```
Throne3d: (ditto here & any other places)

PR #1163, opened 2019-09-01
Save memory on characters list view

teceler:
~~TODO: test the helper methods~~

### DIFF
--- a/app/helpers/character_helper.rb
+++ b/app/helpers/character_helper.rb
@@ -1,0 +1,14 @@
+module CharacterHelper
+  def settings_info(characters)
+    settings = characters.joins(:settings).group(:id)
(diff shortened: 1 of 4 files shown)


app/helpers/character_helper.rb
```diff
@@ -0,0 +1,14 @@
+module CharacterHelper
+  def settings_info(characters)
+    settings = characters.joins(:settings).group(:id)
```
amandaschloss: Do we know how the SQL is mananging to figure out which table's id to group on?

Marri: Do we know how the SQL is mananging to figure out which table's id to group on?

teceler: Afaict rails defaults to telling it to use the main table for these, but I can make it explicit if you'd prefer? It is kinda ambiguous xD

PR #1167, opened 2019-09-07
Use visible_to scope for post previous/next

teceler:
* Uses visible to scope to check post visibility on the query rather sequentially loading individual posts
* Removes restriction of checking a limited number of previous/next posts

### DIFF
--- a/app/controllers/writable_controller.rb
+++ b/app/controllers/writable_controller.rb
@@ -79,21 +79,10 @@ class WritableController < ApplicationController
     use_javascript('paginator')
 
     if @post.board.ordered?
-      next_post = @post.next_post
-      9.times do
-        break unless next_post
-        break if next_post.visible_to?(current_user)
-        next_post = next_post.next_post
-      end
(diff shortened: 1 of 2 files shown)


teceler:
> Looks good to me! Would you be able to confirm the SQL queries this generates, just in case? I think they'll be fine, just making sure it doesn't do anything crazily complicated.

```Post Load (0.7ms)  SELECT  "posts".* FROM "posts" WHERE "posts"."board_id" = $1 AND "posts"."section_id" IS NULL AND "posts"."privacy" = $2 AND (section_order > 0) ORDER BY "posts"."section_order" ASC LIMIT $3  [["board_id", 3], ["privacy", 0], ["LIMIT", 1]]```
```Post Load (0.7ms)  SELECT  "posts".* FROM "posts" WHERE "posts"."board_id" = $1 AND "posts"."section_id" IS NULL AND "posts"."privacy" = $2 AND (section_order < 0) ORDER BY "posts"."section_order" DESC LIMIT $3  [["board_id", 3], ["privacy", 0], ["LIMIT", 1]]```

PR #1169, opened 2019-09-09
Require PostScraper in ScrapePostJob tests

teceler:
`ScrapePostJob sends messages on username exceptions` fails when called alone or when the file is called by itself (rather than as part of the test suite running in general) without this.

### DIFF
--- a/spec/jobs/scrape_post_job_spec.rb
+++ b/spec/jobs/scrape_post_job_spec.rb
@@ -1,4 +1,5 @@
 require "spec_helper"
+require Rails.root.join("app", "services", "post_scraper.rb")
 
 RSpec.describe ScrapePostJob do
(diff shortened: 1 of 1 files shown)


Marri:
Are we sure something else isn't wrong? I thought rspec was supposed to include all the app/ files. Is something wrong with spec_helper or rails_helper?

teceler:
> Are we sure something else isn't wrong? I thought rspec was supposed to include all the app/ files. Is something wrong with spec_helper or rails_helper?

It does include it fine as long as this file isn't run specifically; Throne suggested it had something to do with how it loads things. _I_ suspect there's some weird behavior around defining multiple classes in the same file.

teceler:
Note for future us that it fails without this _specifically_ if That One Test is run first.

PR #1170, opened 2019-09-10
Limit line length

teceler:


app/controllers/application_controller.rb
```diff
@@ -150,9 +151,10 @@ def post_or_reply_link(reply)
 
   def posts_from_relation(relation, no_tests: true, with_pagination: true, select: '', max: false)
     if max
-      posts = relation.select('posts.*, max(boards.name) as board_name, max(users.username) as last_user_name, bool_or(users.deleted) as last_user_deleted'+ select)
+      posts = relation.select("posts.*, max(boards.name) as board_name, max(users.username) as last_user_name, \
```
Throne3d: Can we concatenate strings here, rather than escaping the newline? (I think this is going to add whitespace to the SQL query, if we ever look at it, and I find newline escaping kinda strange)

app/models/gallery.rb
```diff
@@ -29,7 +29,12 @@ class Gallery < ApplicationRecord
     # fetches an array of
     # galleries.map(&:gallery_groups).map{|group| [f1: group.id, f2: group.name]}
     # ordered by tag name
-    select("ARRAY(SELECT row_to_json(ROW(tags.id, tags.name)) FROM tags LEFT JOIN gallery_tags ON gallery_tags.tag_id = tags.id WHERE gallery_tags.gallery_id = galleries.id AND tags.type = 'GalleryGroup' ORDER BY gallery_tags.id ASC) AS gallery_groups_data_internal")
+    select("ARRAY(\
```
Throne3d: Could we use HEREDOCs here? Especially squiggly ones, which will remove additional indentation for us!

Something like such (though I'm not sure about the position of the final "SQL", maybe it looks better elsewhere, but it seems like we can't put the end paren right after it):

```rb
  select(<<~SQL
    ARRAY(
      SELECT row_to_json(ROW(tags.id, tags.name)) FROM tags
      LEFT JOIN gallery_tags ON gallery_tags.tag_id = tags.id
      WHERE gallery_tags.gallery_id = galleries.id AND tags.type = 'GalleryGroup'
      ORDER BY gallery_tags.id ASC
    ) AS gallery_groups_data_internal
    SQL
  )
```

This should work, I think (but maybe test it? I'm not 100% sure!).

Throne3d: There are some other `select`s that could do with this, too!

spec/jobs/scrape_post_job_spec.rb
```diff
@@ -30,7 +30,8 @@
     stub_fixture(url, 'scrape_no_replies')
     board = create(:board)
 
-    expect(ScrapePostJob).to receive(:notify_exception).with(an_instance_of(UnrecognizedUsernameError), url, board.id, nil, Post::STATUS_COMPLETE, false, board.creator_id).and_call_original
+    with = [an_instance_of(UnrecognizedUsernameError), url, board.id, nil, Post::STATUS_COMPLETE, false, board.creator_id]
+    expect(ScrapePostJob).to receive(:notify_exception).with(*with).and_call_original
```
Throne3d: Can we do this instead?

```rb
expect(ScrapePostJob).to receive(:notify_exception).with(
  an_instance_of(UnrecognizedUsernameError),
  url,
  board.id,
  nil,
  Post::STATUS_COMPLETE,
  false,
  board.creator_id,
).and_call_original
```

It's a bit long, but it "feels" better than splatting an array here, for me

app/models/gallery.rb
Throne3d: ❤️

spec/jobs/scrape_post_job_spec.rb
Throne3d: @teceler did you have opinions on this suggestion?

Throne3d:
After that one last comment, I'd be happy getting this merged!

spec/jobs/scrape_post_job_spec.rb
teceler: it's long and I don't really like it but I can do it if you super want it

Throne3d: We could put some of the other params on the same line, so long as other linters don't complain about that!

Maybe this?

```rb
expect(ScrapePostJob).to receive(:notify_exception).with(
  an_instance_of(UnrecognizedUsernameError),
  url, board.id, nil, Post::STATUS_COMPLETE, false, board.creator_id,
).and_call_original
```

The splatting of `with` seems really strange, so I'd prefer this.

teceler: Is that the only instance of that construction or do I need to hunt for others?

Throne3d: The only one, as far as I know!

PR #1172, opened 2019-09-16
Prevent hiatusing of threads from updating tagged_at

teceler:
Widely requested (along with making abandoned do the same, but that seemed like it might require community input). This will prevent marking threads hiatused bumping them on Recently Updated and Unread.


teceler:
Superseded by #1253

Issue #681, opened 2018-01-13
Make board.ordered a flag on a board


Throne3d:
Currently a board is ordered when it either has marked authors or it has board sections. This works for the cases "Effulgence" (authors & board sections, ordered), "Smol Continuities" (no authors, but board sections, so ordered), "Sandboxes" (no authors, no board sections, hence unordered), but falls down around, e.g., MWF or other delineated buckets like that (no authors, board sections, but *should*, I think, be unordered).

This could be made into a toggle/flag on the board model itself, instead of checked by heuristic.

Kinda related to #155.

PR #1178, opened 2019-10-05
Improve board author and open status handling

### LINKED ISSUE #1177: Don't iterate over all the boards looking for the ones the user can write in in the post editor
https://github.com/Marri/glowfic/blob/a349f55cdda738172456c31f824a604b0317839c/app/views/posts/_write.haml#L36

teceler:
Fixes #1177, semi-related to #681

### DIFF
--- a/app/controllers/boards_controller.rb
+++ b/app/controllers/boards_controller.rb
@@ -134,14 +134,14 @@ class BoardsController < ApplicationController
   private
 
   def set_available_cowriters
-    @authors = @cameos = User.active.ordered
+    @coauthors = @cameos = User.active.ordered
     if @board
-      @authors -= @board.cameos
-      @cameos -= @board.coauthors
-      @authors -= [@board.creator]
+      @coauthors -= @board.cameos
+      @cameos -= @board.writers
+      @coauthors -= [@board.creator]
       @cameos -= [@board.creator]
     else
-      @authors -= [current_user]
+      @coauthors -= [current_user]
       @cameos -= [current_user]
     end
     use_javascript('boards/editor')
@@ -163,7 +163,7 @@ class BoardsController < ApplicationController
 
   def og_data
     metadata = []
-    metadata << @board.writers.reject(&:deleted?).pluck(:username).sort_by(&:downcase).join(', ') if @board.authors_locked?
+    metadata << @board.writers.where.not(deleted: true).ordered.pluck(:username).join(', ') if @board.authors_locked?
     post_count = @board.posts.where(privacy: Concealable::PUBLIC).count
     stats = "#{post_count} " + "post".pluralize(post_count)
     section_count = @board.board_sections.count
--- a/app/controllers/boards_controller.rb
+++ b/app/controllers/boards_controller.rb
@@ -71,9 +71,7 @@ class BoardsController < ApplicationController
     @page_title = 'Edit Continuity: ' + @board.name
     use_javascript('boards/edit')
     @board_sections = @board.board_sections.ordered
-    unless @board.open_to_anyone? && @board_sections.empty?
-      @unsectioned_posts = @board.posts.where(section_id: nil).ordered_in_section
-    end
+    @unsectioned_posts = @board.posts.where(section_id: nil).ordered_in_section if @board.ordered?
   end
 
   def update
@@ -165,7 +163,7 @@ class BoardsController < ApplicationController
 
   def og_data
     metadata = []
-    metadata << @board.writers.reject(&:deleted?).pluck(:username).sort_by(&:downcase).join(', ') unless @board.open_to_anyone?
+    metadata << @board.writers.reject(&:deleted?).pluck(:username).sort_by(&:downcase).join(', ') if @board.authors_locked?
     post_count = @board.posts.where(privacy: Concealable::PUBLIC).count
     stats = "#{post_count} " + "post".pluralize(post_count)
     section_count = @board.board_sections.count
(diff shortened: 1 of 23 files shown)


app/models/board.rb
```diff
@@ -10,34 +10,27 @@ class Board < ApplicationRecord
...
-
-  def writer_ids
-    coauthor_ids + [creator_id]
-  end
-
   def open_to?(user)
     return false unless user
-    return true if open_to_anyone?
-    return true if creator_id == user.id
+    return true unless self.authors_locked?
```
Marri: I'd keep the creator check here; it's logically unnecessary but it does save on the db call.

app/models/board.rb
```diff
@@ -59,6 +52,11 @@ def move_posts_to_sandbox
     UpdateModelJob.perform_later(Post.to_s, {board_id: id}, {board_id: 3, section_id: nil})
   end
 
+  def add_creator_to_authors
+    return if coauthor_ids.include?(creator_id)
```
Marri: If this is an after_create how is this possible?

app/models/index.rb
```diff
@@ -10,7 +10,7 @@ class Index < ApplicationRecord
 
   def editable_by?(user)
     return false unless user
-    return true if open_to_anyone?
+    return false if authors_locked?
```
Marri: This seems like it's going to break a lot of things? I think you need return true unless, not return false if.

db/migrate/20191005002659_add_authors_locked_to_boards.rb
```diff
@@ -0,0 +1,18 @@
+class AddAuthorsLockedToBoards < ActiveRecord::Migration[5.2]
+  def up
+    add_column :boards, :authors_locked, :boolean, default: true
+    locked_boards = Board.order(id: :asc).joins(:board_authors).select('boards.id').distinct.pluck('boards.id')
```
Marri: Do we need anything from Board here? Or can we just pluck all distinct board_ids from BoardAuthor and just use that directly?

app/models/board.rb
teceler: ...I just copy-pasted this from the post authors after_create, which _does_ have a reason for that, but uh... theoretically if you did something screwy? I can remove it

app/models/index.rb
teceler: ...I think you're right, yes

db/migrate/20191005002659_add_authors_locked_to_boards.rb
teceler: Hm! that would probably work yes

app/helpers/application_helper.rb
```diff
@@ -249,9 +249,7 @@ def author_links(post, linked: true, colored: false)
   end
 
   def allowed_boards(obj, user)
-    valid_ids = BoardAuthor.where(user: user).pluck(:board_id)
-    valid_ids << obj.board_id if obj.board_id.present?
-    have_authors = BoardAuthor.select(:board_id).distinct.pluck(:board_id)
-    Board.where(id: valid_ids).or(Board.where(creator: current_user)).or(Board.where.not(id: have_authors)).ordered
+    authored_ids = BoardAuthor.where(user: user).pluck(:board_id)
```
Throne3d: I think this can use `.select(:board_id)` and then it'll end up being used as a subquery in the next query. This doesn't change much, though, so we can leave as is if you want!

teceler: Which way does the performance effect go on that?

Throne3d: Deferring to the database is ~always faster than doing things in our own code. It probably isn't a huge difference, though, because this will be a relatively tiny list, but loading it into our own code to generate another query just adds another round trip to the database.

teceler: So, like, if we were checking it more than one place, we might want to load it into code so the database doesn't do it five times, but under these circumstances...

Throne3d: Yeah, exactly!

app/controllers/boards_controller.rb
```diff
@@ -180,7 +179,13 @@ def og_data
     }
   end
 
+  def include_creator
+    return unless params.key?(:board) && params[:board].key?(:writer_ids)
+    return if params[:board][:writer_ids].include?(@board.creator_id)
+    params[:board][:writer_ids] << @board.creator_id
```
Marri: does this do anything funny if you submit an invalid update and it renders the page with the submitted params?

teceler: Nope, but also you inspired me to refactor to not need this hack

app/views/boards/_editor.haml
```diff
@@ -8,14 +8,19 @@
...
+      = f.collection_select(:coauthor_ids, @coauthors, :id, :username, {selected: @board.coauthors.pluck(:user_id)}, {multiple: true})
   %tr
     %th.sub Cameos
     %td.odd
       = f.collection_select(:cameo_ids, @cameos, :id, :username, {selected: @board.cameos.pluck(:user_id)}, {multiple: true})
+  %tr
+    %th.sub Locked?
+    %td.even
+      = f.check_box :authors_locked, class: 'vmid', checked: @board.authors_locked
+      = f.label :authors_locked, 'Only allow authors to create new posts in this board', class: 'no-margin'
```
Marri: continuity not board

teceler: ...whoops, :turkey:

PR #1179, opened 2019-10-05
Limit galleries#add to 150 images max

Marri:
Limit fairly arbitrarily chosen, can be tweaked later but should *exist*

### DIFF
--- a/app/assets/javascripts/galleries/uploader.js
+++ b/app/assets/javascripts/galleries/uploader.js
@@ -26,6 +26,8 @@ function randomString() {
 }
 
 function bindFileInput(fileInput, form, submitButton, formData) {
+  var limit = form.data('limit');
+  var numFiles = 0;
   var uploadArgs = {
     fileInput: fileInput,
     url: form.data('url'),
@@ -40,6 +42,17 @@ function bindFileInput(fileInput, form, submitButton, formData) {
     imageMaxHeight: 400,
 
     add: function(e, data) {
+      if (typeof limit !== 'undefined' && limit > 1) {
+        var isFirstFile = (fileInput[0].files[0] == data.files[0]);
+        var numUploading = fileInput[0].files.length;
+        if (isFirstFile) numFiles += numUploading;
+        if (numFiles > limit) {
+          if (isFirstFile) alert("You cannot upload more than "+limit+" files at once. Please try again.");
+          if(fileInput[0].files[numUploading - 1] == data.files[0]) numFiles -= numUploading;
+          return;
+        }
+      }
+
       var fileType = data.files[0].type;
       if (!fileType.startsWith('image/')) {
         alert("You must upload files with an image filetype such as .png or .jpg - please retry with a valid file.");
--- a/app/assets/javascripts/galleries/uploader.js
+++ b/app/assets/javascripts/galleries/uploader.js
@@ -2,6 +2,7 @@
 
 var uploadedIcons = {};
 var formKey = '';
+var numFiles = 0;
 
 $(document).ready(function() {
   var form = $('form.icon-upload');
@@ -27,7 +28,6 @@ function randomString() {
 
 function bindFileInput(fileInput, form, submitButton, formData) {
   var limit = form.data('limit');
-  var numFiles = 0;
   var uploadArgs = {
     fileInput: fileInput,
     url: form.data('url'),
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/galleries/uploader.js
```diff
@@ -40,6 +42,17 @@ function bindFileInput(fileInput, form, submitButton, formData) {
     imageMaxHeight: 400,
 
     add: function(e, data) {
+      if (typeof limit !== 'undefined' && limit > 1) {
```
Marri: Checks > 1 because of the limit is set to 1 then it's simply "don't allow multiple uploads but to allow switching them", as in icon#edit or galleries#edit. The existing maxNumberOfFiles argument does correctly turn this into a single file select so we can leave it be.

app/assets/javascripts/galleries/uploader.js
```diff
@@ -40,6 +42,17 @@ function bindFileInput(fileInput, form, submitButton, formData) {
...
+        var isFirstFile = (fileInput[0].files[0] == data.files[0]);
+        var numUploading = fileInput[0].files.length;
+        if (isFirstFile) numFiles += numUploading;
```
Marri: Don't want to let them upload limit *at a time*, we want to let them upload limit *total*.

app/assets/javascripts/galleries/uploader.js
```diff
@@ -40,6 +42,17 @@ function bindFileInput(fileInput, form, submitButton, formData) {
...
+        if (numFiles > limit) {
+          if (isFirstFile) alert("You cannot upload more than "+limit+" files at once. Please try again.");
+          if(fileInput[0].files[numUploading - 1] == data.files[0]) numFiles -= numUploading;
```
Marri: ...but once done complaining, we should recognize that they did not in fact upload any of these files and we should discount that fact. So if, say, limit is 2 and they've uploaded 1 and try 2 more, we should stop them but still allow them to upload 1 more.

app/assets/javascripts/galleries/uploader.js
```diff
@@ -40,6 +42,17 @@ function bindFileInput(fileInput, form, submitButton, formData) {
     imageMaxHeight: 400,
 
     add: function(e, data) {
+      if (typeof limit !== 'undefined' && limit > 1) {
+        var isFirstFile = (fileInput[0].files[0] == data.files[0]);
+        var numUploading = fileInput[0].files.length;
+        if (isFirstFile) numFiles += numUploading;
+        if (numFiles > limit) {
+          if (isFirstFile) alert("You cannot upload more than "+limit+" files at once. Please try again.");
```
Marri: Definitely do not show this alert per file

teceler: Can you do string interpolation here or not because ~javascript~

Throne3d: Can with template strings, but I think those are ES6 (so we don't yet use that in our codebase).

app/assets/javascripts/galleries/uploader.js
```diff
@@ -26,6 +26,8 @@ function randomString() {
 }
 
 function bindFileInput(fileInput, form, submitButton, formData) {
+  var limit = form.data('limit');
+  var numFiles = 0;
```
Throne3d: Should we be subtracting in a `remove` function or when we delete an input line or something? This is a sensible approximation but I can't see a subtraction and users may be confused if they upload a bunch, delete them as a mistake, then re-upload them.

app/assets/javascripts/galleries/uploader.js
Throne3d: Does this safely handle users uploading 5 and then 150 files? Looks like it'll add 5 files, so `fileInput[0].files` has a bunch in, and then when it runs with the next 150, isFirstFile will be false (`fileInput[0].files` is presumably just concatenated with the new 50, and so `fileInput[0].files[0] != newFile`, so we don't add more here, so we never trigger the `numFiles > limit` thingy)?

I'm thinking this might be simpler if we just add to `numFiles` each time we iterate this, and then we cut off once it gets to 150. Possibly less straightforward for the user, because we may not upload in order, but we still limit to 150 on the page. I think we might already be providing similar behavior, but I'm not sure.

Marri: It does in fact safely handle uploading 5 and then 150, in that the 5 will still have succeeded and then it will error on the second round. I tested a bunch with limit set to 3 and upload 2 and then 2 again (failed) and then 1 (worked), etc etc.

PR #1186, opened 2019-10-07
Improve characters#new feature specs

### LINKED ISSUE #1093: Test that invalid character preserves inputs
https://github.com/Marri/glowfic/blob/ad00696b4d2eeb0a3e19aec9fe6adf270a45fd79/spec/features/characters/new_spec.rb#L4-L36

teceler:
Fixes #1093 

~~I encountered some problems around icon selection and template selection b/c the js isn't getting loaded; fixing that may require #1146~~ (fixed)

### DIFF
--- a/spec/features/characters/new_spec.rb
+++ b/spec/features/characters/new_spec.rb
@@ -1,7 +1,7 @@
 require "spec_helper"
 
 RSpec.feature "Creating a new character", :type => :feature do
-  scenario "Create an invalid character" do
+  scenario "Create an invalid character", js: true do
     # view new character form without being logged in
     visit new_character_path
     expect(page).to have_selector('.flash.error')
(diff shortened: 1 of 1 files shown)


Throne3d:
Can you leave an explicit TODO in the codebase around the things we currently have commented out? That way we know what the comments are for directly.

spec/features/characters/new_spec.rb
```diff
@@ -50,7 +69,62 @@
...
+    expect(page).to have_no_selector('.flash.error')
+    expect(page).to have_selector('.flash.success')
+    within('.flash.success') do
+      expect(page).to have_text('Character saved successfully.')
+    end
+
+    within('.character-info-box') do
+      expect(page).to have_selector('.character-name', text: 'Example character')
+      expect(page).to have_selector('.character-screenname', text: 'example_screenname')
+      expect(page).to have_selector('.character-icon')
```
Throne3d: Can you check that this character-icon has the correct alt text? Just to be sure it used the clicked icon when producing the character.

spec/features/characters/new_spec.rb
```diff
@@ -50,7 +69,62 @@
     end
   end
 
-  scenario "Creating character with icon, description and extant template"
+  scenario "Creating character with icon, description and extant template", js: true do
+    user = login
+    icon = create(:icon, user: user, keyword: 'Example icon')
```
Throne3d: Like in the previous test, can we create another couple of random icons here to be sure it selects the correct one?

spec/features/characters/new_spec.rb
```diff
@@ -50,7 +69,62 @@
...
+    template = create(:template, user: user, name: 'Example template')
```
Throne3d: Can we make another template to be sure it selected this intentionally rather than randomly?

Throne3d:
I like these tests! My comments are just to be surer that this is testing correct behavior, instead of accidentally-correct behavior, but it's a significant improvement! :D

spec/features/characters/new_spec.rb
teceler: I think checking the id from the link should do that already, actually, but now it checks the alt text too I guess?? xD

PR #1199, opened 2019-10-14
Feature spec sessions#new and sessions#destroy

Throne3d:
~~Currently does destroy in a relatively ugly way: since we require the `DELETE` verb, and Rails hacks that in by using JavaScript, the button fails in our (JS-disabled) RackTest driver. Accordingly, I'm using `page.driver.submit` to directly force RackTest to submit `DELETE` to the logout route, rather than actually testing it in a more integrational way. There's a TODO to make sure we add that if/when #1146 is finished and merged.~~ Uses the new selenium-webdriver of #1146 to test logging out, since that requires the `DELETE` verb!

### DIFF
--- a/spec/features/sessions/destroy_spec.rb
+++ b/spec/features/sessions/destroy_spec.rb
@@ -1,0 +1,19 @@
+require "spec_helper"
+
+RSpec.feature "Logging out", :type => :feature do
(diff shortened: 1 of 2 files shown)


Throne3d:
Updated to use `click_button "Log out"` instead of `page.driver....`!

PR #1200, opened 2019-10-14
Use Node 10 in Docker

Throne3d:
Node 8 expires this December, per <https://github.com/nodejs/Release>. Since Node 10 is the active LTS, I suggest we move to that (supported until 2021!).


Marri:
Do we know what Heroku is using?

Throne3d:
node-v10.15.3-linux-x64, according to our last build! <https://github.com/heroku/heroku-buildpack-ruby/blob/master/CHANGELOG.md#v201-6232019> says that the buildpack uses this as of v201, so we should be good to track that page if we want to keep up to date on this.

PR #1206, opened 2019-10-17
Calculate what boards a post can be in via sql

### LINKED ISSUE #1177: Don't iterate over all the boards looking for the ones the user can write in in the post editor

teceler:
So I have a better fix for this in #1178, the sql solution with our current columns is a bit horrifying, but it does perform the important task of *not doing the thing we are at the moment*, and thus fixes #1177.

As idk when #1178 is going to be done given how it's going, I present this a hopefully temporary fix.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -247,4 +247,11 @@ module ApplicationHelper
     others = linked ? link_to("#{total-1} others", stats_post_path(post)) : "#{total-1} others"
     first_link + ' and ' + others
   end
+
+  def allowed_boards(obj, user)
+    valid_ids = BoardAuthor.where(user: user).pluck(:board_id)
+    valid_ids << obj.board_id if obj.board_id.present?
+    have_authors = BoardAuthor.select(:board_id).distinct.pluck(:board_id)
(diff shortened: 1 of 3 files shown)


app/helpers/writable_helper.rb
```diff
@@ -63,4 +63,11 @@ def shortened_desc(desc, id)
       content_tag(:span, sanitize_simple_link_text(desc[255..-1]), class: 'hidden', id: "desc-#{id}") +
       content_tag(:a, 'more &raquo;'.html_safe, href: '#', id: "expanddesc-#{id}", class: 'expanddesc')
   end
+
+  def allowed_boards(post, user)
+    valid_ids = BoardAuthor.where(user: user).pluck(:board_id)
+    valid_ids << post.board_id if post.board_id.present?
```
Marri: Is this valid? How would you get into this state and why is it permitted?

teceler: Mod edits of posts in boards they can't post in, iirc

PR #1208, opened 2019-10-20
Remove redundant box-sizing: border-box

teceler:


teceler:
> On the faith that these do in fact exist elsewhere, turtle

I mean I should _probably_ check we aren't setting content-box anywhere these are overriding, but I'm pretty sure they're just legacy from before we added the root statement

PR #1213, opened 2019-11-01
Use change(count) rather than change { count }

teceler:
Rubocop *probably* has a point that the latter is less clear?


teceler:
Actually This Doesn't Work, Thanks Rubocop

PR #1214, opened 2019-11-02
Optimize daily report to make less database calls

teceler:
It currently checks the reply count for the day and finds the earliest reply *per post*, which is sub-optimal when we can instead do this.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -140,13 +140,19 @@ class ApplicationController < ActionController::Base
 
   def post_or_reply_link(reply)
     return unless reply.id.present?
-    if reply.is_a?(Reply)
-      reply_path(reply, anchor: "reply-#{reply.id}")
-    else
-      post_path(reply)
-    end
+    post_or_reply_mem_link(id: reply.id, klass: reply.class)
   end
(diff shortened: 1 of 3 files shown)


teceler:
This does currently include a reply object for each post, so it might lose on memory, but I'm not positive how to fix that _and_ tell `post_or_reply_link` whether it's looking for a post or reply

app/controllers/application_controller.rb
```diff
@@ -140,13 +140,19 @@ def tos_skippable?
 
   def post_or_reply_link(reply)
     return unless reply.id.present?
-    if reply.is_a?(Reply)
-      reply_path(reply, anchor: "reply-#{reply.id}")
+    post_or_reply_mem_link({id: reply.id, class: reply.class})
+  end
+  helper_method :post_or_reply_link
+
+  def post_or_reply_mem_link(obj)
```
Throne3d: Tiny nit: we could use keyword parameters with defaults of `nil` instead of a hash, maybe? That way it's clearer that we're only getting `id` and `class`, if anything.

Throne3d: Might make the code a bit uglier in the partial, though, so it's up to you!

PR #1217, opened 2019-11-08
Avoid N+1 when calculating number of unread replies

teceler:


app/views/posts/_list_item.haml
```diff
@@ -44,8 +44,10 @@
   %td.width-70.post-replies.vtop{class: klass}= link_to replies_count, stats_post_path(post)
   - if logged_in? && local_assigns[:show_unread_count]
     %td.width-70.vtop{class: klass}
-      - if @opened_ids.include?(post.id) && unread_ids.include?(post.id)
-        = Reply.where(post_id: post.id).where('created_at > ?', post.last_read(current_user)).count
+      - if @unread_counts.key?(post.id)
```
Throne3d: It seems like this might cause strange behavior if we forget to add `with_unread: true` to `posts_from_relation` but `posts/list_item` expects an unread count. Should we fall back to the n-query version?

app/controllers/application_controller.rb
```diff
@@ -175,15 +175,20 @@ def posts_from_relation(relation, no_tests: true, with_pagination: true, select:
         post = posts.detect { |p| p.id == view.post_id }
         post && view.read_at < post.tagged_at
       end.map(&:post_id)
+
+      if with_unread
+        @unread_counts = Reply.where(post_id: @unread_ids).joins('INNER JOIN post_views ON replies.post_id = post_views.post_id')
+        @unread_counts = @unread_counts.where('replies.created_at > post_views.read_at').group(:post_id).count
+      end
```
Throne3d: Can we add a test for how `posts_from_relation` works with `with_unread: true`, maybe checking that it gets the right number for 2, 1, and 0 unread replies (and skips a post that's fully unread)?

Throne3d:
I'm not confident in the logic here, is part of why I'm wondering about the test & the fallback.

app/views/posts/_list_item.haml
teceler: Do we spec this? if we do it should raise an error and we should... notice

teceler: Yeah, it errors out if I take off `with_unread: true`, and we feature spec posts#unread and favorites#index. It's -- not _impossible_ something will get added with `show_unread_count` that fails to set that, but I would _expect_ us to check it renders properly :P

teceler: ...it raises an error with `.key?` but not with `.try`, tbc

PR #1223, opened 2019-11-12
Set up rubocop cops that duplicate codeclimate checks 

teceler:
To increase chances of detecting something it'll object to beforehand

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -117,7 +117,7 @@ Metrics/BlockNesting:
   Max: 4
 
 Metrics/ClassLength:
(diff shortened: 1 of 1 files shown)


teceler:
...[sigh] Apparently this for After We Clean Up Those Metrics

Throne3d:
Strange that it sometimes runs these but it's catching more when we enable them manually :thinking:

teceler:
Afaict it's running it's own versions, and so when we turn these on it duplicates the warnings rather than doing something sane.

PR #1224, opened 2019-11-13
Clean up PostScraper

teceler:


app/controllers/posts_controller.rb
```diff
@@ -370,11 +370,18 @@ def editor_setup
   end
 
   def import_thread
-    begin
-      importer = PostImporter.new(params[:dreamwidth_url])
-      importer.import(params[:board_id], current_user.id, section_id: params[:section_id], status: params[:status], threaded: params[:threaded])
-    rescue PostImportError => e
-      flash.now[:error] = e.api_error
+    importer = PostImporter.new(params[:dreamwidth_url])
```
Throne3d: Should this be `import_params[:dreamwidth_url]`? Or should we remove `:dreamwidth_url` from the allowed list below (since we're just skipping it, and I'm assuming we don't use it elsewhere)?

app/controllers/posts_controller.rb
```diff
@@ -370,11 +370,18 @@ def editor_setup
...
+    importer.import(import_params, user: current_user)
+    if importer.errors.present?
+      if importer.errors.key?(:username)
+        flash.now[:error] = {
+          message: importer.errors[:base].first,
+          array: importer.errors[:username].flatten
+        }
+      else
+        # if we don't have a username array there'll only be one error message
+        flash.now[:error] = importer.errors.full_messages.first
```
Throne3d: To be more general / future compatible, maybe we should use error message and array? Message something like "There was an error importing your thread!" and array just `importer.errors.full_messages`

teceler: Mostly I would have to work out how to _test_ that, but yes

app/services/post_importer.rb
```diff
@@ -1,55 +1,60 @@
...
-    validate_usernames!
+  def import(params, user:)
+    validate_url
+    return false if @errors.present?
+    validate_duplicate(params[:board_id]) unless params[:threaded]
+    return false if @errors.present?
+    validate_usernames
+    return false if @errors.present?
 
     # note that the arg order for this import method does not match the order of ScrapePostJob
```
Throne3d: Comment no longer true

app/services/post_scraper.rb
```diff
@@ -1,88 +1,59 @@
...
-        import_replies_from_doc(doc)
+      import_post_from_doc(html_doc)
+      if threads.present?
+        threads.each { |thread| evaluate_links(doc_from_url(thread)) }
+      else
+        evaluate_links(html_doc)
       end
       finalize_post_data
     end
+    return if @errors.present?
```
Throne3d: I'm concerned that we might not roll back transactions if an error occurs, any more – we'll be populating `@errors`, but may not raise anything to the Post.transaction block.

I'm not sure how to resolve this, but I think this transaction might require exception flow. Alternatively, maybe we could move this `return if @errors.present?` just into the transaction block – then we'll still be potentially do lots of work after encountering an error, but at least we'll ensure database consistency.

app/services/post_scraper.rb
```diff
@@ -1,88 +1,59 @@
...
   end
+  alias scrape! scrape
 
-  # works as an alternative to scrape! when you want to scrape particular top-level threads of a post sequentially
+  # works as an alternative to scrape when you want to scrape particular top-level threads of a post sequentially
   # "threads" are URL permalinks to the threads to scrape, which it will scrape in the given order
-  def scrape_threads!(threads)
-    raise RuntimeError.new('threaded_import must be true to use scrape_threads!') unless @threaded_import
+  def scrape_threads(threads)
+    @errors.add(:base, 'threaded_import must be true to use scrape_threads') && return unless @threaded_import
```
Throne3d: Does `@errors.add` definitely return a truthy value? If not, we won't return. Might be worthwhile turning this into a block instead.

app/services/post_scraper.rb
```diff
@@ -1,88 +1,59 @@
...
-        end
-      end
-      finalize_post_data
+  attr_writer :errors
+
+  def evaluate_links(base_doc)
+    import_replies_from_doc(base_doc)
+    links = page_links(base_doc)
+    links.each_with_index do |link, i|
+      Resque.logger.debug "Scraping '#{@post.subject}': page #{i+1}/#{links.count}"
```
Throne3d: What's up with all the `logger.debug` (etc) calls turning into `Resque.logger.debug`? Did something change?

app/services/post_scraper.rb
```diff
@@ -143,179 +109,80 @@ def threaded_page_links
...
   def import_post_from_doc(doc)
     subject = @subject || doc.at_css('.entry .entry-title').text.strip
-    logger.info "Importing thread '#{subject}'"
+    Resque.logger.info "Importing thread '#{subject}'"
 
     username = doc.at_css('.entry-poster b').inner_html
-    img_url = doc.at_css('.entry .userpic img').try(:attribute, 'src').try(:value)
-    img_keyword = doc.at_css('.entry .userpic img').try(:attribute, 'title').try(:value)
+    img_node = doc.at_css('.entry .userpic img')
+    img_url = img_node.try(:attribute, 'src').try(:value)
```
Throne3d: `img_node.try(:[], 'src')` would do the same thing, afaik – or maybe `img_node&.[]('src')`, but that's a lot uglier

app/services/post_scraper.rb
```diff
@@ -143,179 +109,80 @@ def threaded_page_links
...
+    if (@subject || !@threaded_import) && (subj_post = Post.find_by(subject: subject, board_id: @board_id))
+      @errors.add(:post, "was already imported! #{ScrapePostJob.view_post(subj_post.id)}")
+      raise ActiveRecord::Rollback
     end
 
-    set_from_username(@post, username)
-    @post.last_user_id = @post.user_id
-
-    set_from_icon(@post, img_url, img_keyword)
+    @post = Post.new(board_id: @board_id, section_id: @section_id, subject: subject, status: @status, is_import: true)
```
Throne3d: note for me to check: Do we set `created_at`, `updated_at`, `edited_at` in ReplyScraper?

app/services/post_scraper.rb
```diff
@@ -143,179 +109,80 @@ def threaded_page_links
...
-    tag.character = character
-    tag.user = character.user
+  def clean_url(url)
+    uri = URI(url)
+    query = Rack::Utils.parse_query(uri.query)
+    return url if check_query(query)
+    query.except!('style')
+    query['view'] = 'flat' unless @threaded_import
+    new_query = Rack::Utils.build_query(query)
+    URI::HTTPS.build(host: uri.host, path: uri.path, fragment: uri.fragment, query: new_query).to_s
```
Throne3d: Is there a need to specify everything here? I think we can just do `uri.query = query` and then `uri.to_s`.

app/services/post_scraper.rb
Throne3d: We do!

app/services/reply_scraper.rb
```diff
@@ -0,0 +1,148 @@
...
+    Audited.audit_class.as_user(@reply.user) do
+      @reply.save!
+    end
+  end
+
+  private
+
+  def post_setup
+    @reply.last_user_id = @reply.user_id
+    @reply.edited_at = @reply.created_at
```
Throne3d: Won't this be nil? The reply hasn't been persisted in the database yet, and Rails doesn't fill this in until it is.

app/services/post_scraper.rb
teceler: logger was a method on PostScraper. I can revert that if you want?

app/services/post_scraper.rb
teceler: Yes :P

app/services/reply_scraper.rb
teceler: ...but I set the created_at?

spec/services/post_scraper_spec.rb
```diff
@@ -10,13 +10,13 @@
     url = 'http://wild-pegasus-appeared.dreamwidth.org/403.html?view=flat'
     scraper = PostScraper.new(url)
     expect(scraper.url.sub('view=flat', '')).not_to include('view=flat')
```
Throne3d: nit (not your fault): this is more readable as `expect(scraper.url.scan(/view=flat/).length).to eq(1)`

spec/services/post_scraper_spec.rb
```diff
@@ -10,13 +10,13 @@
...
+    expect(scraper.url.length).to eq(url.length)
   end
 
-  it "should add site style to the url" do
-    url = 'http://wild-pegasus-appeared.dreamwidth.org/403.html?view=flat'
+  it "should remove non-site styles from the url" do
+    url = 'http://wild-pegasus-appeared.dreamwidth.org/403.html?style=mine&view=flat'
     scraper = PostScraper.new(url)
-    expect(scraper.url).to include('&style=site')
+    expect(scraper.url).not_to include('&style=mine')
```
Throne3d: I'd remove the `&` from the start of this string – `?style=mine`, as in the current string, would falsely pass this test

app/services/post_scraper.rb
```diff
@@ -143,179 +109,80 @@ def threaded_page_links
...
 
-  def prompt_for_user(username)
-    raise UnrecognizedUsernameError.new("Unrecognized username: #{username}") unless @console_import
-    print('User ID or username for ' + username + '? ')
-    input = STDIN.gets.chomp
-    return User.find_by_id(input) if input.to_s == input.to_i.to_s
-    User.where('lower(username) = ?', input.downcase).first
+  def check_query(query)
+    # query parameters are good if both:
+    # - style is absent or site
```
Throne3d: Style being absent isn't actually always fine: see, e.g., https://belltower.dreamwidth.org/8579.html, which has a different style from `?style=site`.

app/services/post_scraper.rb
Throne3d: Ah! It's probably fine.

teceler:
Oh heck. I forgot that some styles/communities would do that, yeah. I'll fix it

app/services/reply_scraper.rb
teceler: ...But I set created_at? I guess I didn't set user_id if that's what you mean?

Throne3d: Oh, no, I just totally missed the part where you set created_at. Should be fine!

app/services/post_scraper.rb
teceler: ...I looked and looked for a way to do that but I didn't think of this, thank you.

app/services/post_scraper.rb
teceler: Any point where errors are added we should also spawn a rollback. I can add another interrupt in here just in case?

Throne3d: Hmm. I think it'd be sensible to add it just in case, and so we can make changes in the future without having to remember this. Does that sound okay?

teceler: I'm not sure how to make it be tested

Throne3d: Try creating a post like normal, but then:

`allow(post_scraper).to receive(:finalize_post_data) { post_scraper.instance_variable_set(:@errors, ['an error!']) }`

And make sure it isn't persisted?

Throne3d: (Alternatively we could use getters and setters for the errors, and in this block here set an error using the setter. That way it'd be slightly less brittle, but also more work to implement now.)

app/services/post_importer.rb
```diff
@@ -3,23 +3,20 @@ def initialize(url)
...
   def self.valid_dreamwidth_url?(url)
     # this is simply checking for a properly formatted Dreamwidth URL
     # errors when actually querying the URL are handled by ScrapePostJob
-    return false if url.blank?
-    return false unless url.include?('dreamwidth')
+    return false if url.blank? || url.exclude?('dreamwidth')
     parsed_url = URI.parse(url)
-    return false unless parsed_url.host
-    parsed_url.host.ends_with?('dreamwidth.org')
+    parsed_url.host&.ends_with?('dreamwidth.org')
```
Throne3d, line 19:
```suggestion
    !!parsed_url.host&.ends_with?('dreamwidth.org')
```
might want this to ensure that `parsed_url.host` being nil returns false rather than nil (to match the `?` on the end of the method name)

app/services/post_scraper.rb
```diff
@@ -125,18 +97,13 @@ def threaded_page_links
...
-    @post.section_id = @section_id
-    @post.subject = subject
-    @post.status = @status
-    @post.is_import = true
+    @post = Post.new(board_id: @board_id, section_id: @section_id, subject: subject, status: @status, is_import: true)
 
     # detect already imported
     # skip if it's a threaded import, unless a subject was given manually
-    if (@subject || !@threaded_import) && (subj_post = Post.find_by(subject: @post.subject, board_id: @board_id))
+    if (subject || !@threaded_import) && (subj_post = Post.find_by(subject: subject, board_id: @board_id))
```
Throne3d, line 106: This first condition, `subject || !@threaded_import`, will always be true - either we were passed a manual subject or we fetched it from the page

Previously it was checking specifically that it was passed manually to do the disambiguation here, I guess so people can force a reimport for some reason? We should probably keep that, or if we're sure we don't need that, reformulate this condition to just always do the disambiguation

```suggestion
    if (@subject || !@threaded_import) && (subj_post = Post.find_by(subject: subject, board_id: @board_id))
```

app/services/reply_scraper.rb
```diff
@@ -28,116 +28,126 @@ def initialize(reply, console: false)
...
-    host_url = url.gsub(/https?:\/\//, "")
-    https_url = 'https://' + host_url
-    icon = Icon.find_by(url: https_url)
-    tag.icon = icon and return if icon
+    create_icon(url, keyword)
+  end
 
+  def parse_url(url)
+    uri = URI(url)
+    uri.host ||= 'v.dreamwidth.org'
```
Throne3d, line 103: Looks like we also want to set the scheme to `https` (rather than http / blank), not sure we're doing that anymore

spec/controllers/posts/create_spec.rb
```diff
@@ -44,31 +45,55 @@
...
 
-    it "requires valid dreamwidth url" do
-      login_as(user)
-      post :create, params: { button_import: true, dreamwidth_url: 'http://www.google.com' }
-      expect(response).to render_template(:new)
-      expect(flash[:error]).to eq("Invalid URL provided.")
-    end
+    context "with importer" do
+      let(:user) { create(:importing_user) }
+      let(:url) { 'http://wild-pegasus-appeared.dreamwidth.org/403.html?style=site&view=flat' }
```
Throne3d, line 50: It looks like these two `let`s are the same as in the parent `context "scrape"`?

```
    let(:user) { create(:importing_user) }
    let(:url) { 'http://wild-pegasus-appeared.dreamwidth.org/403.html?style=site&view=flat' }
...
      let(:user) { create(:importing_user) }
      let(:url) { 'http://wild-pegasus-appeared.dreamwidth.org/403.html?style=site&view=flat' }
```

spec/controllers/posts/create_spec.rb
```diff
@@ -44,31 +45,55 @@
...
+      it "queues job" do
+        create(:character, user: user, screenname: 'wild-pegasus-appeared')
+        stub_fixture(url, 'scrape_no_replies')
+        post :create, params: { button_import: true, dreamwidth_url: url }
+        expect(response).to redirect_to(posts_url)
+        expect(flash[:success]).to eq("Post has begun importing. You will be updated on progress via site message.")
+        expect(ScrapePostJob).to have_been_enqueued.with(url, {}, user: user).on_queue('low')
+      end
+
+      it "performs scrape" do
```
Throne3d, line 81: <3

Throne3d (approved):

app/services/post_importer.rb
teceler: I'm somewhat confused on how it's ensuring that? how much of this is the !! affecting.

Throne3d: it should be the whole thing, `(parsed_url.host&ends_with?('dreamwidth.org')`, which evaluates to `parsed_url.host` if it's nil and otherwise the overall thing (a bool) if it's not, and then applies `!!` to that

```rb
irb(main):001> require 'active_support'
=> true
irb(main):002> require 'active_support/core_ext' # for String#ends_with?
=> true
irb(main):004> x = nil
=> nil
irb(main):006> x&.ends_with?('dreamwidth.org')
=> nil
irb(main):008> !!x&.ends_with?('dreamwidth.org')
=> false
irb(main):010> x = 'url.dreamwidth.org'
=> "url.dreamwidth.org"
irb(main):012> x&.ends_with?('dreamwidth.org')
=> true
irb(main):014> !!x&.ends_with?('dreamwidth.org')
=> true
irb(main):016> x = 'example.com'
=> "example.com"
irb(main):018> x&.ends_with?('dreamwidth.org')
=> false
irb(main):020> !!x&.ends_with?('dreamwidth.org')
=> false
```

teceler: Huh, I would definitely expect it to just operate on x, but okay.

PR #1226, opened 2019-11-13
Rewrite status code

teceler:
* Makes `params[:status]` an int rather than a string
* Creates `status_state` helper
* Adds a status validation on the post model

### DIFF
--- a/app/helpers/writable_helper.rb
+++ b/app/helpers/writable_helper.rb
@@ -55,6 +55,22 @@ module WritableHelper
     image_tag("icons/#{img}.png", class: 'vmid', title: name)
   end
 
+  STATUS_MAP = {
+    Post::STATUS_ACTIVE    => ['In Progress', 'book_open'],
+    Post::STATUS_COMPLETE  => ['Complete', 'book'],
+    Post::STATUS_HIATUS    => ['On Hiatus', 'hourglass'],
+    Post::STATUS_ABANDONED => ['Abandoned', 'book_grey'],
+  }
+
+  def status_state(status, only_icon: false, only_text: false, short_title: false)
+    name = STATUS_MAP[status][0]
+    img = STATUS_MAP[status][1]
+    return name if only_text
+    icon = image_tag("icons/#{img}.png", class: 'vmid', title: "#{'Thread' unless short_title} #{name}")
+    return icon if only_icon
+    [icon, name].join(' ')
+  end
+
   def menu_img
     return 'icons/menu.png' unless current_user.try(:layout).to_s.start_with?('starry')
     'icons/menugray.png'
--- a/app/helpers/writable_helper.rb
+++ b/app/helpers/writable_helper.rb
@@ -70,7 +70,7 @@ module WritableHelper
     return name if only_text
     icon = image_tag("icons/#{img}.png", class: 'vmid', title: "#{'Thread' unless short_title} #{name}")
     return icon if only_icon
-    [icon, name].join(' ')
+    [icon, name].safe_join(' ')
   end
 
   def menu_img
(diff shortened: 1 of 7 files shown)


app/helpers/writable_helper.rb
```diff
@@ -55,6 +55,22 @@ def privacy_icon(privacy)
...
+    Post::STATUS_ABANDONED => ['Abandoned', 'book_grey'],
+  }
+
+  def status_state(status, only_icon: false, only_text: false, short_title: false)
+    name = STATUS_MAP[status][0]
+    img = STATUS_MAP[status][1]
+    return name if only_text
+    icon = image_tag("icons/#{img}.png", class: 'vmid', title: "#{'Thread' unless short_title} #{name}")
+    return icon if only_icon
+    [icon, name].join(' ')
```
Throne3d: Does this not need to `safe_join`?

app/helpers/writable_helper.rb
```diff
@@ -55,6 +55,22 @@ def privacy_icon(privacy)
...
+  STATUS_MAP = {
+    Post::STATUS_ACTIVE    => ['In Progress', 'book_open'],
+    Post::STATUS_COMPLETE  => ['Complete', 'book'],
+    Post::STATUS_HIATUS    => ['On Hiatus', 'hourglass'],
+    Post::STATUS_ABANDONED => ['Abandoned', 'book_grey'],
+  }
+
+  def status_state(status, only_icon: false, only_text: false, short_title: false)
+    name = STATUS_MAP[status][0]
+    img = STATUS_MAP[status][1]
```
Throne3d: Maybe we should do some kind of check to make sure `status` can be found in `STATUS_MAP`, or else error, so we can catch mistakes better in development? I guess we'd get it anyway, as a NilMethodError (`STATUS_MAP[status][0]` would be `nil[0]`), but probably better to be explicitly.

app/views/posts/stats.haml
```diff
@@ -28,19 +28,7 @@
...
-        - when Post::STATUS_COMPLETE
-          = image_tag "icons/book.png", class: 'vmid', title: "In Progress"
-          Complete
-        - when Post::STATUS_HIATUS
-          = image_tag "icons/hourglass.png", class: 'vmid', title: "In Progress"
-          On Hiatus
-        - when Post::STATUS_ABANDONED
-          = image_tag "icons/book_grey.png", class: 'vmid', title: "In Progress"
-          Abandoned
+        =status_state(@post.status, short_title: true)
```
Throne3d: (Stylistic comment) there should be a space after the = symbol

Throne3d:
I'm not 100% sure it makes sense to rewrite our status codes so they're externally understandable – though this is only within the web interface for now, in the future we might use similar code for the API, and someone might start depending on us using `1` for COMPLETE (or whatever), and then we might want to change that in the future.

I do like the `status_state` and status validation things, though! They help clean up the code quite nicely.

@Marri do you have opinions on exposing the enum values? I guess we're not expecting people to start hacking together HTTP requests at us, and we're not intending to support that, so maybe it's not a problem, but I feel a little concerned about it.

teceler:
Honestly I _started_ this because brakeman was complaining about const_get being used in our status update code and that theoretically being super unsafe. There's probably other ways to solve that -- a method that checks it's there and then gives the right value -- but.

app/helpers/writable_helper.rb
teceler: What should I have it do if it can't find it?

Throne3d:
Since we have a `STATUS_MAP` anyway, we could add a new parameter to those that we can use to work backwards? From e.g. `completed` to `Post::STATUS_COMPLETED` (maybe as a third parameter, or maybe by turning them into hashes, or maybe by creating a new hash for the backwards map). That way, we wouldn't be doing const_get, but we'd still be keeping our enum values internal.

app/helpers/writable_helper.rb
Throne3d: We can raise an error, since it should be caught in development or else we should see error messages loudly when we push a broken change.

PR #1227, opened 2019-11-14
Use safe_join (and html_safe) in author_links

teceler:
Mostly to avoid having to flag the output with html_safe

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -234,7 +234,7 @@ module ApplicationHelper
     authors = post.authors.reject(&:deleted?).sort_by{|a| a.username.downcase}
     num_deleted = total - authors.size
     deleted = 'deleted user'.pluralize(num_deleted)
-    return "(#{deleted})".html_safe if authors.empty?
+    return "(#{deleted})" if authors.empty?
 
     if total < 4
       links = authors.map { |author| linked ? user_link(author, colored: colored) : author.username }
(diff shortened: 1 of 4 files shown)


app/helpers/application_helper.rb
```diff
@@ -234,18 +234,19 @@ def author_links(post, linked: true, colored: false)
     authors = post.authors.reject(&:deleted?).sort_by{|a| a.username.downcase}
     num_deleted = total - authors.size
     deleted = 'deleted user'.pluralize(num_deleted)
-    return "(#{deleted})" if authors.empty?
+    return "(#{deleted})".html_safe if authors.empty?
```
Throne3d: I don't think this needs to be marked `html_safe` (contains no links, just plain text), and we should be wary of marking things `html_safe` in general. Is it okay to remove?

Throne3d:
This looks good! Just the one small comment, so we can try to remove as much `html_safe` as possible =)

app/helpers/application_helper.rb
teceler: does haml sanitize helper return values okay? if so yeah I can do that. Mostly was trying to make the returns consistently safe xD

Throne3d: The HTML sanitize helpers should be fine! Are those used here?

I think the main thing shouldn't be just marking all the returns consistently html_safe, but making them html_safe iff they only have valid and safe HTML inside (i.e. only if they're _actually_ safe – which this is, but it doesn't need to be) – so in the cases of links we're manipulating with `safe_join`, we're definitely happy it's returning as html_safe (so that it displays HTML as HTML), but in the case of just displaying random text, we don't need the `html_safe`ty.

PR #1233, opened 2019-11-16
Improve test coverage

Marri:
Should now be 99.58%. Also incidentially removes the "helpfully" auto-created tab for the Channels group in our SimpleCov output, since we don't have channels.


teceler:
I'm concerned about the fact that the post previous/next changes add Yet Another Method to Post, which already has Too Many, but otherwise this looks good!

Issue #828, opened 2018-06-12
Accessibility


Throne3d:
I can't find an overarching issue for this topic, so I'm creating this epic here.

Useful resources:

- https://www.w3.org/standards/webdesign/accessibility
- https://www.w3.org/TR/html-aria/

echurilov:
Pastebin snippet from a conversation about my use case https://pastebin.com/T9d1XaWK

echurilov:
This is almost certainly going to be an undertaking but it's something I care about a lot and would like to try working towards.

PR #1242, opened 2019-12-08
Add alt text to image tags

### LINKED ISSUE #1241: Add alt text on unread link

teceler:
Fixes #1241 and improves #828

### DIFF
--- a/app/helpers/writable_helper.rb
+++ b/app/helpers/writable_helper.rb
@@ -46,13 +46,14 @@ module WritableHelper
   }
 
   def privacy_state(privacy)
-    privacy_icon(privacy) + ' ' + PRIVACY_MAP[privacy][0]
+    privacy_icon(privacy, false) + ' ' + PRIVACY_MAP[privacy][0]
   end
 
-  def privacy_icon(privacy)
+  def privacy_icon(privacy, alt=true)
     name = PRIVACY_MAP[privacy][0]
     img = PRIVACY_MAP[privacy][1]
-    image_tag("icons/#{img}.png", class: 'vmid', title: name)
+    text = alt ? name : ''
+    image_tag("icons/#{img}.png", class: 'vmid', title: name, alt: text)
   end
 
   def menu_img
--- a/app/views/galleries/_expandable.haml
+++ b/app/views/galleries/_expandable.haml
@@ -12,16 +12,16 @@
   %td.width-70.gallery-buttons{class: klass}
     - if gallery.user_id == current_user.try(:id)
       = link_to add_gallery_path(gallery), title: "Add Icons", class: 'gallery-add' do
-        = image_tag "icons/add.png"
+        = image_tag "icons/add.png", alt: 'Add Icons'
       = link_to edit_gallery_path(gallery), title: "Edit Gallery", class: 'gallery-edit' do
-        = image_tag "icons/pencil.png"
+        = image_tag "icons/pencil.png", alt: 'Edit'
       = link_to gallery_path(gallery), title: "Delete Gallery", method: :delete, data: { confirm: "Are you sure you want to delete #{gallery.name}?" }, class: 'gallery-delete' do
-        = image_tag "icons/cross.png"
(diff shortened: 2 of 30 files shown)


app/views/galleries/_expandable.haml
```diff
@@ -12,16 +12,16 @@
...
+        = image_tag "icons/pencil.png", alt: 'Edit'
       = link_to gallery_path(gallery), title: "Delete Gallery", method: :delete, data: { confirm: "Are you sure you want to delete #{gallery.name}?" }, class: 'gallery-delete' do
-        = image_tag "icons/cross.png"
+        = image_tag "icons/cross.png", alt: 'Delete'
   %td.centered.width-70{class: klass}
     .gallery-box.vmid{id: "minmax-#{gallery.id}", style: 'display: inline-block;', data: {id: gallery.id}}
       .view-button{style: 'padding: 5px 7px;'}
-        = image_tag "icons/bullet_arrow_down.png", class: 'vmid down-arrow'
-        = image_tag "icons/bullet_arrow_up.png", class: 'vmid hidden up-arrow'
+        = image_tag "icons/bullet_arrow_down.png", class: 'vmid down-arrow', alt: 'Display'
```
Throne3d: I'd probably go with "Show", since Show/Hide tend to be a common pair (and Display can be a noun too)

Throne3d: Same for the other Display/Hide pairs

app/views/posts/_list_item.haml
```diff
@@ -3,23 +3,23 @@
...
   %td.post-subject.vtop{class: klass}
     - if logged_in?
       - if unread_post?(post, unread_ids) || (@show_unread && !opened_post?(post, opened_ids))
         = link_to unread_path(post) do
-          = image_tag unread_img, class: 'vmid', title: 'First Unread'
+          = image_tag unread_img, class: 'vmid', title: 'First Unread', alt: 'First Unread'
       - elsif @show_unread
         = link_to unread_path(post) do
-          = image_tag lastlink_img, class: 'vmid', title: 'First Unread'
+          = image_tag lastlink_img, class: 'vmid', title: 'First Unread', alt: ''
```
Throne3d: I think this happens when the post is totally read but we're showing a link anyway (like the posts owed page) – this should have an alt too, right?

app/helpers/writable_helper.rb
```diff
@@ -49,10 +49,11 @@ def privacy_state(privacy)
     privacy_icon(privacy) + ' ' + PRIVACY_MAP[privacy][0]
   end
 
-  def privacy_icon(privacy)
+  def privacy_icon(privacy, alt=false)
```
Throne3d: Where do we use `privacy_icon`? I have the feeling that we should show alt text by default, and only remove it if we explicitly don't need it (e.g. because we have the text nearby, like in the above `privacy_state` method)

Throne3d:
This looks really good! Would you be able to rebase this onto master and look at my comments?

Throne3d:
For future readers, here's a good guideline to how we should be doing alt text: https://www.w3.org/WAI/tutorials/images/decision-tree/

app/helpers/writable_helper.rb
teceler: I can flip the default, sure.

app/views/posts/_list_item.haml
teceler: This was one of the ones I was more uncertain on, because it seemed to straddle the line between functional and decorative, but if you think it should have alt text I can do that. I just... need alt text for it :P

Throne3d: I'd go with `First Unread` or `Jump to First Unread`, same as the other, even if that doesn't technically describe it perfectly. Alternatively, `Jump to end` or maybe `Jump to end (read)` would be good?

Throne3d: (I think basically every image inside a link without accompanying text is functional, fwiw)

PR #1253, opened 2019-12-17
Only update tagged_at for status changes when post is completed

teceler:


app/models/post.rb
```diff
@@ -310,7 +312,7 @@ def set_timestamps
     return if skip_edited
     self.edited_at = self.updated_at
     return if skip_tagged
-    return if replies.exists? && !status_changed?
+    return if replies.exists? && status != Post::STATUS_COMPLETE
```
Marri: I think you might still need to check if the status changed here, as well as what it is?

app/models/post.rb
```diff
@@ -200,7 +200,9 @@ def read_time_for(viewing_replies)
 
     # testing for case where the post was changed in status more recently than the last reply
     audits_since_last_reply = audits.where('created_at > ?', most_recent.created_at)
-    audit = audits_since_last_reply.detect { |a| a.audited_changes.key?('status') }
+    audit = audits_since_last_reply.detect do |aud|
```
Marri: plausibly we can do something fancier with the actually queryable column here

app/models/post.rb
teceler: ...oh. yes. I did... not entirely think that through, yeah xD

app/models/post.rb
```diff
@@ -199,9 +199,9 @@ def read_time_for(viewing_replies)
     return most_recent.updated_at if most_recent.updated_at > edited_at
 
     # testing for case where the post was changed in status more recently than the last reply
-    audits_since_last_reply = audits.where('created_at > ?', most_recent.created_at)
-    audit = audits_since_last_reply.detect { |a| a.audited_changes.key?('status') }
-    return most_recent.updated_at unless audit
+    audits_exist = audits.where('created_at > ?', most_recent.created_at).where(action: 'update')
+    audits_exist = audits_exist.where("(audited_changes -> 'status' ->> 1)::integer = ?", Post::STATUS_COMPLETE).exists?
+    return most_recent.updated_at unless audits_exist
```
Marri: for viewport reasons I propose:

    audits = audits.where('created_at > ?', most_recent.created_at).where(action: 'update')
    audits = audits_exist.where("(audited_changes -> 'status' ->> 1)::integer = ?", Post::STATUS_COMPLETE).exists?
    return most_recent.updated_at unless audits.exists?

teceler: ...you've got .exists? there twice, but

Marri: YOU SAW NOTHING NOTHING HAPPENED

PR #1262, opened 2019-12-22
Remove unused gems from production

Marri:
- rails_12factor is handled natively by Rails 5 and the correct variables are set in production
- we do not schedule Resque jobs
- we do not write coffeescript


Marri:
To answer your questions in order:
- coffee rails is included by default by Rails and it took me this long to consider that perhaps it's not actually integral to application function
- rails_12factor was actually required for Rails 4
- resque_scheduler we were going to use and abandoned when we realized it would require another worker process and we already use our one worker for processing resque jobs

PR #1264, opened 2019-12-22
Require HTTP basic auth for the Resque web console

Marri:
We used to have this, it went away when we moved from the resque_web gem to the built-in server that comes with the resque gem, this puts it back. Production has these ENV variables already set to reasonable values.

### DIFF
--- a/config/initializers/resque.rb
+++ b/config/initializers/resque.rb
@@ -1,3 +1,5 @@
+require 'resque/server'
+
 Resque.before_fork do
   defined?(ActiveRecord::Base) and ActiveRecord::Base.connection.disconnect!
 end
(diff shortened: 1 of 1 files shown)


config/initializers/resque.rb
```diff
@@ -16,6 +18,12 @@
 }
 Resque::Mailer.excluded_environments = [] # I explicitly want this to run in tests; don't exclude them.
 
+Resque::Server.use(Rack::Auth::Basic) do |user, password|
+  username = ENV['RESQUE_WEB_HTTP_BASIC_AUTH_USER'] || 'user'
+  pw = ENV['RESQUE_WEB_HTTP_BASIC_AUTH_PASSWORD'] || 'secret'
+  [user, password] == [username, pw]
+end
```
teceler: This seems unnecessary on dev but idk how hard it'd be to disable there safely? As long as the login is persisted it's not a huge deal.

PR #1266, opened 2019-12-22
Don't paginate character list now that the memory reduction is in place

Marri:
@teceler


teceler:
You missed the horrifyingly convoluted code for paginating template list view

PR #1267, opened 2019-12-22
Tests

Marri:
- Test the code to log out a nonexistent user
- Test the index#new code, and incidentally fix a bug
- Test the Daily Report sorting code, and incidentally make it db-safer
- Bump code coverage

### DIFF
--- a/app/views/indexes/_editor.haml
+++ b/app/views/indexes/_editor.haml
@@ -4,7 +4,7 @@
     %td.even= f.text_field :name, placeholder: "Index Name", class: 'text'
   %tr
     %th.sub Owner
-    %td.odd= @inder.user.deleted? ? '(deleted user)' : @index.user.username
+    %td.odd= @index.user.deleted? ? '(deleted user)' : @index.user.username
   %tr
     %th.sub Privacy
     %td.even
--- a/spec/controllers/reports_controller_spec.rb
+++ b/spec/controllers/reports_controller_spec.rb
@@ -50,6 +50,42 @@ RSpec.describe ReportsController do
       expect(response).to have_http_status(200)
     end
 
+    it "sorts by timestamp" do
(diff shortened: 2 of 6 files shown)


app/views/indexes/_editor.haml
```diff
@@ -4,7 +4,7 @@
     %td.even= f.text_field :name, placeholder: "Index Name", class: 'text'
   %tr
     %th.sub Owner
-    %td.odd= @inder.user.deleted? ? '(deleted user)' : @index.user.username
+    %td.odd= @index.user.deleted? ? '(deleted user)' : @index.user.username
```
teceler: Eep! good job finding and fixing that!

spec/controllers/reports_controller_spec.rb
```diff
@@ -50,6 +50,42 @@
...
+      board2 = create(:board, name: 'dd')
+      post1 = Timecop.freeze(today) { create(:post, board: board1) }
+      post2 = Timecop.freeze(today + 2.hours) { create(:post, board: board1) }
+      post3 = Timecop.freeze(today) { create(:post, board: board2) }
+      post4 = Timecop.freeze(today + 1.hour) { create(:post, board: board2) }
+      user = create(:user, timezone: 'UTC')
+      login_as(user)
+      get :show, params: { id: 'daily', day: today.to_date.to_s, sort: 'continuity' }
+      expect(assigns(:posts).map(&:id)).to eq([post2, post1, post4, post3].map(&:id))
+    end
```
teceler: Might be worth making a describe and sharing some of these variables?

app/views/indexes/_editor.haml
Marri: well we'd've gotten an error email if anyone had tried to load the page and we haven't, sooooo I don't think this has had a ton of impact :joy: but yes it was useful!

PR #1268, opened 2019-12-22
Index improvements

Marri:
Get indexes to the point where they are a semi-usable feature, rather than an in progress thing no one knows or cares about.

- [ ] Do we want to disallow repeating posts in an index or index section? There's no validations, so leaving it for now. If we do disallow, we need validations and also api/posts should filter duplicates in index_posts#new.
- [ ] Do we want to allow changing the IndexSection for arbitrary IndexPosts? Right now it just shows/links.


Throne3d:
> Do we want to disallow repeating posts in an index or index section? There's no validations, so leaving it for now. If we do disallow, we need validations and also api/posts should filter duplicates in index_posts#new.

Within a section, I'd think it makes sense to forbid. Within an overall index, not sure! Since it's a bit awkward and doesn't really add much for us from a development perspective, I think we can just leave it as is, though.

>  Do we want to allow changing the IndexSection for arbitrary IndexPosts? Right now it just shows/links.

What does this mean? Moving a post around between sections in an index seems fine – is there something more to it than that?

app/models/index_section.rb
```diff
@@ -2,15 +2,23 @@ class IndexSection < ApplicationRecord
   include Orderable
 
   belongs_to :index, inverse_of: :index_sections, optional: false
-  has_many :index_posts, inverse_of: :index_section, dependent: :destroy
-  has_many :posts, through: :index_posts, dependent: :destroy
+  has_many :index_posts, inverse_of: :index_section, dependent: false # This is handled in callbacks
```
Throne3d: Is there a reason not to use `dependent: :nullify` here? (Wanting to audit when this changes?)

app/views/boards/_reorder.haml
```diff
@@ -0,0 +1,51 @@
+- reset_cycle
+#reorder-sections-table
+  .content-header
+    = content_for :reorder_title
+    #loading.float-right.hidden= loading_tag # TODO duplicate id
+    #saveerror.float-right.hidden
+      - # TODO duplicate id
+      = image_tag "icons/exclamation.png", title: 'Error', class: 'vmid', alt: '!'
```
Throne3d: `alt`s should be changed in line with #1242 once that gets merged (in this case, it should be empty)

app/views/boards/_reorder.haml
```diff
@@ -0,0 +1,51 @@
...
+  %ul.sortable.table-list
+    - sections.each do |section|
+      %li.padding-5.section-ordered{class: cycle('even', 'odd'), data: {id: section.id, type: section.class.to_s, order: section.section_order}}
+        .section-ordered-handle
+          = image_tag "icons/arrow_double.png", title: 'Reorder', class: 'disabled-arrow', alt: '↕'
+        = link_to section.name, section
+        .float-right
+          = image_tag "icons/arrow_up.png", class: "section-up disabled-arrow"
+          = image_tag "icons/arrow_down.png", class: "section-down disabled-arrow"
+          = link_to url_for(section) + "/edit" do # TODO
```
Throne3d: What's TODO here? `edit_section_url(section)` or something?

app/views/boards/_reorder.haml
```diff
@@ -0,0 +1,51 @@
...
+        Saved
+    %ul.sortable.table-list
+      - unsectioned_posts.each do |post|
+        %li.section-ordered{class: cycle('even', 'odd'), data: {id: post.id, type: post.class.to_s, order: post.section_order}}
+          .section-ordered-handle
+            = image_tag "icons/arrow_double.png", title: 'Reorder', class: 'disabled-arrow', alt: '↕'
+          = link_to post.subject, post_path(post)
+          .float-right
+            = image_tag "icons/arrow_up.png", class: "section-up disabled-arrow"
+            = image_tag "icons/arrow_down.png", class: "section-down disabled-arrow"
```
Throne3d: (missing trailing newline)

app/views/index_sections/_single.haml
```diff
@@ -1,25 +1,25 @@
...
-- section_posts = posts_from_relation(section.posts.ordered_by_index, with_pagination: !in_collection, select: ', index_posts.description as index_description')
+    %th.subber.index-description{colspan: 7}= sanitize_written_content(section.description)
+- section_posts = posts_from_relation(section.posts.ordered_by_index, with_pagination: !in_collection, select: ', index_posts.description as index_description, index_posts.id as index_post_id')
 - if section_posts.present?
-  = render partial: 'posts/list_item', collection: section_posts, as: :post
+  = render partial: 'posts/list_item', collection: section_posts, as: :post, locals: {index: section.index}
 - else
   %tr
-    %td.centered.padding-10.no-posts{ class: cycle('even', 'odd'), colspan: 6 } — No posts yet —
+    %td.centered.padding-10.no-posts{ class: cycle('even', 'odd'), colspan: 7 } — No posts yet —
```
Throne3d: small: We should probably extract colspan to a variable so future diffs are more readable

app/views/boards/_reorder.haml
Marri: Problem is that it's either edit_board_section_url or edit_index_post_url; url_for will give the correct thing for the model but there's no equivalent for the edit url and appending "/edit" is not exactly Good Form

teceler: use a helper and check the class?

Throne3d: `url_for(section, action: 'edit')`?

Throne3d: or even probably `link_to section, action: 'edit'`, since I think it defers to url_for directly

Throne3d: If that doesn't work, it looks like `edit_polymorphic_url(section)` may? https://api.rubyonrails.org/classes/ActionDispatch/Routing/PolymorphicRoutes.html

app/models/index_section.rb
Marri: There is indeed a reason, and it is that we need the callbacks to handle reordering them

app/views/boards/_reorder.haml
Marri: the action ones didn't work but edit_polymorphic did

app/views/boards/_reorder.haml
Marri: Did get to go through merge conflict fun, yep xD

PR #1269, opened 2019-12-23
Try backstopjs

teceler:


script/before_backstop.rb
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env ruby
+
+layout ||= ARGV[0] || nil
```
Throne3d: This line is strange - `layout = ARGV[0]` should work identically

backstop/starrylight.js
```diff
@@ -0,0 +1,9 @@
+var config = require('./core.js');
+
+config.id = 'starrylight';
+config.paths.bitmaps_reference = "then/starrylight";
```
Throne3d: Could maybe extract all these `config.paths.` things out into a function in `core.js` (or such) that accepts `layout` and generates them, so here we just have `setConfig(config, 'starrylight');`

script/before_backstop.rb
teceler: ...I think that's an artifact of this being basically before_wraith renamed, which was _also_ used in another script. or at least the `||=` is. idk about the `|| nil` xD

backstop/starrylight.js
teceler: Possibly but I didn't want to do _more_ javascript things when it was easy enough to do this.

Throne3d: That's fair ­– if we ever get around to editing this list, we should clean it up then, then, since the duplication may make it hard to keep consistent.

PR #1277, opened 2019-12-26
Use Docker for PostgreSQL & Redis dependencies by default

Throne3d:
Upgrade route:

- Docker-only users: nothing should change, `docker-compose up` as normal
- Users of Docker for dependencies only: you are now able to run the server with `docker-compose up postgres redis &` then `rails s`
- Users of local PostgreSQL & Redis: first, run `echo -e "GLOWFIC_DATABASE_USER=\nGLOWFIC_DATABASE_PASS=" > .env`, then `rails s` as normal

For those using a local PostgreSQL & Redis, this `.env` file will be read by the application and override the new defaults, causing the application to try to connect to your database with your Unix username and whatever password you set inside `~/.pgpass`. Redis requires no changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,8 @@ This will set up the database and add some sample information – currently thes
 
 If you encounter an error involving the 'citext' extension:
 
-*   Execute `sudo -u postgres bin-docker/psql`, and then in the prompt that appears:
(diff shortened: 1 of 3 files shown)


Throne3d:
Updated a bit more – see PR body for details of how to proceed (editing it now).

PR #1280, opened 2019-12-27
Use ActsAsTaggableOn for Tags

teceler:
Todo:
- [x] Fix reset_warnings
- [x] Fix has_content_warnings
- [x] Fix tags#index (normal and api)
- [x] Fix tag#show somehow
- [x] Work out what ids_for_select is doing and replace it
- [ ] Fix select2
- [x] Fix post preview


app/controllers/tags_controller.rb
```diff
@@ -112,24 +111,45 @@ def og_data
...
+    children_ids = params.fetch(:tag, {})[:children_ids]
+    return unless children_ids.present?
+    children_ids -= @tag.children_ids
+    children_ids.each do |owner_id|
+      ActsAsTaggableOn::Tagging.create(taggable_type: 'ActsAsTaggableOn::Tag', taggable_id: owner_id, tag: @tag, context: 'settings')
+    end
+  end
+
   def tag_params
     permitted = [:type, :description, :owned]
```
teceler: NTS: remove `:owned` here

db/migrate/20191231000858_migrate_tags.rb
```diff
@@ -0,0 +1,109 @@
...
+    drop_table :character_tags
+    drop_table :gallery_tags
+    drop_table :tag_tags
+    remove_column :tags, :user_id
+    remove_column :tags, :owned
+
+    ActsAsTaggableOn::Tag.reset_column_information
+    ActsAsTaggableOn::Tag.find_each do |tag|
+      ActsAsTaggableOn::Tag.reset_counters(tag.id, :taggings)
+    end
```
teceler: On consideration this basically serves no purpose for us and should probably go.

PR #1288, opened 2019-12-29
Add user moiety & created_at to API

Throne3d:
Moieties requested by DanielH, created_at to give parity with the user index of the UI. Builds on #1286 to reduce diffs.

### DIFF
--- a/app/presenters/user_presenter.rb
+++ b/app/presenters/user_presenter.rb
@@ -8,7 +8,23 @@ class UserPresenter
   def as_json(options={})
     return {} unless user
     return {id: user.id, username: '(deleted user)'} if user.deleted?
+    return detailed_json(options) if options[:detailed]
+    summary_json(options)
+  end
+
+  private
+
+  def detailed_json(options={})
     attrs = %w(id username moiety moiety_name created_at)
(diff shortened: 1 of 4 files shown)


app/presenters/user_presenter.rb
```diff
@@ -8,6 +8,7 @@ def initialize(user)
   def as_json(options={})
     return {} unless user
     return {id: user.id, username: '(deleted user)'} if user.deleted?
-    user.as_json_without_presenter({only: [:id, :username]}.reverse_merge(options))
+    attrs = %w(id username moiety moiety_name created_at)
```
Marri: given my general feelings about moieties, could this maybe be returned in... many fewer places xD

Throne3d: I couldn't figure out where we return users – I checked for explicit `to_json` calls. Would I have to check the whole API? :/

Throne3d: Oh, I guess maybe our examples are comprehensive enough that I can focus on those?

Throne3d: Is this better? I put it in users#index only, but it still shows up three times.

PR #1293, opened 2020-01-05
Use numeric badge not glow for Daily Report

Marri:


app/views/layouts/application.haml
```diff
@@ -112,13 +112,12 @@
...
             %li= link_to 'Tags', tags_path
-            - if (day = DailyReport.unread_date_for(current_user)).present?
-              %li
-                = link_to report_path(id: :daily, day: day), style: 'text-shadow: 0 0 0.8em #F87, 0 0 0.8em #F87' do
-                  Daily Report
-                  %em= '(' + Date.parse(day).strftime("%b %d, %Y") + ')'
-            - else
-              %li= link_to 'Daily Report', report_path(id: :daily)
+            - day = DailyReport.unread_date_for(current_user)
+            - daybadge = day.present? ? (DateTime.now.to_date - day).to_i : 0
```
Throne3d: Is `DateTime.now` definitely the safe timezone version? As far as I know, that'll be the built-in Ruby one, which will default to the server's system time instead. (Looks like [`DateTime.current`](https://api.rubyonrails.org/classes/DateTime.html#method-c-current) is probably safe, or we could explicitly do `Time.zone.now.to_datetime`?)

Marri: Hmmm I was expecting it to be safe because our views are wrapped in timezone-setting code but that may not in fact be the case; if I make this any more complex  I should maybe extract it to DailyReport too though.

PR #1295, opened 2020-01-08
Private post notes

Marri:
Screenshot from 2020-01-09 07-44-20
Screenshot from 2020-01-09 07-45-49

### DIFF
--- a/app/assets/stylesheets/replies.scss
+++ b/app/assets/stylesheets/replies.scss
@@ -288,8 +288,6 @@ div.post-subheader { width: 100%; }
     label { min-width: 95px; }
   }
 
-  .clear { height: 1px; }
-
   input[type='checkbox'] {
     margin-left: 5px;
     position: relative;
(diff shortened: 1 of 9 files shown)


teceler:
Hmm, I feel like whether notes should be visible to all authors is... a question to consider? (Rn i think they're not?)

Marri:
Oh yeah I mean long term goal I'd also like to have notes that are visible to all authors as well as notes that are visible to anyone reading the post? This is just the notes-to-only-yourself piece. And if we add more later they can go neatly into the new Notes section of the post editor UI, just, in this there's only the one.

app/assets/stylesheets/replies.scss
```diff
@@ -288,8 +288,6 @@ div.post-subheader { width: 100%; }
     label { min-width: 95px; }
   }
 
-  .clear { height: 1px; }
```
Marri: Moving this because it was broken; this is no longer the correct parent element.

PR #1298, opened 2020-01-15
Add minio to replicate S3 bucket functionality locally

Throne3d:
This is not in the most perfect of states, as we're cluttering the aws initializer with a lot of dev code (including creating a bucket if it doesn't exist). Opinions on how to clean this up, or is it alright as is?

I haven't tested this out under Docker 100% – instead I had Redis, PostgreSQL and MinIO running under Docker, and used this `.env` file (because my Docker runs in a virtual machine):

```
DATABASE_URL=postgres://192.168.99.100:5432
REDIS_URL=redis://192.168.99.100:6379
MINIO_ENDPOINT=http://192.168.99.100:9000/
```


Throne3d:
Travis seems to work both with and without minio explicitly enabled! Not sure what's going on.

PR #1301, opened 2020-01-18
Prevent update_board_cameos from trying to create duplicate board cameos

teceler:
Okay the edge case where the weird caching happened was in #705 but this fix, unlike most of that branch, doesn't _need_ to be in there, and also other things could trigger that edge case.

### DIFF
--- a/app/concerns/owable.rb
+++ b/app/concerns/owable.rb
@@ -53,9 +53,10 @@ module Owable
 
       # adjust for the fact that the associations are managed separately
       all_authors = authors + unjoined_authors + joined_authors + tagging_authors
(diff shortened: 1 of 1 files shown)


Throne3d:
To prevent this issue in the future, I think it'd be good to turn these scopes (like `post.cameos`) into functions that return relations, rather than being true "relations" – that way we have to handle them properly in all code paths.

This is good to merge separately if you want, though!

teceler:
I feel like functions that result relations get us a _different_ kind of problems! :thinking: I wonder if you can make regular rails relations read-only.

PR #1302, opened 2020-01-18
Test that post#preview and reply#preview don't create user associations

teceler:
(extracted from #705)

Tests that post#create and post#update preview does not create PostAuthors, PostViewers, or BoardAuthors
Tests that reply#create preview does not create PostAuthors or BoardAuthors


Throne3d:
Do we also test BoardAuthors aren't updated on post previews? If not, that'd be good to add! Previewing a post edit would be good, too, if we don't yet do that – that's one of the particularly bad error cases (almost-updating an object, and ending up updating its relations as a result of that).

This is good regardless, though!

teceler:
Now tests BoardAuthors and post#update preview :)

PR #1303, opened 2020-01-19
Extend rack timeout

teceler:


Throne3d:
I don't _think_ this is the right way to approach this, but we're still discussing it in Discord – note that this only extends the PostgreSQL statement timeout, and the issues we had were with Rack::Timeout

Throne3d:
Closing this for now – we made some other changes to timeouts, and how we respond to further failures will probably be different from this.

PR #1306, opened 2020-01-19
Add bullet tests for all render_views specs

Throne3d:
Very much WIP.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -56,6 +56,7 @@ group :development do
 end
 
 group :development, :test do
+  gem 'bullet'
   gem 'byebug'
   gem 'database_cleaner'
   gem 'dotenv-rails'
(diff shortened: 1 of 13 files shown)


teceler:
Can we generically enable this on feature specs?

Throne3d:
Hmm, I think we could try but I don't know if it'll work as we expect – feature specs load up in a separate server, don't they? I suppose exceptions during that process will cause tests to fail, but it might be with some confusing error (like that the page failed to load, rather than why). I can try to look into it!

Throne3d:
Adding a `skip_bullet` helper for tests - the way Bullet is set up here, it'll catch any associations with a `.includes` where we materialize it without using the associations, which is also true of `expect(assigns(:characters)).to match_array([...])`, where we're not checking the associated template/user. To avoid that, I'm wrapping that in a skip_bullet helper so it doesn't warn.

It feels a bit weird to enable bullet on a subsection of specs, then disable it on a subsection of assertions, but I'm not sure of an easy way to have Bullet only wrap the `get`/`post`/... part of the spec, which is what we're most interested in. I think this is OK for now but we can iterate on the approach to improve it!

PR #1309, opened 2020-01-19
Avoid characters#index spawning queriers per template

teceler:


Throne3d:
Can we rebase this? (Also, are these pages particularly slow, do we know the impact this will have? Seems like a complex MR, wondering about the readability vs optimization question)

PR #1310, opened 2020-01-19
Fix posts#unread for memory and query logic

Throne3d:
Currently loads all posts! And does stuff with it. This is not great. Pending a more proper fix, but this might be a good WIP improvement.


teceler:
I thiiink we fixed this at least somewhat by removing the unpaginationated posts_from_relation call?

Throne3d:
Yeah - I'm still investigating, but I think the first commit here can be removed totally. (The rationale doesn't look as clear as it was, and my attempted simplifications basically turn it into a no-op?)

Throne3d:
I think the rest of this is still cleaner, though! It removes some of the logic from the view (knowing where opened_ids are stored, etc), and I think the `conditional_tag` helper could be useful to us in other situations.

app/helpers/application_helper.rb
```diff
@@ -123,4 +123,12 @@ def user_mem_link(user_id, username, deleted)
...
     link_to username, user_path(user_id)
   end
+
+  def conditional_tag(tag_name, condition, &block)
+    if condition
+      content_tag tag_name, capture(&block)
+    else
+      capture(&block)
+    end
+  end
```
teceler, line 133: ...this looks like *black magic*. How does it *work*.

(Also probably it should be in post_helper?)

Throne3d: So I think I actually grabbed the original setup from the internet (not sure where exactly, but e.g. https://stackoverflow.com/questions/5809917/conditional-tag-wrapping-in-rails-erb#5810148 uses `capture` too)

[`capture`](https://api.rubyonrails.org/classes/ActionView/Helpers/CaptureHelper.html#method-i-capture) takes the result of the block (which is how tag wrapping works in HAML, when you do things like `link_to <url> do`, the `do` bit is indicating a block, and then you have stuff nested inside the link on the next line)

If the condition is true, it takes the capture result and passes it to `content_tag tag_name` ([`content_tag`](https://apidock.com/rails/ActionView/Helpers/TagHelper/content_tag) takes content as its second parameter, or you can pass a block) otherwise it just returns the value of the block itself, which is how these wrapping setups usually seem to be done in helpers!

This could maybe be in post_helper! I'm not sure, I figured it could stay in application_helper since it's quite a generic helper and could be used elsewhere, but since we only use it in posts views right now, we could move it there and bring it back to application_helper only if/when we do use it elsewhere?

teceler: ...I do not feel much less like I am looking at black magic. The only part of that I understood was 'this involves blocks somehow'. and that it accepts a generic condition to check if it uses the tag its given or just the block.

...kind of feel like the genericization may be making this more confusing than it needs to be?

PR #1311, opened 2020-01-19
Improve memory change log

Throne3d:
Improve the log format to something like:

```
Memory usage in posts#unread: 246 MB -> 258 MB (+12.1 MB) | PID: 6314
```

This may help us with debugging memory issues, not sure, as we can try to grep in sumologic for high figures here.

I switched to [this library](https://github.com/schneems/get_process_mem) because I didn't trust the numbers we were getting – they seemed to always be given in KB (e.g. 123KB) despite being in MB according to my system. Plus, this library apparently calculates these values faster on Linux-based systems, and is written by someone who wrote a very sensible post about Ruby timeouts from a Heroku perspective, so the author seems pretty decent.


Throne3d:
I'll rebase this later, but then should be good to merge!

PR #1312, opened 2020-01-19
Fix reports#daily to use subqueries rather than Ruby

Throne3d:
This removes some really staggeringly-large queries when lots of posts (or replies) were created in a single day. I managed to load 5000 posts on a single report day and it took 300ms of controller time, so assuming this is still logically correct, this is a very good improvement! On master it only took about 500ms, anyway, so this isn't too bad, but it presumably helps some with memory usage.

### DIFF
--- a/app/services/daily_report.rb
+++ b/app/services/daily_report.rb
@@ -7,10 +7,10 @@ class DailyReport < Report
 
   def posts(sort='')
     range = day.beginning_of_day .. day.end_of_day
-    created_no_replies = Post.where(last_reply_id: nil, created_at: range).pluck(:id)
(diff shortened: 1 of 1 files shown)


app/services/daily_report.rb
```diff
@@ -7,10 +7,10 @@ def initialize(day)
 
   def posts(sort='')
     range = day.beginning_of_day .. day.end_of_day
-    created_no_replies = Post.where(last_reply_id: nil, created_at: range).pluck(:id)
-    by_replies = Reply.where(created_at: range).pluck(:post_id)
```
teceler: Why was this not a distinct select before. this is a _good_ change, wow.

PR #1315, opened 2020-01-20
Split posts_from_relation into two/three methods

teceler:
Also tweaked a few other things to clarify or optimize the logic.

(Advise viewing with w=1)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -91,27 +91,34 @@ class ApplicationController < ActionController::Base
   helper_method :tos_skippable?
 
   def posts_from_relation(relation, no_tests: true, with_pagination: true, select: '', max: false)
-    if max
-      reports_select = <<~SQL
+    posts = posts_list_relation(relation, no_tests: no_tests, select: select, max: max)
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -91,51 +91,57 @@ def tos_skippable?
...
-      @unread_ids ||= []
-      @unread_ids += opened_posts.select do |view|
-        post = posts.detect { |p| p.id == view.post_id }
-        post && view.read_at < post.tagged_at
-      end.map(&:post_id)
+  def calculate_view_status(posts)
+    @opened_ids ||= PostView.where(user_id: current_user.id).where.not(read_at: nil).pluck(:post_id)
+
+    opened_posts = PostView.where(user_id: current_user.id).where.not(read_at: nil).where(post_id: posts.map(&:id)).select([:post_id, :read_at])
+    opened_posts = opened_posts.select do |view|
```
Throne3d: I'd probably name this `unread_views` by this point – we've already limited to the ones where the `read_at < tagged_at`

teceler: Sure. (Also consolidated the PostView where while I was in there b/c why this xD)

PR #1316, opened 2020-01-20
Clean up ApplicationController methods

teceler:


Throne3d:
A bunch of these methods could probably be extracted to some sort of forward compatibility concern – like, check password warning, the TOS stuff, redirection to the right URL – if we wanted to reduce the number of methods directly in ApplicationController. Did these methods get moved to `private` because it doesn't complain about private methods in classes, or is it just because we don't call them elsewhere?

teceler:
I admit I didn't search _every single one_ of them, I can double-check, it does still complain about private methods but given the assumption we didn't call them elsewhere it seemed the thing to do. Forward-compatibility concern is :thinking: maybe good?

teceler:
The only one of those methods called elsewhere is arguably `set_timezone` in `ApiController`, which overrides it without using super, so I think that's safe. :P

PR #1318, opened 2020-01-21
Add simple feature specs for character list views

Throne3d:


Throne3d:
Tec already tested these out and they seem to be useful, so I'll get merging them!

PR #1319, opened 2020-01-21
Convert feature specs to system specs

Throne3d:


Throne3d:
I'm no longer 100% sure of the motive for this PR, but it sounds like system specs are the more modern version of feature specs, and in particular – it's easier to run JS tests with feature specs, because they handle test data isolation using transactions correctly: https://relishapp.com/rspec/rspec-rails/docs/system-specs/system-spec (feature spec JS tests have to handle that more carefully with DatabaseCleaner =/).

Throne3d:
Doesn't seem super important right now, tbh - closing this until/unless we want system spec support

PR #1323, opened 2020-01-24
Use where or rather than arel on favorites#index

teceler:


app/controllers/favorites_controller.rb
```diff
@@ -3,36 +3,23 @@ class FavoritesController < ApplicationController
...
-      end
-    end
+    user_favorites = @favorites.where(favorite_type: User).select(:favorite_id)
+    author_posts = PostAuthor.where(user_id: user_favorites, joined: true).select(:post_id)
+    board_favorites = @favorites.where(favorite_type: Board).select(:favorite_id)
+    post_favorites = @favorites.where(favorite_type: Post).select(:favorite_id)
 
+    @posts = Post.unscoped
+    @posts = @posts.where(id: author_posts).or(@posts.where(id: post_favorites)).or(@posts.where(board_id: board_favorites))
     @posts = posts_from_relation(@posts.where(where_calc).ordered)
```
Throne3d: Should this be:

```rb
@posts = Posts.where(id: author_posts).or(@posts.where(id: post_favorites)).or(@posts.where(board_id: board_favorites))
@posts = posts_from_relation(@posts.ordered)
```

? It seems to be failing in CI.

I'm also not sure if `favorite_type: User` will work directly, or if you'll need to put it in quotes – seems plausible that'll be necessary.

This looks neat, though!

app/controllers/favorites_controller.rb
```diff
@@ -3,36 +3,22 @@ class FavoritesController < ApplicationController
   before_action :login_required
 
   def index
+    @page_title = 'Favorites'
+
     return if params[:view] == 'bucket'
-    unless current_user.favorites.present?
+    unless (@favorites = current_user.favorites).exists?
```
Marri: This should in fact be `present?` I think. If it doesn't exist, we haven't loaded anything and don't care anyway; if it does exist, we're just going to need to load it again on the next line. No reason to do the extra query.

app/controllers/favorites_controller.rb
```diff
@@ -3,36 +3,22 @@ class FavoritesController < ApplicationController
...
-        where_calc = where_calc.or(new_calc)
-      end
-    end
+    user_favorites = @favorites.where(favorite_type: User.to_s).select(:favorite_id)
+    author_posts = PostAuthor.where(user_id: user_favorites, joined: true).select(:post_id)
+    board_favorites = @favorites.where(favorite_type: Board.to_s).select(:favorite_id)
+    post_favorites = @favorites.where(favorite_type: Post.to_s).select(:favorite_id)
 
-    @posts = posts_from_relation(@posts.where(where_calc).ordered)
+    @posts = Post.where(id: author_posts).or(Post.where(id: post_favorites)).or(Post.where(board_id: board_favorites))
```
Marri: I'm not sure I have the spoons to have the conversation out right now but we'd be moving from a single table indexed-ID based select to a pretty complex join and that should probably be considered at all.

teceler: Subqueries, not a join, but it does increase the query complexity a lot by doing it all on one query, agreed.

app/controllers/favorites_controller.rb
teceler: ...right, it would cache xD I apologize for my overenthusiam there

app/controllers/favorites_controller.rb
Throne3d: Subqueries can be flattened into a join, sometimes, by PostgreSQL automatically, when it makes sense.

Beyond that – for this to get computationally expensive on Postgres, we need a large user favorites list (as otherwise, Postgres will opt to first filter the favorites table to this user, and then do a simple loop over that like we're doing, a sequential scan). If we get a large user favorites list, then currently our Ruby code would have to handle this instead. We can run an EXPLAIN on this against someone's favorites in production to be sure, but it should be significantly cheaper than, e.g., the "recently updated" page (which doesn't seem to flag itself when we look for slow queries).

Throne3d: Forgot to mention, but: in the complex user case, we're currently polling the BoardAuthor table repeatedly for users' post_ids – with a lot of user favorites, the old code doing a lot of repeated scanning of that table with different IDs each time, approximately randomly, which the database has no opportunity to optimize. We may genuinely reduce the overhead PostgreSQL experiences by modifying this query, and it gives us the opportunity to add missing indexes if this query does end up getting slow (rather than obscuring it in lots of smaller queries).

teceler: ...the PostAuthor table, I presume?

Anyway the EXPLAIN ANALYZE for the new query is 

```
Seq Scan on posts  (cost=40.31..55.89 rows=29 width=792) (actual time=0.163..0.227 rows=31 loops=1)
  Filter: ((hashed SubPlan 1) OR (hashed SubPlan 2) OR (hashed SubPlan 3))
  Rows Removed by Filter: 2
  SubPlan 1
    ->  Nested Loop Semi Join  (cost=4.19..14.94 rows=7 width=4) (actual time=0.095..0.105 rows=1 loops=1)
          Join Filter: (post_authors.user_id = favorites.favorite_id)
          Rows Removed by Join Filter: 40
          ->  Seq Scan on post_authors  (cost=0.00..1.65 rows=41 width=8) (actual time=0.014..0.036 rows=41 loops=1)
                Filter: joined
                Rows Removed by Filter: 24
          ->  Materialize  (cost=4.19..12.68 rows=1 width=4) (actual time=0.001..0.001 rows=1 loops=41)
                ->  Bitmap Heap Scan on favorites  (cost=4.19..12.67 rows=1 width=4) (actual time=0.023..0.024 rows=1 loops=1)
                      Recheck Cond: (user_id = 3)
                      Filter: ((favorite_type)::text = 'User'::text)
                      Rows Removed by Filter: 2
                      Heap Blocks: exact=1
                      ->  Bitmap Index Scan on index_favorites_on_user_id  (cost=0.00..4.19 rows=5 width=0) (actual time=0.012..0.012 rows=3 loops=1)
                            Index Cond: (user_id = 3)
  SubPlan 2
    ->  Bitmap Heap Scan on favorites favorites_1  (cost=4.19..12.67 rows=1 width=4) (actual time=0.007..0.009 rows=1 loops=1)
          Recheck Cond: (user_id = 3)
          Filter: ((favorite_type)::text = 'Post'::text)
          Rows Removed by Filter: 2
          Heap Blocks: exact=1
          ->  Bitmap Index Scan on index_favorites_on_user_id  (cost=0.00..4.19 rows=5 width=0) (actual time=0.004..0.004 rows=3 loops=1)
                Index Cond: (user_id = 3)
  SubPlan 3
    ->  Bitmap Heap Scan on favorites favorites_2  (cost=4.19..12.67 rows=1 width=4) (actual time=0.005..0.006 rows=1 loops=1)
          Recheck Cond: (user_id = 3)
          Filter: ((favorite_type)::text = 'Board'::text)
          Rows Removed by Filter: 2
          Heap Blocks: exact=1
          ->  Bitmap Index Scan on index_favorites_on_user_id  (cost=0.00..4.19 rows=5 width=0) (actual time=0.002..0.002 rows=3 loops=1)
                Index Cond: (user_id = 3)
Planning Time: 0.837 ms
Execution Time: 0.431 ms```

with which the only problem I'm seeing is that it's joining post_authors and _then_ filtering by type, which seems... bad xD I can probably fix that with an explict join, just not sure that's always going to be better.

(Idk how to run an explain on the old one that handles it's subqueries properly, though, and also I have... three favorites in seeds...)

Throne3d: This seemed to run immediately with Marri as the sample user on production, so I'm happy with it! I had a bit of trouble getting psql to play nicely, since even simple queries (`SELECT 1`) were taking multiple seconds, but resorting to `rails c` it ended up working fine. We can watch how the page goes in NewRelic for now.

PR #1325, opened 2020-01-25
Set dropdownParent on select2s

teceler:
(Depends on #1324)


teceler:
Note to self: this is broken on posts#new/edit

PR #1327, opened 2020-01-27
Use partials instead of looping in character lists

teceler:
Should produce memory improvements.

(Extracted from #1309)

### DIFF
--- a/app/views/characters/_list_item.haml
+++ b/app/views/characters/_list_item.haml
@@ -1,0 +1,30 @@
+- char_id, char_name, nickname, screenname, pb, user_id, username, user_deleted = character
+- if show_template
+  - template_id, template_name = character[-2..]
+- klass = cycle('even', 'odd')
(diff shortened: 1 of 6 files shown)


app/views/characters/_list_item.haml
```diff
@@ -0,0 +1,30 @@
+- char_id, char_name, nickname, screenname, pb, user_id, username, user_deleted = character
+- if show_template
+  - template_id, template_name = character[-2..]
+- klass = cycle('even', 'odd')
+%tr
+  %td.padding-5{class: klass}
+    = link_to char_name, character_path(char_id)
+  - if local_assigns[:show_template]
```
Throne3d: Since we already ate `local_assigns` from the above `show_template` reference, here too?

teceler: Whoops! yup I'll fix that

PR #1329, opened 2020-02-02
Reload URL hash after accepting TOS if applicable

Throne3d:
This unfortunately has to be done by JavaScript because we submit the form to a POST page in the middle (and the resulting redirect has no access to the browser's fragment value, so it clears it). The checks are so that submitting the ToS on two different pages should still work here, and so it doesn't mess up unrelated page loads if somehow the action fails to complete.

There are very potentially edge cases, but it'd be, like, hitting submit on the TOS page, not loading the page afterwards, and then loading it at some point in the future (where it would then reload the fragment). This should be very rare, though, and only happens once, since it deletes the local storage value afterwards.

Oh, and the header of this TOS page wasn't showing right, so this fixes that too.

### DIFF
--- a/app/assets/javascripts/accept_tos.js
+++ b/app/assets/javascripts/accept_tos.js
@@ -5,4 +5,11 @@ $(document).ready(function() {
     $("#content").hide();
     $("#tos").show();
   }
+  $("#tos_form").submit(function() {
+    /* save fragment (e.g. #reply-1234) for redirect */
+    if (location.hash) {
+      localStorage.setItem("tos.old_path", location.pathname);
+      localStorage.setItem("tos.old_fragment", location.hash);
+    }
+  });
 });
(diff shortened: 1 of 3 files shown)


Throne3d:
It might be somewhat safer to pass it through the form itself, but if so, it'd be because that way you can definitely attach the fragment to a specific form submission. What's "this" in context of being an unsafe variable? `localStorage` is just a way to store data on the client side.

(Though now I look it up, turns out there's a `sessionStorage` object that gets automatically cleared when the page session ends, which might have been better for this!)

I can change this to update a form value.

Throne3d:
Hmm, if I make it a form value, I'm going to have to update the `previous_url` value or modify it in two places, appropriate so it has a hash. Not sure sure how to do that while making sure it's a valid URL – does Rails have particularly flexible helpers for mutating URLs? I think the current approach is easier.

Throne3d:
Session storage should be good! MDN has some information about it: https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage

More or less, it allows the site to store some data locally (temporarily), scoped to the current tab + page, and we can use it to make sure the page anchor is persisted through the TOS form, for paths like <http://localhost:3000/replies/275#reply-275>.

PR #1331, opened 2020-02-02
Set up post writtens

teceler:
(Second part of data migration for #705, after #1351)

### DIFF
--- a/app/concerns/orderable.rb
+++ b/app/concerns/orderable.rb
@@ -36,6 +36,7 @@ module Orderable
       others.each_with_index do |other, index|
         next if other.order == index
         other.order = index
+        other.order += 1 if other.is_a? Reply
         other.save!
       end
     end
@@ -52,6 +53,7 @@ module Orderable
       return unless new_record? || order_change?(false)
       return if new_record? && self.order.present?
       self.order = ordered_items.count
+      self.order += 1 if self.is_a? Reply
     end
 
     def order_change?(is_after)
--- a/app/concerns/orderable.rb
+++ b/app/concerns/orderable.rb
@@ -35,7 +35,7 @@ module Orderable
 
       others.each_with_index do |other, index|
         correct_order = index
-        correct_order += 1 if other.is_a? Reply
+        correct_order += 1 if other.is_a?(Reply) && !other.post.written.present?
         next if other.order == correct_order
         other.update!(order: correct_order)
       end
@@ -53,7 +53,7 @@ module Orderable
       return unless new_record? || order_change?(false)
       return if new_record? && self.order.present?
       self.order = ordered_items.count
-      self.order += 1 if self.is_a? Reply
+      self.order += 1 if self.is_a?(Reply) && !self.post.written.present?
     end
 
     def order_change?(is_after)
--- a/app/concerns/orderable.rb
+++ b/app/concerns/orderable.rb
@@ -34,10 +34,10 @@ module Orderable
       return unless others.present?
 
       others.each_with_index do |other, index|
-        next if other.order == index
-        other.order = index
-        other.order += 1 if other.is_a? Reply
-        other.save!
+        correct_order = index
+        correct_order += 1 if other.is_a? Reply
+        next if other.order == correct_order
+        other.update!(order: correct_order)
       end
     end
 
--- a/app/concerns/orderable.rb
+++ b/app/concerns/orderable.rb
@@ -35,7 +35,7 @@ module Orderable
 
       others.each_with_index do |other, index|
         correct_order = index
-        correct_order += 1 if other.is_a?(Reply) && post.written.blank?
+        correct_order += 1 if other.is_a?(Reply)
         next if other.order == correct_order
         other.update!(order: correct_order)
       end
--- a/app/concerns/orderable.rb
+++ b/app/concerns/orderable.rb
@@ -35,7 +35,7 @@ module Orderable
 
       others.each_with_index do |other, index|
         correct_order = index
-        correct_order += 1 if other.is_a?(Reply) && !other.post.written.present?
+        correct_order += 1 if other.is_a?(Reply) && post.written.blank?
         next if other.order == correct_order
         other.update!(order: correct_order)
       end
@@ -53,7 +53,7 @@ module Orderable
       return unless new_record? || order_change?(false)
       return if new_record? && self.order.present?
       self.order = ordered_items.count
-      self.order += 1 if self.is_a?(Reply) && !self.post.written.present?
+      self.order += 1 if self.is_a?(Reply) && post.written.blank?
     end
 
     def order_change?(is_after)
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -38,6 +38,8 @@ class Post < ApplicationRecord
 
   has_many :notifications, inverse_of: :post, dependent: :destroy
 
+  has_one :written, -> { where(reply_order: 0) }, class_name: 'Reply', inverse_of: :post
+
   attr_accessor :is_import
   attr_writer :skip_edited
 
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -38,7 +38,7 @@ class Post < ApplicationRecord
 
   has_many :notifications, inverse_of: :post, dependent: :destroy
 
-  has_one :written, -> { where(reply_order: 0) }, class_name: 'Reply', inverse_of: :post
+  has_one :written, -> { where(reply_order: 0) }, class_name: 'Reply', inverse_of: :post, dependent: :destroy
 
   attr_accessor :is_import, :skip_written # TODO: delete skip_written after the migration is done
   attr_writer :skip_edited
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -93,7 +93,7 @@ class Post < ApplicationRecord
   # rubocop:enable Style/TrailingCommaInArguments
 
   scope :with_reply_count, -> {
-    select('(SELECT COUNT(*) FROM replies WHERE replies.post_id = posts.id) AS reply_count')
+    select('(SELECT COUNT(*) FROM replies WHERE replies.post_id = posts.id AND replies.reply_order > 0) AS reply_count')
   }
 
   scope :visible_to, ->(user) {
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -40,7 +40,7 @@ class Post < ApplicationRecord
 
   has_one :written, -> { where(reply_order: 0) }, class_name: 'Reply', inverse_of: :post
 
-  attr_accessor :is_import, :skip_written
+  attr_accessor :is_import, :skip_written # TODO: delete skip_written after the migration is done
   attr_writer :skip_edited
 
   validates :subject, presence: true, length: { maximum: 255 }
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -14,7 +14,7 @@ class Post < ApplicationRecord
   belongs_to :last_user, class_name: 'User', inverse_of: false, optional: false
   belongs_to :last_reply, class_name: 'Reply', inverse_of: false, optional: true
   has_one :flat_post, dependent: :destroy
-  has_many :replies, inverse_of: :post, dependent: :delete_all
+  has_many :replies, -> { where('reply_order > 0') }, inverse_of: :post, dependent: :delete_all
   has_many :reply_drafts, dependent: :destroy
 
   has_many :post_viewers, inverse_of: :post, dependent: :destroy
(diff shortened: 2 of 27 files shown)


app/concerns/orderable.rb
```diff
@@ -36,6 +36,7 @@ def reorder_others(is_after)
       others.each_with_index do |other, index|
         next if other.order == index
         other.order = index
+        other.order += 1 if other.is_a?(Reply) && post.written.blank?
```
Throne3d, line 39: Could this use `post.written.nil?` instead? Or is this some kind of builder association? (To confirm my understanding)

app/models/post.rb
```diff
@@ -13,7 +13,7 @@ class Post < ApplicationRecord
   belongs_to :last_user, class_name: 'User', inverse_of: false, optional: false
   belongs_to :last_reply, class_name: 'Reply', inverse_of: false, optional: true
   has_one :flat_post, dependent: :destroy
-  has_many :replies, inverse_of: :post, dependent: :delete_all
+  has_many :replies, -> { where('reply_order > 0') }, inverse_of: :post, dependent: :delete_all
```
Throne3d, line 16: The general strategy here is to make `post.replies` default to ignoring `post.written`, and to update all the `Reply.` queries to explicitly ignore it, right?

app/models/post.rb
```diff
@@ -248,7 +254,7 @@ def has_content_warnings?
 
   def reply_count
     return read_attribute(:reply_count) if has_attribute?(:reply_count)
-    replies.count
+    replies.where.not(reply_order: 0).count
```
Throne3d, line 257: I think this is redundant with the above scope on `replies`

app/models/post.rb
```diff
@@ -332,4 +338,34 @@ def invalidate_caches
...
+      post: self,
+      reply_order: 0,
+      user: user,
+      content: content,
+      icon: icon,
+      character: character,
+      character_alias: character_alias,
+      created_at: created_at,
+      updated_at: edited_at,
+      skip_regenerate: true,
```
Throne3d, line 354: Do we need to worry about `skip_notify`, `is_import`, or `skip_post_update`? (I'm guessing yes, especially on import posts, but not 100% sure.)

app/models/post.rb
```diff
@@ -332,4 +338,34 @@ def invalidate_caches
...
+
+  def update_written
+    return unless written.present?
+    return if (changed_attributes.keys - WRITTEN_ATTRS).empty?
+    written.update!(
+      content: content,
+      icon: icon,
+      character: character,
+      character_alias: character_alias,
+      updated_at: edited_at,
```
Throne3d, line 368: Do we need `skip_post_update` here, too? It looks like it'd be re-triggering this post update if not, recursively.

app/models/reply.rb
```diff
@@ -97,7 +97,7 @@ def notify_other_authors
   end
 
   def previous_reply
-    @prev ||= post.replies.find_by(reply_order: reply_order - 1)
+    @prev ||= post.replies.find_by(reply_order: reply_order - 1) || post.written
```
Throne3d, line 100: I'm not 100% sure we want post.written here, but it seems like we only use it in `reply#delete` (and it'd give 1 here, and defaults to 1), so it's probably okay (right?).

script/create_written.rb
```diff
@@ -0,0 +1,37 @@
...
+          icon: post.icon,
+          character: post.character,
+          character_alias: post.character_alias,
+          created_at: post.created_at,
+          updated_at: post.edited_at,
+        )
+        reply.audits.delete_all
+        audits = post.audits.where("audits.audited_changes ?| array['content', 'icon_id', 'character_id', 'character_alias_id']")
+        audits.each do |audit|
+          reply.audits.create!(
```
Throne3d, line 21: Quickly looking at the schema of an audit, I think we also want to save `username` and `comment` (not sure exactly what `username` is for when we have `user`).

spec/models/post_spec.rb
```diff
@@ -12,11 +12,12 @@
     describe "with no replies" do
       let!(:old_tagged_at) { post.tagged_at }
 
-      it "should update edited_at and tagged_at when edited" do
+      it "should update tagged_at but not edited_at when content edited" do
         Timecop.freeze(time) do
-          post.update!(content: 'new content')
+          post.written.update!(content: 'new content')
```
Throne3d, line 17: When updating a post, we still do it directly on the post contents, don't we? Should this instead check that the `post.written` update is synchronized with the `post` update?

(The change to `edited_at` here is because of the change to which model got updated, right?)

spec/script/create_written_spec.rb
```diff
@@ -0,0 +1,121 @@
+require Rails.root.join('script', 'create_written.rb').to_s
```
Throne3d, line 1: Ooh, this could be a nice way to test more complicated logic our migrations! Nice!

app/models/post.rb
```diff
@@ -31,15 +31,19 @@ class Post < ApplicationRecord
   has_many :indexes, inverse_of: :posts, through: :index_posts, dependent: :destroy
   has_many :index_sections, inverse_of: :posts, through: :index_posts, dependent: :destroy
 
-  attr_accessor :is_import
+  has_one :written, -> { where(reply_order: 0) }, class_name: 'Reply', inverse_of: :post
+
+  attr_accessor :is_import, :skip_written
```
Throne3d, line 36: Can we put a TODO to delete skip_written after the migration, or is that stored in a checklist somewhere?

spec/script/create_written_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+    expect(written.content).to eq(post.content)
+    expect(written.icon_id).to eq(icon.id)
+    expect(written.character_id).to eq(character.id)
+    expect(written.character_alias_id).to eq(calias.id)
+    expect(written.created_at).to be_the_same_time_as(post.created_at)
+    expect(written.updated_at).to be_the_same_time_as(post.edited_at)
+  end
+
+  it "handles audits correctly" do
+    Post.auditing_enabled = true
```
Throne3d, line 55: Can we put these audit things into a `before(:each)` and `after(:each)`, so a failure here keeps the state changes local? (It might actually need a `around(:each)` to ensure it puts it back after a failure, not sure, but I think we've done a similar thing elsewhere with audits)

app/concerns/orderable.rb
Throne3d: Looks like it's a builder association!

app/models/post.rb
teceler: Yeah. I might have missed a few instances because I started by specifically excluding it everywhere, but...

app/models/post.rb
teceler: For example, here! because here doesn't match the post.replies pattern!

app/models/post.rb
teceler: skip_notify seems redundant, since this'll always have the same user as the post, but I'll throw the other two on.

app/models/reply.rb
teceler: I think the reason I did that is so that `set_previous_reply_to_last` doesn't wind up with nil. Possibly it should? Idk.

spec/script/create_written_spec.rb
teceler: Yeah, Marri found an article about it! https://alexharv074.github.io/2016/04/23/rspec-testing-a-simple-ruby-script.html

spec/script/create_written_spec.rb
teceler: Huh. I can do that, yeah.

app/services/reply/searcher.rb
```diff
@@ -25,6 +25,7 @@ def setup(params)
   end
 
   def search(params, page: 1)
+    @search_results = @search_results.where.not(reply_order: 0)
```
redmvc, line 28: Why are we not including the top reply in the search? Wouldn't that be one of the pretty big pluses of doing this at all, is allowing top posts to be searched?

redmvc (approved):

app/concerns/orderable.rb
```diff
@@ -34,9 +34,10 @@ def reorder_others(is_after)
       return unless others.present?
 
       others.each_with_index do |other, index|
-        next if other.order == index
-        other.order = index
-        other.save!
+        correct_order = index
+        correct_order += 1 if other.is_a?(Reply)
```
teceler, line 38: ...this seems probably wrong

app/models/reply.rb
```diff
@@ -6,7 +6,7 @@ class Reply < ApplicationRecord
 
   # define this scope here or Orderable will redefine it
   scope :ordered, -> { order(reply_order: :asc) }
-  scope :ordered_manually, -> { ordered }
+  scope :ordered_manually, -> { where.not(reply_order: 0).ordered }
```
teceler, line 9: 🤔 is this correct?

app/models/post.rb
```diff
@@ -373,4 +379,35 @@ def invalidate_caches
...
+    self.written = Reply.create!(
+      post: self,
+      reply_order: 0,
+      user: user,
+      content: content,
+      icon: icon,
+      character: character,
+      character_alias: character_alias,
+      created_at: created_at,
+      updated_at: edited_at,
```
teceler, line 395: should probably have editor_mode here (& in update_written)

db/schema.rb
```diff
@@ -350,26 +350,26 @@
...
     t.datetime "tagged_at", precision: nil
     t.boolean "authors_locked", default: false
+    t.text "content"
+    t.integer "character_id"
+    t.integer "icon_id"
     t.integer "character_alias_id"
     t.string "editor_mode"
+    t.integer "board_id", null: false
+    t.integer "section_id"
+    t.integer "section_order"
```
teceler: what... happened here

teceler:
Hm i think we also need tests that post#create and post#update handle writtens correctly

app/models/reply.rb
redmvc: I think so? This PR is meant to just create the written and not do anything at all with it, right? So we don't want to include it anywhere that does not currently include the top post.

app/concerns/orderable.rb
redmvc: iirc i added this bc some test was failing without it but idr it's been too many days and the context has been unloaded from my brain

app/models/reply.rb
teceler: Yeah I'm mostly puzzled about that being in ordered_manually but not ordered -- tht's an alias but still

db/schema.rb
redmvc: Uh that was probably bc when I started writing my own version of this and the other PR the first thing I did was get rid of the columns shared with replies in a migration and then I rolled back from that branch to go look at this branch and the order got messed up somehow.

app/concerns/orderable.rb
teceler: probably this is correct *when the migration has been applied but the script hasn't been run yet* but idk how to... parameterize that..,

redmvc: oh wait isn't this from the _other_ PR? #1351?

app/models/reply.rb
redmvc: ....i am also puzzled by this. i'd need to look at it again

db/schema.rb
teceler: These things happen but probably we wanna clean this up xD

Also if so wild that it hit board and section

app/concerns/orderable.rb
teceler: Ahhhhh. Yes that make sense but the problem remains at this stage.

db/schema.rb
redmvc: Probably also because of the migration-and-rollback from the secondary boards PR

app/concerns/orderable.rb
redmvc: this should plausibly not be here at all yeah...

teceler: No, we do still want it before the written is actually created, I think?

redmvc: Do we? What for? I don't really understand what this is doing tbh...

teceler: Okay so the problem is:
- if there is a written, it's at reply_order: 0, we don't need to skip that spot (which is what this is doing, if there's a split or deletion this will get called to make sure everything's in order)
- if there's *not* a written we need to keep reply_order: 0 open or else Problems like the first reply vanishing and trouble when we try to make the written later (or it straightup getting skipped because there 'is one')
- we currently have no way except literally comparing the columns of the first reply and toppost to tell which is true, which *technically* works but is pretty computationally expensive i think?

redmvc: _ah_ okay i understand now

I think possibly the way to solve this is to just have the computationally expensive thing for the deploy and script run, and have another PR ready to deploy fixing this as soon as the script has finished running?

alternatively rather than using a script we just create the written post as part of the migration. would that be really slow/expensive? it's all pure DB operations, without any round trips into ruby, just insert 62 thousand rows into the replies table

teceler: update, autofill_order seems to be handling that part fine, if I remove this line where I get test failures is
1) RepliesController POST restore does not update post status
2) RepliesController POST restore handles first reply deletion
3) RepliesController POST restore handles mid reply deletion
4) RepliesController POST restore handles last reply deletion
5) RepliesController DELETE destroy does not clean up post author when other replies exist
6) RepliesController DELETE destroy reorders correctly without a written
7) RepliesController DELETE destroy reorders correctly
8) Reply#destroy_subsequent_replies works

Interestingly a *similar* set of things fail with ``&& post.written.blank?``

PR #1341, opened 2020-02-09
Rework and test Tag#merge_with

teceler:
(Split out from #1280)


app/models/tag.rb
```diff
@@ -71,19 +71,33 @@ def character_count
...
-      CharacterTag.where(tag_id: other_tag.id).where(character_id: character_tags.select(:character_id).distinct.pluck(:character_id)).delete_all
-      CharacterTag.where(tag_id: other_tag.id).update_all(tag_id: self.id)
-      GalleryTag.where(tag_id: other_tag.id).where(gallery_id: gallery_tags.select(:gallery_id).distinct.pluck(:gallery_id)).delete_all
-      GalleryTag.where(tag_id: other_tag.id).update_all(tag_id: self.id)
-      Tag::SettingTag.where(tag_id: other_tag.id, tagged_id: self.id).delete_all
-      Tag::SettingTag.where(tag_id: self.id, tagged_id: other_tag.id).delete_all
-      Tag::SettingTag.where(tag_id: other_tag.id).update_all(tag_id: self.id)
-      Tag::SettingTag.where(tagged_id: other_tag.id).update_all(tagged_id: self.id)
-      other_tag.destroy
+      unless type == GalleryGroup.to_s
```
Marri, line 77: This is a lot of if statements around `type`; should we be delegating most of this to the subclasses?

PR #1343, opened 2020-02-10
Add labels to characters#edit & characters#new inputs

teceler:


app/views/characters/_editor.haml
```diff
@@ -1,19 +1,19 @@
...
+    %th.sub= f.label :screenname
     %td{class: cycle('even', 'odd')}= f.text_field :screenname, placeholder: "Screen Name", class: 'text'
   %tr
-    %th.sub Setting
+    %th.sub= f.label :setting_ids, 'Setting'
     %td{class: cycle('even', 'odd')}= tag_select(@character, f, :settings)
   %tr
     - klass = cycle('even', 'odd')
-    %th.sub.vtop Template
+    %th.sub.vtop= f.label :template_id
```
Marri: Does this need 'Template' specified?

app/views/characters/_editor.haml
```diff
@@ -22,34 +22,35 @@
         = label_tag :new_template, 'Create New Template'
   - if @character.user == current_user
     %tr#create_template{class: ('hidden' unless params[:new_template])}
-      %th.sub &#8627; Name
-      %td{class: klass}
-        = f.fields_for :template do |ff|
+      = f.fields_for :template do |ff|
+        %th.sub= ff.label :name, '↳ Name'
```
Marri: What are the implications of using the character instead of the HTML &char equivalent?

app/views/characters/_editor.haml
teceler: Nope! Rails generates 'Template' correctly from that.

app/views/characters/_editor.haml
teceler: Either b/c it's in an equals or b/c it's singlequoted, that winds up rendering correctly and the &char doesn't seem to.

Throne3d: It's not HTML-safe as a string, so Rails would re-escape it (`&amp;#8627;`). We could use the old thing with `.html_safe` on the end, but this seems fine.

app/views/characters/_editor.haml
```diff
@@ -22,34 +22,35 @@
...
+        %td{class: klass}
           = ff.text_field :name, placeholder: "Template Name", class: 'text'
+          = ff.hidden_field :id
```
Marri: But this is guaranteed to be nil, isn't it? This is the Create New Template form?

teceler: if you hit 'create instance' on a template, it will take you to this form, with 'id' in the subform. Don't ask me _why_ but that's the behaviour. So I uh... told it to put it here so it didn't try to do it in a way that generated invalid html.

Throne3d (approved):

Throne3d (approved):

PR #1345, opened 2020-02-14
Add post splitter script

Throne3d:
As of this commit, it seems to work in production! Marri also has a branch for post splitting, so I'm expecting this to get superseded, but I still didn't want this branch to get lost, hence this PR.

Would be nice to get a tested post splitter added.


script/post_split.rb
```diff
@@ -0,0 +1,92 @@
+new_subject = 'test subject'
+reply_user = 'Throne3d'
+reply_id = 1024 # we split *after* this reply_id
```
teceler: You probably want a way to set these when launching the script? also why are you setting reply_user instead of just using the reply's user?

script/post_split.rb
```diff
@@ -0,0 +1,92 @@
...
+Post.transaction do
+  user = User.find_by(username: reply_user)
+  abort("needs user") unless user
+  reply = Reply.find_by(user_id: user.id, id: reply_id)
+  abort("couldn't find reply") unless reply
+  old_post = reply.post
+  puts "splitting post #{old_post.id}: #{old_post.subject}, after reply #{reply_id}"
+
+  first_reply = old_post.replies.where('reply_order > ?', reply.reply_order).ordered.first
+  other_replies = old_post.replies.where('reply_order > ?', first_reply.reply_order).ordered
```
teceler: Why _are_ we splitting after the given id, that seems to add unnecessary complexity?

script/post_split.rb
```diff
@@ -0,0 +1,92 @@
...
+  [:character_id, :icon_id, :character_alias_id, :user_id, :content, :created_at, :updated_at].each do |atr|
+    new_value = first_reply.send(atr)
+    puts "new_post.#{atr} = #{new_value.inspect}"
+    new_post.send(atr.to_s + '=', new_value)
+  end
+
+  [:board_id, :section_id, :privacy, :status, :authors_locked].each do |atr|
+    new_value = old_post.send(atr)
+    puts "new_post.#{atr} = #{new_value.inspect}"
+    new_post.send(atr.to_s + '=', new_value)
```
teceler: This... seems like a pretty ugly/messy way to do this?

spec/script/post_split_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+
+    new_post = Post.last
+    expect(new_post.subject).to eq(title)
+    expect(new_post.replies.count).to eq(51)
+    expect(new_post.content).to eq(reply.content)
+    expect(new_post.user_id).to eq(reply.user.id)
+    expect(new_post.authors).to match_array([user, coauthor, new_user])
+    expect(new_post.last_reply_id).to eq(last.id)
+    expect(new_post.last_user_id).to eq(last.user_id)
+    expect(new_post.tagged_at).to eq(last.created_at)
```
Throne3d, line 80: Maybe add a quick check for `new_post.replies.first`, to make sure we don't accidentally duplicate a reply? (We checked the old post, the old post's last reply, and the new post, but not the new post's first reply)

script/post_split.rb
```diff
@@ -0,0 +1,123 @@
...
+  new_post.edited_at = first_reply.updated_at
+  puts "new post: #{new_post.inspect}"
+  new_post.save!
+  puts "new post: https://glowfic.com/posts/#{new_post.id}"
+  new_post
+end
+
+def find_authors(other_replies)
+  author_ids = other_replies.except(:order).select(:user_id).distinct.pluck(:user_id)
+  author_ids.to_h { |id| [id, other_replies.find_by(user_id: id)] }
```
Throne3d, line 47: We're just using this for the `.created_at` value, right? I might return that in the hash directly + use that below, to make this easier to follow (I was wondering why we'd select some random reply for the author 😅 but it turns out this is ordered and we only use the created_at)

teceler: This was in fact your fault, I just refactored it xD

Throne3d: I'm super good at this whole "writing bogus code and coming back to it and realizing it's bogus" thing :D my bad

teceler (approved):
We seem to be in agreement this looks good :sparkles:

PR #1351, opened 2020-02-24
Increment reply order

teceler:
Setup for #705 (and #1331)

### DIFF
--- a/app/concerns/orderable.rb
+++ b/app/concerns/orderable.rb
@@ -36,6 +36,7 @@ module Orderable
       others.each_with_index do |other, index|
         next if other.order == index
         other.order = index
+        other.order += 1 if other.is_a? Reply
         other.save!
       end
     end
@@ -52,6 +53,7 @@ module Orderable
       return unless new_record? || order_change?(false)
       return if new_record? && self.order.present?
       self.order = ordered_items.count
+      self.order += 1 if self.is_a? Reply
     end
 
     def order_change?(is_after)
--- a/app/concerns/orderable.rb
+++ b/app/concerns/orderable.rb
@@ -34,10 +34,10 @@ module Orderable
       return unless others.present?
 
       others.each_with_index do |other, index|
-        next if other.order == index
-        other.order = index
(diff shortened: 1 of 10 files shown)


app/concerns/orderable.rb
```diff
@@ -36,6 +36,7 @@ def reorder_others(is_after)
       others.each_with_index do |other, index|
         next if other.order == index
```
Throne3d, line 37: Should this be:

```ruby
        correct_order = index
        correct_order += 1 if other.is_a? Reply
        next if other.order == correct_order
        other.update!(order: correct_order)
```
? Otherwise I think this `next if` will skip the update incorrectly.

db/migrate/20200202044655_increment_reply_order.rb
```diff
@@ -0,0 +1,5 @@
+class IncrementReplyOrder < ActiveRecord::Migration[5.2]
+  def change
+    Reply.update_all('reply_order = reply_order + 1')
```
Throne3d, line 3: I'm looking at this and a bit concerned that it might lock the table for a long time. I vaguely recall that we might've tested this - did we figure out that this ran quite quickly?

teceler: Yeah, I made it a migration because it ran _so_ fast, even when I gave it a lot of replies. We'll probably need to maintenance briefly, but it shouldn't be very long.

Throne3d: Ah, awesome!

app/concerns/orderable.rb
Throne3d: (Do we have a test for this method?)

Throne3d: Yeah, I think this comment still stands - seems like we're not calculating the index correctly to know we're updating things right here(?)

redmvc (approved):

redmvc (approved):

redmvc (approved):

PR #1358, opened 2020-02-28
Clean up index model tests

teceler:
I kind of... got annoyed here when I was doing cleanup xD. They now test count *changes* and are generally much tidier.

### DIFF
--- a/spec/models/index_spec.rb
+++ b/spec/models/index_spec.rb
@@ -2,38 +2,27 @@ require "spec_helper"
 
 RSpec.describe Index do
   describe "destroy behavior" do
-    let(:index) { create(:index) }
-    let(:section) { create(:index_section, index: index) }
-    let(:post) { create(:post) }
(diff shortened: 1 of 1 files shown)


spec/models/index_spec.rb
```diff
@@ -2,61 +2,37 @@
...
     it "should destroy indexposts and indexsections when deleting an index" do
-      index.destroy
-
-      expect(Index.count).to eq(0)
-      expect(IndexPost.count).to eq(0)
-      expect(IndexSection.count).to eq(0)
-      expect(Post.count).to eq(2)
+      expect {
+        index.destroy!
+      }.to change{Index.count}.by(-1).and change{IndexPost.count}.by(-2).and change{IndexSection}.by(-1).and not_change{Post.count}
```
Marri: Is there a way to make this more legible on multiple lines?

teceler: Having researched instead of just going off what's already in our code, it looks like yes!

PR #1359, opened 2020-03-03
Improve flash messages

teceler:
* Make `flash[:success]` and `flash[:error]` messages consistent between controllers
* Delegate flash error generation to a new `render_errors` method
* Log errors that lack a clear cause

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -226,7 +226,7 @@ class ApplicationController < ActionController::Base
     request.get? && !request.xhr?
   end
 
-  def render_errors(model, action:, now: false, class_name: nil, msg: nil)
+  def render_errors(model, action:, now: false, class_name: nil, msg: nil, err: nil)
     class_name ||= model.class.name.underscore.humanize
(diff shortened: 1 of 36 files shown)


app/controllers/application_controller.rb
```diff
@@ -206,4 +206,33 @@ def check_tos
   def standard_request?
     request.get? && !request.xhr?
   end
+
+  def render_errors(model, action:, now: false, class_name: nil, msg: nil)
+    class_name ||= model.class.name.underscore.humanize
+    msg ||= "#{class_name} could not be #{action}"
```
Throne3d: I think this would go better if `action` were a symbol and you had to pick from a list, so we can be sure that this sentence stitches together well.

Something like:

```rb
VALID_ERROR_ACTIONS = [:created, :updated, :deleted]
```

Additonally, this might actually be one place where we _do_ want to use Rails's i18n library, because this kind of stitching often happens when you have to do translations. I'm not sure – I'd have to investigate – but I'm a bit concerned we'll want to refactor this again relatively soon.

PR #1362, opened 2020-03-15
Button to toggle retired view in the character list

Marri:
Might add an account option for this later as well, applicable only to your own characters, but I expect that will rabbit hole me into actually separating account settings into more sensible tables, so here we are. Only open question is, is the fact that a character is considered 'retired' public information or not?

Incidentally fixes a colspan bug, lack of distinct coloring for .view-button.selected in Starry Light layouts, and an overzealous spacer tr.


app/assets/stylesheets/layouts/starrylight.scss
```diff
@@ -73,7 +73,7 @@ $bg_color_linkbox_subber: $colors_wildblueyonder;
 
 // other Buttons
 $bg_color_viewbutton: $colors_scampi;
-$bg_color_viewbutton_selected: $colors_scampi;
+$bg_color_viewbutton_selected: $colors_wildblueyonder;
```
teceler, line 76: What's with this color change?

app/helpers/character_helper.rb
```diff
@@ -12,4 +12,9 @@ def characters_list(characters, show_template)
     attributes += ['templates.id', 'templates.name'] if show_template
     characters.joins(:user).pluck(*attributes)
   end
+
+  def character_menu_link(link_params)
+    link_params = params.permit(:character_split, :retired, :view).to_h.merge(link_params)
+    url_for(**link_params.symbolize_keys)
+  end
```
teceler, line 19: So many parameters. A helper is a good idea xD

app/views/characters/_list_section.haml
```diff
@@ -1,6 +1,7 @@
 - col_count = 6
 - col_count += 1 if local_assigns[:show_template]
 - characters = name.characters.ordered unless local_assigns[:characters]
+- characters = characters.where(retired: false) if params[:retired] == 'false'
```
teceler, line 4: We really need some kinda shared thing for this shared config between icon and list view, but out of scope for this

app/views/characters/index.haml
```diff
@@ -4,9 +4,10 @@
     &raquo;
     %b #{@user.username}'s Characters
 
+- colspan = (page_view == 'list') ? (character_split == 'none' ? 7 : 6) : 1
```
teceler, line 7: that... is a nested... ternary. yikes? I didn't even know you could do that?

app/views/characters/index.haml
```diff
@@ -46,8 +54,9 @@
         = hidden_field_tag :search_screenname, true
         = hidden_field_tag :search_nickname, true
         = submit_tag "Search", class: 'button'
-  %tr
-    %th.odd{style: 'padding: 3px', colspan: colspan}
+  - if character_split == 'template' # spacer between two .subbers
+    %tr
+      %th.odd{style: 'padding: 3px', colspan: colspan}
```
teceler, line 59: ...huh. was that getting included even on non-split? should it not be?
also should possibly go in the... regular 'if view split' right below it

app/assets/stylesheets/layouts/starrylight.scss
Marri: "Incidentally fixes a...lack of distinct coloring for .view-button.selected in Starry Light layouts" < this bit. In the process of testing my various new character link options for the character_menu_link I noticed that, because this color was originally the same as the one above it, the .selected class wasn't actually doing anything visible at all. It is now visually distinct!

app/views/characters/index.haml
Marri: it should not be, as the comment says, in both views there's a search bar at the top with .subber but only in template view would this then immediately be followed by a template name header _also_ with .subber which was confusing and so the spacer separates them; non-split doesn't have the template .subber sections so it doesn't need the spacer.

app/assets/stylesheets/layouts/starrylight.scss
teceler: Aha, yeah, I missed that. Cool.

PR #1363, opened 2020-03-21
Timezoneify timestamps for logged out users

Throne3d:
Also affects flat posts - we cache the timestamps there, so I'm improving this to update them with JavaScript. For now, we make sure the string is very similar to our default format (changing from `Jun 22, 2019 4:38 AM` to `Jun 22, 2019, 4:38 AM`: note the comma), and always in en-US locale, to cohere better with the site, but we could reinvestigate this if we wanted.

Additionally, where we were using `pretty_time` to display timestamps, I've updated _all_ of them (even when logged in) to be wrapped in a `<time>` tag and to additionally show the UTC timestamp on hover.

On a technical level, I use a `<script>` tag with the `defer` attribute to try to reduce the amount the page flickers on load. However, other components (like Select2) already cause page flickering, so I don't think this is a significant consideration.

Note: after this change, we should regenerate flat posts so historical flat posts can take advantage of this feature.

### DIFF
--- a/app/assets/javascripts/datetimes.js
+++ b/app/assets/javascripts/datetimes.js
@@ -1,0 +1,15 @@
+/* global gon */
+var options = {
+  dateStyle: 'medium',
+  timeStyle: 'short',
+};
+var format = Intl.DateTimeFormat('en-US', options);
+
+window.onload = function() {
+  if (gon.logged_in) return;
+  var times = document.getElementsByTagName('time');
+  Array.from(times).forEach(function(time) {
+    var datetime = new Date(time.dateTime);
+    time.innerText = format.format(datetime);
+  });
+};
--- a/app/assets/javascripts/application_deferred.js
+++ b/app/assets/javascripts/application_deferred.js
@@ -1,0 +1,1 @@
+//= require datetimes
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -15,7 +15,9 @@ module ApplicationHelper
 
   def pretty_time(time, format: nil)
     return unless time
-    time.strftime(format || current_user.try(:time_display) || TIME_FORMAT)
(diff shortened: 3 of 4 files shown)


app/assets/javascripts/datetimes.js
```diff
@@ -0,0 +1,15 @@
+/* global gon */
+var options = {
+  dateStyle: 'medium',
+  timeStyle: 'short',
+};
+var format = Intl.DateTimeFormat('en-US', options);
+
+window.onload = function() {
+  if (gon.logged_in && !gon.override_times) return;
```
Marri, line 9: :fire: an extra override_times here

Throne3d: Fixed!

teceler:
Okay, thoughts:
* The javascript is to update old flatposts on load? Or is it additionally doing something else?
* I'm not actually sure why you've reformatted pretty_time calls but it's not a big deal

Throne3d:
> The javascript is to update old flatposts on load? Or is it additionally doing something else?

It also works on flatposts on load, yeah! But the main thing is that, when a user is logged out, we don't have their timezone saved on the user object, and we can't get it via HTTP - we have to use something client-side to find the timezone to use for logged-out users.

Actually, thinking about it, I think this might only work on flat posts for logged-out users right now. I'll test this and check - if it works on flat posts for logged in users, that would be good!

> I'm not actually sure why you've reformatted pretty_time calls but it's not a big deal

It should still use the same display that users selected in their profile if available! It just also wraps it inside a `<time datetime="2022-02-16T22:21Z" title="2022-02-16 22:21 UTC">` element too, now, so that you can check what the real actual timestamp is even if the timezone that's displaying isn't the timezone you expected.

Throne3d:
Fixed flat posts - they weren't doing the timestampification correctly - including for logged-in users.

Throne3d:
Removed the JavaScript from flat posts due to rendering concerns!

PR #1364, opened 2020-03-21
Improve flash messages, centralizing strings under locales/

Throne3d:
Builds on #1359, but uses I18n to build the strings instead. This way, we use symbols to indicate what failed, and translate these to human-readable strings in a central location!


app/controllers/concerns/error_rendering.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  end
+
+  def render_err(chain=:chain, *args, **kwargs)
+    if chain == :chain
+      ErrorRenderer.new(flash, self)
+    else
+      render_err_target(flash, chain, *args, **kwargs)
+    end
+  end
```
teceler, line 19: Quick question, what... is even going on here.

Throne3d: clearly me-of-three-years-ago was having too much fun writing the least clear code possible

I'll take another look at this in a bit

app/controllers/concerns/translations.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    str = t(
+      # pick up a model-specific translation if available
+      t_sym_for(model_name.parameterize.underscore, status, :messages, key),
+      model_name: model_name,
+      scope: [:actioncontroller],
+      # fall back to the basic translation otherwise
+      default: t_sym_for(status, :messages, key),
+    )
```
teceler, line 32: Still unclear on what this call is doing

app/controllers/concerns/translations.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    # control the overall message format via :intro or :final
+    # :final - "<Error>."
+    # :intro - "<Error>, because of the following problems:")
+    t(
+      format,
+      str: str,
+      scope: [:actioncontroller, status, :formats],
+    )
```
teceler, line 41: Or this one for that matter, or why you're calling t() twice, once with the output of the first time

PR #1365, opened 2020-04-04
Reading a post marks its notification read

amandaschloss:
And include read_at because why weren't we doing that already?

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -77,7 +77,7 @@ class MessagesController < ApplicationController
       @messages.each do |m|
         next unless m.unread?
         next unless m.recipient_id == current_user.id
-        m.update(unread: false, read_at: Time.now.in_time_zone)
+        m.update(unread: false, read_at: Time.zone.now)
       end
     end
     @message = Message.new
@@ -95,7 +95,7 @@ class MessagesController < ApplicationController
     if params[:commit] == "Mark Read"
       messages.each do |message|
         next unless message.recipient_id == current_user.id
-        message.update(unread: false, read_at: Time.now.in_time_zone)
+        message.update(unread: false, read_at: Time.zone.now)
       end
     elsif params[:commit] == "Mark Unread"
       messages.each do |message|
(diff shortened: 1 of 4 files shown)


app/controllers/messages_controller.rb
```diff
@@ -107,7 +107,7 @@ def mark
         box_attr = "visible_#{box}"
         user_id_attr = (box == 'inbox') ? 'recipient_id' : 'sender_id'
         Message.where(thread_id: message.thread_id, "#{user_id_attr}": current_user.id).each do |thread_message|
-          thread_message.update(box_attr => false, unread: false)
+          thread_message.update(box_attr => false, unread: false, read_at: Time.zone.now)
```
teceler, line 110: Slightly confused by what's going on with these arguments

app/jobs/notify_followers_of_new_post_job.rb
```diff
@@ -44,21 +44,21 @@ def notify_of_post_joining(post, new_user)
...
-      return true if notification.message.include?(view_post(post.id))
-    end
-    false
+    self.class.notification_about(post, user).present?
   end
 
-  def view_post(post_id)
-    host = ENV['DOMAIN_NAME'] || 'localhost:3000'
-    url = Rails.application.routes.url_helpers.post_url(post_id, host: host, protocol: 'https')
-    "<a href='#{url}'>View it here</a>."
```
teceler, line 62: Was this duplicated on ScrapePostJob? I'm not sure that's the best place to centralize if it was but otherwise this looks fine if so.

Marri: Yeah, it was duplicated. Didn't care too much about where to put it because Someday Notifications and we won't need it.

app/controllers/messages_controller.rb
Marri: We had a database column for "what time did the user read this Message" and we weren't actually storing that information when they read it? Possibly because of the threading thing, hmmm. Maybe I need a ||= here.

PR #1367, opened 2020-04-14
Redo Post::Status to use the Rails enum builtins

Marri:
Open question which I have absolutely used mixed values for here and should consistentize: do we want to use:
`post.update(status: 'complete')`
or
`post.update(status: :complete)`
or
`post.update(status: Post.statuses[:complete])`

I lean towards 2 but am open to opinions.

I believe this fixes an incidental bug in the post metadata page around how we display posts that are active-but-old-enough-to-display-hiatuses, but perhaps that was on purpose?

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -28,13 +28,13 @@ class PostsController < WritableController
       solo = PostAuthor.where(post_id: ids).group(:post_id).having('count(post_id) < 2').pluck(:post_id)
       @posts = @posts.where.not(last_user: current_user).or(@posts.where(id: (drafts + solo).uniq))
     end
-    @posts = @posts.where.not(status: [Post::Status::COMPLETE, Post::Status::ABANDONED])
-    hiatused = @posts.where(status: Post::Status::HIATUS).or(@posts.where('tagged_at < ?', 1.month.ago))
+    @posts = @posts.where.not(status: [:complete, :abandoned])
+    hiatused = @posts.hiatus.or(@posts.where('tagged_at < ?', 1.month.ago))
 
     if params[:view] == 'hiatused'
       @posts = hiatused
     elsif current_user.hide_hiatused_tags_owed?
-      @posts = @posts.where.not(status: Post::Status::HIATUS).where('tagged_at > ?', 1.month.ago)
+      @posts = @posts.where.not(status: :hiatus).where('tagged_at > ?', 1.month.ago)
       @hiatused_exist = true if hiatused.count > 0
     end
 
@@ -264,7 +264,7 @@ class PostsController < WritableController
     if params[:subject].present?
       @search_results = @search_results.search(params[:subject]).where('LOWER(subject) LIKE ?', "%#{params[:subject].downcase}%")
     end
-    @search_results = @search_results.where(status: Post::Status::COMPLETE) if params[:completed].present?
+    @search_results = @search_results.complete if params[:completed].present?
     if params[:author_id].present?
       post_ids = nil
       params[:author_id].each do |author_id|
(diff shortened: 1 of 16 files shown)


app/controllers/posts_controller.rb
```diff
@@ -346,25 +346,23 @@ def mark_hidden
...
+        @post.update!(status: params[:status])
+        @post.mark_read(current_user, @post.tagged_at)
       end
+    rescue ActiveRecord::RecordInvalid
+      flash[:error] = {
+        message: "Status could not be updated.",
+        array: @post.errors.full_messages
+      }
+    else
+      flash[:success] = "Post has been marked #{@post.status}."
```
Marri, line 365: Only changed this because I (incorrectly) thought we were submitting integers here instead of strings. Makes literally no difference, they're the same.

app/models/post.rb
```diff
@@ -192,7 +192,7 @@ def read_time_for(viewing_replies)
 
     # testing for case where the post was changed in status more recently than the last reply
     audits_exist = audits.where('created_at > ?', most_recent.created_at).where(action: 'update')
-    audits_exist = audits_exist.where("(audited_changes -> 'status' ->> 1)::integer = ?", Post::Status::COMPLETE)
+    audits_exist = audits_exist.where("(audited_changes -> 'status' ->> 1)::integer = ?", Post.statuses[:complete])
```
Marri, line 195: This is one of the cases I expect must validly remain calling into the hash for the integer value, given the SQL givens

app/models/post/status.rb
```diff
@@ -1,36 +1,16 @@
...
-      status == ACTIVE
-    end
-
-    def abandoned?
-      status == ABANDONED
+      hiatus? || (active? && tagged_at < 1.month.ago)
     end
   end
-
-  def self.get_status(param)
```
Marri, line 31: We didn't seem to be using this anywhere? Did I miss it?

app/services/post_importer.rb
```diff
@@ -3,7 +3,7 @@ def initialize(url)
     @url = url
   end
 
-  def import(board_id, importer_id, section_id: nil, status: Post::Status::COMPLETE, threaded: false)
+  def import(board_id, importer_id, section_id: nil, status: Post.statuses[:complete], threaded: false)
```
Marri, line 6: I left the post scraper/importer code alone for now and let it keep using integers; possibly this should be refactored to use strings/symbols eventually? Do we care enough that I should do that refactoring now?

app/views/posts/stats.haml
```diff
@@ -28,17 +28,16 @@
     %tr
       %th.sub.width-150 Status
       %td{class: cycle('odd', 'even')}
-        - case @post.status
```
Marri, line 31: Was this bug intentional? Past me, why this?

app/models/post/status.rb
teceler: I think when status is set?

app/controllers/posts_controller.rb
```diff
@@ -346,25 +346,23 @@ def mark_hidden
   end
 
   def change_status
-    begin
-      new_status = Post::Status.get_status(params[:status].upcase)
```
teceler, line 350: this is where we were using get_status, I think. I'd grep the code for any other instances but you might have obsoleted it yeah

app/controllers/posts_controller.rb
```diff
@@ -264,7 +264,7 @@ def search
     if params[:subject].present?
       @search_results = @search_results.search(params[:subject]).where('LOWER(subject) LIKE ?', "%#{params[:subject].downcase}%")
     end
-    @search_results = @search_results.where(status: Post::Status::COMPLETE) if params[:completed].present?
+    @search_results = @search_results.complete if params[:completed].present?
```
teceler, line 267: ...did you make _scopes_ for the statuses?

app/controllers/posts_controller.rb
teceler: I had a PR to submit ints instead of strings but you went 'um this seems suboptimal'. that might be what you were thinking of, though.

app/models/post.rb
teceler: ~SQL~

app/controllers/posts_controller.rb
teceler: It looks like no unless there is Rails Magic going on here, this looks like it will do... something weird...

app/controllers/posts_controller.rb
```diff
@@ -28,13 +28,13 @@ def owed
       solo = PostAuthor.where(post_id: ids).group(:post_id).having('count(post_id) < 2').pluck(:post_id)
       @posts = @posts.where.not(last_user: current_user).or(@posts.where(id: (drafts + solo).uniq))
     end
-    @posts = @posts.where.not(status: [Post::Status::COMPLETE, Post::Status::ABANDONED])
-    hiatused = @posts.where(status: Post::Status::HIATUS).or(@posts.where('tagged_at < ?', 1.month.ago))
+    @posts = @posts.where.not(status: [:complete, :abandoned])
+    hiatused = @posts.hiatus.or(@posts.where('tagged_at < ?', 1.month.ago))
```
teceler, line 32: Same problem with 'what even are you doing here' as below

app/controllers/posts_controller.rb
Marri: There is Rails Magic that makes scopes for the statuses

PR #1371, opened 2020-04-17
Add a button for Only New Today to the daily report

Marri:
Also incidentally fix a bug where if a post was started on day A and possessed replies- none of which were created on day A- it will not appear in the report.

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -27,7 +27,7 @@ class ReportsController < ApplicationController
     end
 
     if @report_type == 'daily'
-      @new_today = (params[:new_today]=='true')
+      @new_today = params[:new_today].present?
       @posts = DailyReport.new(@day).posts(sort, @new_today)
       @posts = posts_from_relation(@posts, max: !@new_today)
       replies_on_day = Reply.where(created_at: @day.beginning_of_day..@day.end_of_day)
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -27,8 +27,9 @@ class ReportsController < ApplicationController
     end
 
     if @report_type == 'daily'
-      @posts = DailyReport.new(@day).posts(sort)
-      @posts = posts_from_relation(@posts, max: true)
+      @new_today = (params[:new_today]=='true')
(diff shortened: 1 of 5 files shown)


app/controllers/reports_controller.rb
```diff
@@ -27,8 +27,9 @@ def show
     end
 
     if @report_type == 'daily'
-      @posts = DailyReport.new(@day).posts(sort)
-      @posts = posts_from_relation(@posts, max: true)
+      @new_today = (params[:new_today]=='true')
```
teceler, line 30: Why `=='true'`?

app/services/daily_report.rb
```diff
@@ -5,11 +5,13 @@ def initialize(day)
     @day = day
   end
 
-  def posts(sort='')
+  def posts(sort='', new_today=false)
     range = day.beginning_of_day .. day.end_of_day
-    created_no_replies = Post.where(last_reply_id: nil, created_at: range)
+    created_today = Post.where(created_at: range)
+    return created_today.select("posts.*, posts.created_at as first_updated_at").order(sort) if new_today
```
teceler, line 11: Why are you aliasing that?

teceler: ...ah, because the select is also pulling replies.created_at later, cool

app/services/daily_report.rb
```diff
@@ -21,8 +23,7 @@ def posts(sort='')
         range.begin,
         range.end,
       ]))
-      .joins("LEFT JOIN replies AS replies_today ON replies_today.post_id = posts.id")
-      .where("replies_today.created_at IS NULL OR (replies_today.created_at between ? AND ?)", range.begin, range.end)
+      .joins("LEFT JOIN replies AS replies_today ON replies_today.post_id = posts.id AND replies_today.created_at between '#{range.begin}' AND '#{range.end}'")
```
teceler, line 26: Can joins not do the substitution thing? I realize this is probably safe in context but

app/services/daily_report.rb
Marri: (also default sort uses first_updated_at)

app/services/daily_report.rb
Marri: Nope, because Rails hates us and wants us to suffer.

PR #1379, opened 2020-04-18
Don't use autogen route because it's wrong now

Marri:


Marri:
For posterity: yes, board#new specifies the URL already

PR #1381, opened 2020-04-19
Let TemplateController have access to the character_split method

Marri:
@teceler ?


teceler:
That is a TINY concern. We might wanna rename it slightly and use it for shared template/character methods in general? I'm sure there will be more of those.

Otherwise, looks good, thank you :green_heart:. (Codeclimate you can shut up.)

Marri:
Yeah the CodeClimate complaint isn't new, it's just relocated with the code.

PR #1384, opened 2020-04-21
Authentication

Marri:
Use proper token based API authentication!
- [x] Finish adding tokens to all API calls in app/assets/javascripts
- [x] Finish fixing all API tests to use api_login (:sob:)
- [x] Handle expired tokens
- [x] Is this how I want to handle edge cases with tokens? Missing, invalid, expired, etc? (see #1400)

Historical note: Guided a lot by this https://www.codementor.io/@omedale/simple-approach-to-rails-5-api-authentication-with-json-web-token-cpqbgrdo6


app/assets/javascripts/global.js
```diff
@@ -172,13 +172,38 @@ function createTagSelect(tagType, selector, formType, scope) {
...
   if (!options.width) { options.width = '100%'; }
-
-  if (options.ajax) {
-    options.ajax.delay = 200;
-    options.ajax.dataType = 'json';
-    options.ajax.cache = true;
-    if (!options.ajax.data) { options.ajax.data = queryTransform; }
-  }
-
+  if (options.ajax) addAjaxOptions(options);
```
teceler, line 175: what... is this doing

Marri: this is making code climate shut up, mostly. or did you mean what are the options themselves doing?

teceler: I'm asking... uh.... why is there an if that doesn't have brackets and a method call on the same line and... stuff

Marri: Oh. Javascript allows single line bracketless ifs. Do we not use that anywhere else, if that's a style thing I can make it match everything else.

Marri: https://github.com/Marri/glowfic/issues/1435 but leaving it for now.

PR #1392, opened 2020-04-25
Fix summary display

### LINKED ISSUE #1390: Make summary/details show consistently between browsers
Firefox currently does not show any indication that summaries can be clicked for details.

In firefox: 
image

In chrome: 
image

teceler:
Firefox does Dumb Things with how `display: block` affects summary 
elements, which we do not want it to do, therefore I have told normalize 
not to set it on summaries. See 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary#Default_style

Fixes #1390

I recognize that this edits normalize, which is why I commented the relevant line rather than deleting it, but ~~our normalize is really outdated anyway~~ I think this is important to fix

### DIFF
--- a/app/assets/stylesheets/normalize.scss
+++ b/app/assets/stylesheets/normalize.scss
@@ -30,6 +30,8 @@ body {
  * Correct `block` display not defined for `main` in IE 11.
  */
 
+/* summary removed from this list due to display: block causing issues with some browsers. */
+
 article,
 aside,
 details,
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/normalize.scss
```diff
@@ -42,7 +42,7 @@ main,
 menu,
 nav,
 section,
```
Throne3d, line 44: This trailing comma is invalid CSS =( Would we be able to comment it out too?

teceler: probably?

PR #1391, opened 2020-04-25
Update gems for April 25, 2020

teceler:
Major
- [x] Updates rack-mini-profiler from 1.1.4 to 2.0.1 ([Changelog](https://github.com/MiniProfiler/rack-mini-profiler/blob/master/CHANGELOG.md#201---2020-03-17)) ([Upgrade info](https://github.com/MiniProfiler/rack-mini-profiler#upgrading-to-version-200))
- [x] Updates rspec-rails from 3.9.0 to 4.0.0 ([Changelog](https://github.com/rspec/rspec-rails/blob/master/Changelog.md#400--2020-03-24))

Minor
- [x] Updates aws-sdk-s3 from 1.55.0 to 1.63.0 ([Changelog](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-s3/CHANGELOG.md#1630-2020-04-22))
- [x] Updates aws-sdk-ses from 1.27.0 to 1.28.0 ([Changelog](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-ses/CHANGELOG.md#1280-2020-03-09))
- [x] Updates bootstrap from 4.3.1 to 4.4.1 ([Diff](https://github.com/twbs/bootstrap-rubygem/compare/v4.3.1...v4.4.1))
- [x] Updates byebug from 11.0.1 to 11.1.3 ([Changelog](https://github.com/deivid-rodriguez/byebug/blob/master/CHANGELOG.md#1113---2020-04-23))
- [ ] Updates capybara from 3.29.0 to 3.32.1 ([Changelog](https://github.com/teamcapybara/capybara/blob/master/History.md#version-3321))
- [ ] Updates database_cleaner from 1.7.0 to 1.8.4 ([Changelog](https://github.com/DatabaseCleaner/database_cleaner/blob/master/History.rdoc#184-2020-04-02-))
- [ ] Updates httparty from 0.17.1 to 0.18.0 ([Changelog](https://github.com/jnunemaker/httparty/blob/master/Changelog.md#0180))
- [ ] Updates newrelic_rpm from 6.7.0.359 to 6.10.0.364 ([Changelog](https://github.com/newrelic/rpm/blob/master/CHANGELOG.md#v6100))
- [ ] Updates nilify_blanks from 1.3.0 to 1.4.0 ([Changelog](https://github.com/rubiety/nilify_blanks/blob/master/CHANGELOG.rdoc#2020-04-18-140-)) (broke user-spec tests)
[...]

PR #1396, opened 2020-04-26
Update to Rails 6

teceler:
~~Depends on #1391 and #1395~~

https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-2-to-rails-6-0


.codeclimate.yml
```diff
@@ -2,7 +2,8 @@
 version: "2"
 plugins:
   brakeman:
-    enabled: true
+    enabled: false
+    # until they fix rails 6 handling
```
Throne3d, line 6: is this fixed now?

Gemfile.lock
```diff
@@ -1,53 +1,66 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    actioncable (5.2.6)
```
Throne3d, line 4: codeclimate is complaining about some gems - I guess we can rebase this onto the gem upgrade once we merge that and then make sure we're on the latest patch versions?

.codeclimate.yml
teceler: I can re-enable it to check but signs point to no <https://docs.codeclimate.com/docs/brakeman>

Gemfile.lock
teceler: If this is the same issue as I had on a previous patch version I think it's codeclimate being confused by the 6.0.4.x versions or something because they were CVEs that were fixed

Throne3d (approved):
LGTM

app/helpers/post_helper.rb
```diff
@@ -1,6 +1,6 @@
 module PostHelper
   def author_links(post, linked: true, colored: false)
-    total = post.authors.size
+    total = post.authors.length
```
Throne3d, line 3: (Relation no longer has size?)

teceler: I think it has size but size does something super weird on this method in particular? I remember a whole adventure finding that bug

Marri: https://stackoverflow.com/questions/6083219/activerecord-size-vs-count size tries to be intelligent about "use array size in memory if queryset is loaded, call db for count if not" but if our db and memory counts differ that could be the problem? If not that I am short term out of ideas.

config/environments/test.rb
```diff
@@ -21,12 +22,17 @@
   # Show full error reports and disable caching.
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
+  config.cache_store = :null_store
```
Marri, line 25: I know we've had cache adventures with tests, is this in fact what we want?

Marri (approved):
I am not thrilled about losing brakeman so we may want to make a pass at fixing brakeman issues before we release this but it otherwise looks reasonably sane? Just want to check on the cache tests first.

config/environments/test.rb
teceler: ...yeah, that is in fact outdated, good catch :turkey:

Gemfile.lock
teceler: Yeah this passes for bundler-audit locally so I think it's fine and just a codeclimate issue

app/helpers/post_helper.rb
teceler: Reloading the posts in tests before calling this seems to fix it, yeah! Little concerned about why this is a problem with the upgrade in case it affects anything else, but.

.codeclimate.yml
teceler: It doesn't look like this will be fixed anytime soon but brakeman does apparently have github actions options we could try instead if we want to keep it, it seems to be a codeclimate problem not a brakeman problem.

Throne3d (approved):

app/helpers/post_helper.rb
Throne3d: Weird! We investigated this in the Discord and couldn't figure out exactly what's going on, but it seems like a potential Rails bug, and doesn't seem to impact our use cases.

Throne3d:
LGTM! Seems to be working well on staging :D

PR #1398, opened 2020-04-29
Live-in-page editor for private post notes

Marri:
People should not go to edit their post notes and discover that they have lost everything they've written.

I considered a popup option to keep the editor more visibly separate from the reply editor. This was immediately vetoed by at least three people and was discarded as an option. But like, for posterity, if folks consider it.

~~Awaits #1384 so that we can properly authenticate the "save post notes" call in the page without reloading.~~

To do:
- [x] Show loading image after clicking Save
- [x] Show success image after successful save
- [x] Show error image if failed save
- [x] Some way to close the editor if you change your mind
- [x] Some way to discard your changes, with a confirmation
- [x] Better titles (and alt text?) on necessary icons
- [x] Generate API docs
- [x] Finish spec tests
- [x] All available error codes in API docs
- [x] Rename Post Notes to Author Notes
- [x] Some kind of warning before submitting or previewing reply with unsaved note

### DIFF
--- a/spec/features/replies/edit_spec.rb
+++ b/spec/features/replies/edit_spec.rb
@@ -34,7 +34,7 @@ RSpec.feature "Creating replies", :type => :feature do
     expect(page).to have_no_selector('.post-container')
     within('#post-editor') do
       fill_in 'reply_content', with: 'other text'
-      click_button 'Save'
+      click_button 'submit_button'
     end
 
     expect(page).to have_no_selector('.error')
@@ -78,7 +78,7 @@ RSpec.feature "Creating replies", :type => :feature do
     within('#post-editor') do
       expect(page).to have_field('reply_content', with: 'other text')
       fill_in 'reply_content', with: 'third text'
-      click_button 'Save'
+      click_button 'submit_button'
     end
 
     expect(page).to have_no_selector('.error')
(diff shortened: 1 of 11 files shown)


spec/features/replies/edit_spec.rb
```diff
@@ -34,7 +34,7 @@ def find_reply_on_page(reply)
     expect(page).to have_no_selector('.post-container')
     within('#post-editor') do
       fill_in 'reply_content', with: 'other text'
-      click_button 'Save'
+      click_button 'submit_button'
```
teceler, line 37: 'Save' is actually clearer as to what the button is doing, did something go wrong with that?

Marri: We now have two save buttons on the page, Save-but-save-the-reply and Save-but-save-the-post-note (just hidden) so this was raising AmbiguousMatch. I did not see a helpful container element that I could use to avoid this problem with a `within`, so I used the CSS id which was already set. I could see about changing the existing CSS id to something slightly less opaque instead if that works better?

teceler: That makes sense! And yeah a less opaque id might be good if that's not too much extra work.

PR #1400, opened 2020-05-03
API token parsing error responses

Marri:
Right now invalid tokens are treated the same as being logged out, but we have endpoints that are handled differently depending on your account state (e.g. post#show will fail if the post isn't visible to you), so we should provide users feedback if they provide a token we can't authenticate as a distinct state from simply not providing a token and operating as a logged out user.

### DIFF
--- a/app/concerns/authentication/api.rb
+++ b/app/concerns/authentication/api.rb
@@ -36,7 +36,7 @@ module Authentication::Api
       HashWithIndifferentAccess.new(body)
     rescue JWT::ExpiredSignature
       error = {message: "Authorization token has expired."}
-      render json: {errors: [error]}, status: :unprocessable_entity
+      render json: {errors: [error]}, status: :unauthorized
       return {}
     rescue JWT::DecodeError
       error = {message: "Authorization token is not valid."}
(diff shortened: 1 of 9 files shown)


app/concerns/authentication/api.rb
```diff
@@ -34,8 +34,14 @@ def decode_api_token
     def decode_value(value)
       body = JWT.decode(value, Rails.application.secrets.secret_key_api)[0]
       HashWithIndifferentAccess.new(body)
+    rescue JWT::ExpiredSignature
+      error = {message: "Authorization token has expired."}
+      render json: {errors: [error]}, status: :unprocessable_entity
```
Marri, line 39: I'm open to returning a 401 here instead

spec/controllers/api/v1/sessions_controller_spec.rb
```diff
@@ -9,17 +9,6 @@
       expect(response.json['errors'][0]['message']).to eq("You must be logged out to call this endpoint.")
     end
 
-    it "counts token expiration as logout" do
```
Marri, line 12: Moved to the api_controller_spec

PR #1403, opened 2020-05-05
Improve reply duplication check

### LINKED ISSUE #1399: Check that new reply =/= last when interrupting for unseen replies
Because internet hiccups happen and apparently unseen replies currently trumps 'hey, you already posted this'

teceler:
Fixes #1399

* Checks for duplicate reply before unseen replies
* Does not save a draft for duplicate replies unless there are other unseen replies present
* Cleans up code slightly

### DIFF
--- a/app/controllers/replies_controller.rb
+++ b/app/controllers/replies_controller.rb
@@ -106,6 +106,11 @@ class RepliesController < WritableController
     reply.user = current_user
 
     if reply.post.present?
+      last_seen_reply_order = reply.post.last_seen_reply_for(current_user).try(:reply_order)
+      @unseen_replies = reply.post.replies.ordered.paginate(page: 1, per_page: 10)
+      @unseen_replies = @unseen_replies.where('reply_order > ?', last_seen_reply_order) if last_seen_reply_order.present?
+      most_recent_unseen_reply = @unseen_replies.last
+
       if reply.user_id.present? && params[:allow_dupe].blank?
         last_by_user = reply.post.replies.where(user_id: reply.user_id).ordered.last
         if last_by_user.present?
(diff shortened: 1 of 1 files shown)


app/controllers/replies_controller.rb
```diff
@@ -110,6 +110,23 @@ def create
...
       @unseen_replies = @unseen_replies.where('reply_order > ?', last_seen_reply_order) if last_seen_reply_order.present?
       most_recent_unseen_reply = @unseen_replies.last
+
+      if params[:allow_dupe].blank?
+        last_by_user = reply.post.replies.where(user_id: reply.user_id).ordered.last
+        match_attrs = ['content', 'icon_id', 'character_id', 'character_alias_id']
+        if last_by_user.present? && last_by_user.attributes.slice(*match_attrs) == reply.attributes.slice(*match_attrs)
+          flash.now[:error] = "This looks like a duplicate. Did you attempt to post this twice? Please resubmit if this was intentional."
+          @allow_dupe = true
+          if @unseen_replies.count == 0 || (@unseen_replies.count == 1 && most_recent_unseen_reply.id == last_by_user.id)
```
Marri, line 120: Why does this use `.count`? On a relation that'll trigger a new query, which seems less ideal than just using the most_recent_unseen_reply object we've already got?

teceler: Using the object would work for 'does this exist' I guess but doesn't work for 'are there more than one'. I'm not sure if there's a more efficient way to do that?

Marri: Well, in the case where we expect that @unseen_replies will be loaded into memory anyway, we can use .size or whatever it is that gets "number of memory objects" instead of "SQL query", and in the case where we do not expect this we could at least change the first check? Maybe?

teceler: Oh hm yeah that's valid @unseen_replies would be loaded into memory if it exists for rendering the page a little after this. Although it's going through replies/_single I think so ti might not be loaded per-say? Anyway we can fix the first check either way.

PR #1404, opened 2020-05-06
Improve backstop files

teceler:
* Lowers the threshold for reporting differences
* Updates the html test on which the image changed slightly, I suspect for external reasons


teceler:
Superseded by #1486

PR #1409, opened 2020-05-11
Split PostScraper into PostScraper and ReplyScraper

teceler:
(split from #1224)

Codeclimate has complaints because I tried to change things as minimally as possible in the move.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -369,7 +369,7 @@ Rails/NegateInclude:
 
 Rails/Output:
   Exclude:
-    - 'app/services/post_scraper.rb' # post_scraper.rb is run manually in the command line so stdout is fine
+    - 'app/services/reply_scraper.rb' # post_scraper.rb (which calls ReplyScraper) is run manually in the command line
     - 'db/seeds.rb'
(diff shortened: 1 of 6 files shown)


.rubocop.yml
```diff
@@ -200,7 +200,7 @@ Rails/ReadWriteAttribute:
 # post_scraper.rb is run manually in the command line so stdout is fine
```
Marri, line 200: also this

app/services/post_scraper.rb
```diff
@@ -1,25 +1,5 @@
 class PostScraper < Object
   SANDBOX_ID = Rails.env.production? ? 3 : 5
```
Marri, line 2: is this even still correct?

teceler: It is not!

.rubocop.yml
teceler: PostScraper is in fact the part that's run so it seemed more correct to leave that?

Marri: Valid, could be ~any more clear that post_scraper calls reply_scraper but it's probably fine as is

Marri (approved):

PR #1411, opened 2020-05-12
Make sandbox id a constant on Board

### LINKED ISSUE #1081: Update move_posts_to_sandbox to hard-code board name instead of id
https://github.com/Marri/glowfic/blob/ad00696b4d2eeb0a3e19aec9fe6adf270a45fd79/app/models/board.rb#L57-L60

teceler:
Fixes #1081


teceler:
Okay, if this counts as not hardcoding the id this should fix #1081 too

PR #1414, opened 2020-05-15
Move PostAuthor to Post::Author

Marri:
LET THE SORTING CONTINUE


script/seed_dumper.rb
```diff
@@ -16,7 +16,7 @@
   TagTag: ['created_at', 'updated_at', 'suggested'],
   PostTag: ['created_at', 'updated_at', 'suggested'],
   'Audited::Audit': [],
-  PostAuthor: ['created_at', 'updated_at'],
+  Post::Author: ['created_at', 'updated_at'],
   Message: ['created_at', 'updated_at'],
   Post::View: ['created_at', 'updated_at', 'ignored', 'notify_message', 'notify_email', 'warnings_hidden'],
```
teceler, line 21: These both need `=>`

spec/models/post_author_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.describe PostAuthor do
+RSpec.describe Post::Author do
```
teceler, line 1: According to rubocop this should be moved to spec/models/post/author_spec.rb

spec/controllers/posts_controller_spec.rb
```diff
@@ -2502,16 +2502,16 @@ def restore(reply)
       expect(flash[:success]).to eq("Post deleted.")
     end
 
-    it "deletes PostAuthors" do
+    it "deletes Post::Authors" do
```
teceler, line 2505: Not sure if this was intended or overeager find/replace but is fine either way, just stood out to me.

Marri: Both, find/replace found it but I did intentionally decide that if the message is specifying the name of the class then I would specify the _correct_ name of the class at least.

PR #1415, opened 2020-05-16
Improve query for users' boards in og data

teceler:


teceler:
Yeah, I changed it so we did, and then... missed this query in the cleanup, afaict

PR #1422, opened 2020-05-19
Don't redefine the ordered_manually scope

Marri:
Causes Orderable not to define ordered_manually if the class has already defined it, which should keep us from getting those 'scope already defined' warnings.


teceler:
Oh is that how that happens. 

This looks good?

PR #1423, opened 2020-05-19
Reducing icon#show footprint

Marri:
- We were plucking icon attributes slightly too late in galleries/_single, which meant the icons.present? line would load them into memory anyway. We have nice shiny low memory partials now and we should take advantage of this by not loading the icons into memory at all.

- We were querying Gallery up to five different times for icon#show, specifically:
  - Pluck on icon.galleries in name in og_data
  - Count on icon.galleries to determine whether to show a breadcrumb
  - [if previous check passed] Fetching icon.galleries to link to icon.galleries.first in breadcrumbs
  - icon.galleries.exist? to determine whether to render the galleries/single partial collection
  - [if previous check passed] Another icon.galleries fetch to actually render the collection, because this fetch used an order and so it would not reuse the association cache.

### DIFF
--- a/app/controllers/icons_controller.rb
+++ b/app/controllers/icons_controller.rb
@@ -55,6 +55,7 @@ class IconsController < UploadingController
       @times_used = (posts_using.count + replies_using.count)
       @posts_used = (posts_using.pluck(:id) + replies_using.select(:post_id).distinct.pluck(:post_id)).uniq.count
     end
+    @galleries = @icon.galleries.ordered_by_name
     @meta_og = og_data
   end
 
@@ -180,7 +181,7 @@ class IconsController < UploadingController
   end
 
   def og_data
-    galleries = @icon.galleries.pluck(:name)
+    galleries = @galleries.map(&:name)
     if galleries.present?
       desc = "Gallery".pluralize(galleries.count) + ": " + galleries.join(', ')
     else
--- a/app/views/icons/show.haml
+++ b/app/views/icons/show.haml
@@ -7,8 +7,8 @@
       &raquo;
       = link_to "#{@icon.user.username}'s Galleries", user_galleries_path(@icon.user)
   &raquo;
-  - if @icon.galleries.count == 1
-    = link_to @icon.galleries.first.name, gallery_path(@icon.galleries.first)
+  - if @galleries.present?
+    = link_to @galleries.first.name, url_for(@galleries.first)
   - else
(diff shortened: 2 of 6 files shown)


app/controllers/icons_controller.rb
```diff
@@ -55,6 +55,7 @@ def show
       @times_used = (posts_using.count + replies_using.count)
       @posts_used = (posts_using.pluck(:id) + replies_using.select(:post_id).distinct.pluck(:post_id)).uniq.count
     end
+    @galleries = @icon.galleries.ordered_by_name
```
teceler, line 58: Oh that's clever. Although wouldn't it use the association cache anyway?

app/views/icons/show.haml
```diff
@@ -7,8 +7,8 @@
       &raquo;
       = link_to "#{@icon.user.username}'s Galleries", user_galleries_path(@icon.user)
   &raquo;
-  - if @icon.galleries.count == 1
-    = link_to @icon.galleries.first.name, gallery_path(@icon.galleries.first)
+  - if @galleries.present?
+    = link_to @galleries.first.name, url_for(@galleries.first)
```
teceler, line 11: Won't this load-into-memory? I guess if it's using the association cache that doesn't matter but I vaguely recall a thing that queried exists if it wasn't loaded and the cache if it was.

Marri: Yes, @galleries is loaded in memory, because galleries/single still uses Gallery objects. I could try to unravel it to one query for <all icon#show> and another to get Galleries for view=galleries if we wanted though?

app/controllers/icons_controller.rb
Marri: I'm not sure which cache you're gesturing at? But @icon.galleries and @icon.galleries.ordered_by_name wouldn't use the same cache afaict.

app/views/icons/show.haml
teceler: Yeah I'm mostly eying it because I have learned to be leery of `.present?` and we're only using the first one here afaict

app/controllers/icons_controller.rb
teceler: Huh. I guess that makes sense, and so it makes sense to add this. :+1:

Marri: Yeah none of the other cases using galleries seemed to care what the order was, so no reason not to be consistent :shrug:

app/views/icons/show.haml
Marri: I mean I do literally at a later point fix a bad usage of `.present?` you are not wrong xD

PR #1429, opened 2020-05-23
Clean up various rubocop lints

teceler:
* Compresses single line if/unless statmeents
* Cleans up an unused assignment
* Adds empty lines after before hooks (even if they were immediately followed by after hooks, _rubocop_)
* Cleans up some whitespace
* Removes some leading zeros that rubocop objected to
* Changes keys.each to each_key
* Moves Post::View specs to `spec/models/post/view_spec.rb` to match their model location

### DIFF
--- a/app/controllers/characters_controller.rb
+++ b/app/controllers/characters_controller.rb
@@ -249,9 +249,7 @@ class CharactersController < ApplicationController
       success_msg = " in the specified " + 'post'.pluralize(params[:post_ids].size)
     end
 
-    if @character.aliases.exists? && params[:orig_alias] != 'all'
-      wheres[:character_alias_id] = orig_alias.try(:id)
-    end
+    wheres[:character_alias_id] = orig_alias.try(:id) if @character.aliases.exists? && params[:orig_alias] != 'all'
 
     UpdateModelJob.perform_later(Reply.to_s, wheres, updates)
     wheres[:id] = wheres.delete(:post_id) if params[:post_ids].present?
@@ -341,9 +339,7 @@ class CharactersController < ApplicationController
     @groups = user.character_groups.order('name asc') + [new_group]
     use_javascript('characters/editor')
     gon.character_id = @character.try(:id) || ''
-    if @character.present? && @character.template.nil? && @character.user == current_user
-      @character.build_template(user: user)
-    end
+    @character.build_template(user: user) if @character.present? && @character.template.nil? && @character.user == current_user
     gon.user_id = user.id
     @aliases = @character.aliases.ordered if @character
     gon.mod_editing = (user != current_user)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -33,9 +33,7 @@ module ApplicationHelper
   end
 
   def quick_switch_tag(image_url, short_text, hover_name, char_id)
-    if image_url.nil?
-      return content_tag :div, short_text, class: CHAR_ICON_FAKE, title: hover_name, data: { character_id: char_id }
-    end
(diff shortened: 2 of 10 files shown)


app/controllers/characters_controller.rb
```diff
@@ -341,9 +339,7 @@ def editor_setup
     @groups = user.character_groups.order('name asc') + [new_group]
     use_javascript('characters/editor')
     gon.character_id = @character.try(:id) || ''
-    if @character.present? && @character.template.nil? && @character.user == current_user
-      @character.build_template(user: user)
-    end
+    @character.build_template(user: user) if @character.present? && @character.template.nil? && @character.user == current_user
```
Marri, line 342: We can probably shorten this with a &. somewhere (maybe `@character&.user == current_user` assuming we have login_required?) but that can be a problem for future-us

app/helpers/application_helper.rb
```diff
@@ -33,9 +33,7 @@ def loading_tag(**args)
   end
 
   def quick_switch_tag(image_url, short_text, hover_name, char_id)
-    if image_url.nil?
-      return content_tag :div, short_text, class: CHAR_ICON_FAKE, title: hover_name, data: { character_id: char_id }
-    end
+    return content_tag :div, short_text, class: CHAR_ICON_FAKE, title: hover_name, data: { character_id: char_id } if image_url.nil?
```
Marri, line 36: Surprised this doesn't trigger a line width complaint but I guess not 🤷 my usual check is if it's readable on Github, which this isn't but honestly not a big deal.

app/models/post.rb
```diff
@@ -160,9 +160,7 @@ def recent_characters_for(user, count)
       .pluck(:character_id)
 
     # add the post's character_id to the last one if it's not over the limit
-    if character_id.present? && user_id == user.id && recent_ids.length < count && !recent_ids.include?(character_id)
-      recent_ids << character_id
-    end
+    recent_ids << character_id if character_id.present? && user_id == user.id && recent_ids.length < count && !recent_ids.include?(character_id)
```
Marri, line 163: Same boat here :shrug:

spec/support/spec_feature_helper.rb
```diff
@@ -8,9 +8,7 @@ def login(user=nil, password='known')
     fill_in "Username", with: user.username
     fill_in "Password", with: password
     click_button "Log in"
-    if page.all('.flash.error').present?
-      raise(RuntimeError, "Failed to log in as '#{user.username}':\n" + page.find('.flash.error').text)
-    end
+    raise(RuntimeError, "Failed to log in as '#{user.username}':\n" + page.find('.flash.error').text) if page.all('.flash.error').present?
```
Marri, line 11: 👒

app/controllers/characters_controller.rb
teceler: ooh does `&.` work when the statement is part of an equality check

app/helpers/application_helper.rb
teceler: It's actually like 15 characters short of the max, apparently

app/controllers/characters_controller.rb
Marri: I mean it's more that @character&.user would return nil if @character is nil, and `nil == current_user` will therefore return false since this is only called within the context of a login_required. If we expect logged out users to someday be able to edit characters that would break but doesn't seem all that likely tbh.

That said this is all rambling and if we want to just do this that's fine, "whine whine github scrollbar" is not a particularly valid complaint xD

teceler: honestly I think it's a very valid complaint, the github scrollbar is terrible, but.

PR #1430, opened 2020-05-23
Improve PostAuthor error checking

teceler:
* Moves `out_out_of_owed` and `opt_into_owed` to PostAuthor
* Checks outcome of create/update/destroy operations, and merges errors if they fail


app/concerns/owable.rb
```diff
@@ -40,14 +39,33 @@ def author_for(user)
 
     private
 
+    def change_owed
+      success = yield
+      errors.merge(errors) unless success
```
Throne3d, line 44: This looks like it'll be broken - should it be

```suggestion
      errors.merge(author.errors) unless success
```

like below? (Also can we add a quick test to see what these errors look like?)

teceler: yeees I suspect that's a failure from a point in the process where I'd moved those methods over to postauthor before deciding that was Way Out Of Scope xD. and sure I can write a test, I just have to think of an error that would trip it. how do you get invalid post authors, hmmmm.

PR #1442, opened 2020-05-27
Add slightly more block navigation for users

Marri:
Adds a link to either create or edit a block from user#show, and adds breadcrumbs to block#new and block#edit for slightly easier subsequent navigation, and supports passing a blocked_user_id to block#new.

Screenshot from 2020-05-26 23-02-14
Screenshot from 2020-05-26 22-59-13


app/views/blocks/new.haml
```diff
@@ -1,3 +1,8 @@
+- content_for :breadcrumbs do
+  = link_to 'Blocks', blocks_path
+  &raquo;
+  %b New
```
teceler, line 4: Maybe display the blockee here if they're in params?

PR #1446, opened 2020-06-10
Users cannot hide your shared posts from you by blocking you

Marri:


app/models/user.rb
```diff
@@ -145,12 +145,14 @@ def username_not_reserved
...
 
-  def blocked_or_hidden_posts(keyword, user_ids)
+  def blocked_or_hidden_posts(keyword, user_ids, block_own: true)
     Rails.cache.fetch(Block.cache_string_for(self.id, keyword), expires_in: 1.month) do
-      Post.unscoped.where(
+      post_ids = Post.unscoped.where(
         authors_locked: true,
         id: Post::Author.where(user_id: user_ids).select(:post_id),
       ).pluck(:id)
+      post_ids -= Post::Author.where(user_id: self.id).pluck(:post_id) unless block_own
```
teceler, line 154: I think I would write this as
```post_ids = Post::Author.where(user_id: user_ids).select(:post_id)
post_ids = post.ids.where.not(user_id: self.id) if keyword == 'hidden'
Post.unscoped.where(authors_locked: true, id: post_ids).pluck(:id)```

PR #1448, opened 2020-06-10
Fix post blocking caches

teceler:
* Clears caches on block create/update/destroy
* Clears caches on post create/update
* Tests this behaviour

### DIFF
--- a/app/models/block.rb
+++ b/app/models/block.rb
@@ -18,7 +18,7 @@ class Block < ApplicationRecord
   POSTS = 1
   ALL = 2
 
-  CACHE_VERSION = 2
+  CACHE_VERSION = 3
 
   def editable_by?(user)
     return false unless user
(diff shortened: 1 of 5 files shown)


app/models/block.rb
```diff
@@ -65,7 +65,6 @@ def mark_messages_read
   end
 
   def invalidate_caches
-    return unless created_at == updated_at || saved_change_to_hide_me? || saved_change_to_hide_them?
```
Marri, line 68: Porque?

teceler: Because I want it to actually trigger consistently on destroys xD
Also I think I was assuming when I wrote this that the caches won't exist yet on creates but that's not... actually... true...

Marri: But you were already doing this on creates so that's fine, I assume the problem was just that delete wasn't showing things as relevantly changed?

teceler: ...oh, right, that's what the equals was. Okay that makes more sense as to what past!me was thinking. 

Anyway I could create a method that calls this one to be called on update and check if the right things changed, but it seemed simpler to just potentially over-clear slightly.

But yes delete wasn't. Apparently.

PR #1449, opened 2020-06-11
Fix cache calls in visible_posts tests to ensure accurate results

teceler:
I think this isn't working properly because of the `allow` being on `Rails.cache` but I'm not sure I just know I was getting false all the time when I was testing blocking caches.


Throne3d:
Maybe we could use a `memory_store`, and clear it between tests, like in https://stackoverflow.com/questions/13309121/is-rails-cache-purged-between-tests?

Marri:
Work uses fakeredis iirc, we can look into that?

teceler:
Superseded by #1537

PR #1450, opened 2020-06-14
Avoid plucking single-use ids

teceler:


app/models/character.rb
```diff
@@ -61,9 +61,9 @@ def self.deletable_by?(user, user_id)
 
   def recent_posts
     return @recent unless @recent.nil?
-    reply_ids = replies.group(:post_id).pluck(:post_id)
-    post_ids = posts.select(:id).map(&:id)
-    @recent ||= Post.where(id: (post_ids + reply_ids).uniq).ordered
+    reply_ids = replies.select(:post_id).distinct
+    post_ids = posts.map(&:id)
```
Throne3d, line 65: This looks to be loading all this character's posts, with all associated model details - is that intentional?

spec/models/user_spec.rb
```diff
@@ -139,33 +139,6 @@ def generate_emailless_user
       expect(blocker).not_to have_interaction_blocked(only_posts)
       expect(blocker).to have_interaction_blocked(blockees.first)
     end
-
-    describe "content" do
```
Throne3d, line 143: What's going on with this change?

app/models/character.rb
teceler: ...huh, I assumed it wouldn't load everything until it was actually called?

spec/models/user_spec.rb
Throne3d: Oh, we don't use hidden_post_users or blocking_post_users anymore? Huh, strange.

teceler: We weren't using that code! So I took a machete to it

app/models/character.rb
Throne3d: Unfortunately `.map` is one of the Relation methods that loads everything and starts applying attribute calls to each model in the list - `.pluck` probably works better here, or `.select(:id)`, like you did elsewhere?

teceler: ...I'm confused on why that map is still there, yup. I think I might have deleted the wrong thing. Good catch, I didn't even see that xD

Throne3d:
I think you're good to merge this if you rebase it to master again! There have been more changes to the base since, so I'd recommend merging it as soon as tests pass after the rebase.

Throne3d (changes requested):
Blocked while we have database concerns – https://discordapp.com/channels/521407923629457428/521411396613701662/767456739871162378

PR #1461, opened 2020-07-12
Fix minor model callback behaviors

teceler:
* Clearing IndexPost section ids will now use UpdateModelJob
* FlatPosts will now queue to regenerate immediately when replies are destroyed
* Various callbacks which queue jobs will only occur after the model is committed to the database

### DIFF
--- a/app/models/icon.rb
+++ b/app/models/icon.rb
@@ -22,7 +22,7 @@ class Icon < ApplicationRecord
   before_validation :use_icon_host
   before_save :use_https
   before_update :delete_from_s3
-  after_destroy :clear_icon_ids, :delete_from_s3
+  after_destroy_commit :clear_icon_ids, :delete_from_s3
 
   scope :ordered, -> { order(Arel.sql('lower(keyword) asc'), created_at: :asc, id: :asc) }
 
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -10,7 +10,7 @@ class Message < ApplicationRecord
 
   before_validation :set_thread_id, :remove_deleted_recipient
   before_create :check_recipient
-  after_create :notify_recipient
+  after_create_commit :notify_recipient
 
   scope :ordered_by_id, -> { order(id: :asc) }
   scope :ordered_by_thread, -> { order(thread_id: :asc, id: :desc) }
--- a/app/models/reply.rb
+++ b/app/models/reply.rb
@@ -12,10 +12,11 @@ class Reply < ApplicationRecord
   validate :author_can_write_in_post, on: :create
   audited associated_with: :post, except: :reply_order, update_with_comment_only: false
 
-  after_create :notify_other_authors, :destroy_draft, :update_active_char, :set_last_reply, :update_post, :update_post_authors
-  after_save :update_flat_post
+  after_create :notify_other_authors, :destroy_draft, :update_active_char, :set_last_reply, :update_post
+  after_create_commit :update_post_authors # this can generate notification jobs
   after_update :update_post
   after_destroy :set_previous_reply_to_last, :remove_post_author
+  after_commit :update_flat_post
 
   attr_accessor :skip_notify, :skip_post_update, :is_import, :skip_regenerate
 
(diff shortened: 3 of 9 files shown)


app/models/reply.rb
```diff
@@ -12,10 +12,11 @@ class Reply < ApplicationRecord
   validate :author_can_write_in_post, on: :create
   audited associated_with: :post, except: :reply_order, update_with_comment_only: false
 
-  after_create :notify_other_authors, :destroy_draft, :update_active_char, :set_last_reply, :update_post, :update_post_authors
-  after_save :update_flat_post
+  after_create :notify_other_authors, :destroy_draft, :update_active_char, :set_last_reply, :update_post
+  after_create_commit :update_post_authors # this can generate notification jobs
```
Marri, line 16: Is there a way to get just the notification job into an after_commit instead?

app/models/icon.rb
```diff
@@ -22,7 +22,7 @@ class Icon < ApplicationRecord
   before_validation :use_icon_host
   before_save :use_https
   before_update :delete_from_s3
-  after_destroy :clear_icon_ids, :delete_from_s3
+  after_destroy_commit :clear_icon_ids, :delete_from_s3
```
Marri, line 25: Does this still work with the destroyed? check and the key_was? Not sure what gets cleared in an after_commit vs after_destroy.

app/models/message.rb
```diff
@@ -10,7 +10,7 @@ class Message < ApplicationRecord
 
   before_validation :set_thread_id, :remove_deleted_recipient
   before_create :check_recipient
-  after_create :notify_recipient
+  after_create_commit :notify_recipient
```
Marri, line 13: Did we figure out if `deliver` aliases to `now` or `later`?

app/models/reply.rb
teceler: ...probably, if it you split it out into it's own method? That seems like overkill to me but you _could_

app/models/icon.rb
teceler: I... kind of assumed we have tests on this that would be failing if it didn't? I think they're the same for this purpose, or at least rails guides didn't say differently.

app/models/message.rb
teceler: I don't think so, but either way we want this on an after_commit?

app/models/reply.rb
Throne3d: Exceptions raised in `commit` blocks are basically ignored – since they're after a commit, they can't influence whether the base model gets saved or not, meaning we risk an error when creating a post author here and would have an inconsistency in our PostAuthor models.

I think it'd be better to leave this in `after_create` (potentially notifying someone erroneously) rather than putting it into `after_create_commit` (potentially having PostAuthor inconsistent with the post), unless we can split out just the notification logic to the `after_commit`.

app/models/message.rb
Throne3d: It looks like `deliver` comes from `resque_mailer`: https://github.com/zapnap/resque_mailer/blob/35c2a7c7ff9038b9bff95c50190d162f95afcd5f/lib/resque_mailer.rb#L137

It defaults to a behavior like `deliver_later` if it can do that through resque, or else falls back to synchronous (like `deliver_now`), but really isn't either of them! Maybe we should investigate using the ActionMailer native jobs?

app/models/icon.rb
Throne3d: We have a job checking that it correctly queues `DeleteIconFromS3Job` on destroy: https://github.com/Marri/glowfic/blob/a8c6f790cb89ddea443c995cba8d562deb99a652/spec/models/icon_spec.rb#L102

Should be good to go!

app/models/icon.rb
```diff
@@ -22,7 +22,7 @@ class Icon < ApplicationRecord
   before_validation :use_icon_host
   before_save :use_https
   before_update :delete_from_s3
```
Throne3d, line 24: Maybe this should actually be an `after_update_commit`? 🤔 

Could be a future PR though, imo

Throne3d (changes requested):
The comment about PostAuthor consistency should be addressed before merging imo, but beyond that looks good!

PR #1462, opened 2020-08-08
Fix delete_history

teceler:


app/controllers/posts_controller.rb
```diff
@@ -162,14 +162,18 @@ def delete_history
...
-    if @audits.present?
-      @audit = @audits.first
+    if @deleted_audits.present?
+      @audit = @deleted_audits.first
       @deleted = Reply.new(@audit.audited_changes)
       @preceding = @post.replies.where('id < ?', @audit.auditable_id).order(id: :desc).limit(2).reverse
       @preceding = [@post] unless @preceding.present?
       @following = @post.replies.where('id > ?', @audit.auditable_id).order(id: :asc).limit(2)
+      @audits = @post.associated_audits.where(auditable_id: @following.map(&:id))
+      @audits = @audits.or(@post.associated_audits.where(auditable_id: @preceding.map(&:id))) unless @preceding.first.is_a?(Post)
```
Marri, line 174: Can we get an explain on doing this with an or instead of just calculating the auditable_id first and then passing it to a single queryset? Otherwise this looks fine.

teceler: good... question? the preceding code was slightly terrifying, I guess? xD that should work fine

Marri: Better belated plan: Let's just set this to a `{}` and accept that the deleted replies page will not show history links? They seem unnecessary honestly?

app/controllers/posts_controller.rb
```diff
@@ -162,14 +162,18 @@ def delete_history
...
     audit_ids = audit_ids.group(:auditable_id).pluck(Arel.sql('MAX(audits.id)')) # only most recent per reply
-    @audits = Audited::Audit.where(id: audit_ids).paginate(per_page: 1, page: page)
+    @deleted_audits = Audited::Audit.where(id: audit_ids).paginate(per_page: 1, page: page)
 
-    if @audits.present?
-      @audit = @audits.first
+    if @deleted_audits.present?
+      @audit = @deleted_audits.first
       @deleted = Reply.new(@audit.audited_changes)
       @preceding = @post.replies.where('id < ?', @audit.auditable_id).order(id: :desc).limit(2).reverse
```
Marri, line 170: nts this is super not reply_order compliant

PR #1463, opened 2020-08-09
Use a user context for UpdateModelJob audits

teceler:


app/jobs/update_model_job.rb
```diff
@@ -6,9 +6,25 @@
 class UpdateModelJob < ApplicationJob
   queue_as :low
 
-  def perform(klass, where_vals, new_attrs)
+  def perform(klass, where_vals, new_attrs, user_id=0)
+    if user_id == 0
+      UpdateModelJob.notify_exception(ArgumentError, klass, where_vals, new_attrs)
+    elsif !User.exists?(user_id)
```
Throne3d, line 12: Stylistic, but I'd recommend:

```rb
    elsif (user = User.find_by(id: user_id)).nil?
```

And then I'd use this `user` later in:

```rb
      Audited.audit_class.as_user(user) do
```

That way we avoid two different calls to find the user (I haven't used `User.exists?` before like this)

app/models/application_record.rb
```diff
@@ -1,3 +1,10 @@
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
+
+  def audited_user_id
+    user = ::Audited.store[:current_user].call if ::Audited.store[:current_user].respond_to?(:call)
```
Throne3d, line 5: Looks like [audited fetches the audit user like so](https://github.com/collectiveidea/audited/blob/4ad3a9c2a7dfacedc52f9f0c2a47f7b57a73e26e/lib/audited/audit.rb#L179):

```rb
    def set_audit_user
      self.user ||= ::Audited.store[:audited_user] # from .as_user
      self.user ||= ::Audited.store[:current_user].try!(:call) # from Sweeper
      nil # prevent stopping callback chains
    end
```

It probably doesn't matter much, but how come we're inverting the order (and not just copying the code directly)?

Issue #1434, opened 2020-05-25
Stats page doesn't have og_data


teceler:
I just linked to the stats page of a post and it didn't produce og data despite being public. This seems suboptimal and would be nice to fix.

l1n:
This has been fixed in #1467, ok to close?

Marri:
Yep thanks for pointing it out, closing. I think it'll do it automatically if your PR description is "Fixes" instead of "for" or something like that? I'd have to double check the exact Github magic.

PR #1467, opened 2020-08-15
Add og:meta tags to stats page

l1n:
For #1434


app/controllers/posts_controller.rb
```diff
@@ -174,6 +174,21 @@ def delete_history
...
+    post_description = generate_short(@post.description)
+    post_description += ' ('
+    post_description += helpers.author_links(@post, linked: false)
+    post_description += ')'
+    post_description.strip!
+
+    @meta_og = {
+      title: @post.subject + ' · ' + post_location,
+      description: post_description,
+      url: @meta_canonical
```
Throne3d, line 190: I think this variable isn't set to anything in posts#stats, is it?

Throne3d (commented):
I was looking at the code and wondering if we could join it with the `og_data_for_post` function in `writable_controller`, but I think it's not worth the effort (too many differences that are important). Looks good to me, but marking comment because I'm unsure of the url field (other comment)

app/controllers/posts_controller.rb
teceler: idk, how does that get set?

Throne3d: I could only see it declared in `def show_post(cur_page=nil)` on writable_controller – which I think we don't call for the stats page

l1n: Based on the test at `spec/controllers/post_controller_spec.rb`, line 844 it should be set.

Throne3d: That line's not showing anything relevant for me: https://github.com/Marri/glowfic/blob/aa2ceed9e72deca3788f0b6e892a1b9dd7d40de5/spec/controllers/posts_controller_spec.rb#L844 (maybe it was different at a previous commit?)

l1n: Whoops, misread it as the :show test - updating with a test and setting the url.

spec/controllers/posts_controller_spec.rb
```diff
@@ -1433,6 +1433,18 @@ def restore(reply)
...
     end
 
+    it "calculates OpenGraph meta" do
+      user = create(:user, username: 'example user')
+      board = create(:board, name: 'board')
+      post = create(:post, subject: 'title', user: user, board: board)
+      get :stats, params: { id: post.id }
+
+      meta_og = assigns(:meta_og)
+      expect(meta_og[:url]).to eq(post_url(post))
```
Throne3d, line 1443:
```suggestion
      expect(meta_og[:url]).to eq(stats_post_url(post))
```

Throne3d (commented):
Looks good after this final change, though!

Throne3d (approved):

Throne3d:
I'm not 100% sure this works right with fresh_when (does it need to be called first? last?), but I'll see if I can do some local testing on it quickly before approving!

app/controllers/posts_controller.rb
```diff
@@ -176,6 +176,22 @@ def delete_history
...
+    post_description = generate_short(@post.description)
+    post_description += ' ('
+    post_description += helpers.author_links(@post, linked: false)
+    post_description += ')'
+    post_description.strip!
+
+    @meta_og = {
+      title: @post.subject + ' · ' + post_location,
+      description: post_description,
+      url: stats_post_url(@post)
```
Throne3d, line 192:
```suggestion
      url: stats_post_url(@post),
```

Throne3d:
The fresh_when thing seems to work both before and after, which I guess makes sense anyway since meta_og is only used in page render afterwards.

Seems fine!

Issue #918, opened 2018-11-06
Word counts confused by HTML tags


Marri:
Something like 
```
<a href="link">Hello</a>
```
renders as word count of 2 because of spaces.

Throne3d:
Solved in #1468

PR #1468, opened 2020-08-15
Strip HTML tags when doing word counts

l1n:
per #918

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -216,7 +216,7 @@ class Post < ApplicationRecord
   def total_word_count
     return word_count unless replies.exists?
     contents = replies.pluck(:content)
-    contents[0] = contents[0].split.size
+    contents[0] = contents[0].gsub(/<[^>]*>/ui,'').split.size
     word_count + contents.inject{|r, e| r + e.split.size}.to_i
   end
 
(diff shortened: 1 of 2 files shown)


Throne3d (approved):
Looks good to me! @Marri this is more or less what you were thinking, right?

app/models/post.rb
```diff
@@ -226,7 +227,8 @@ def word_count_for(user)
     return sum unless replies.where(user_id: user.id).exists?
 
     contents = replies.where(user_id: user.id).pluck(:content)
-    contents[0] = contents[0].split.size
+    full_sanitizer = Rails::Html::FullSanitizer.new
+    contents[0] = full_sanitizer.sanitize(contents[0]).split.size
     sum + contents.inject{|r, e| r + e.split.size}.to_i
```
Throne3d, line 232:
```suggestion
    sum + contents.inject{|r, e| r + full_sanitizer.sanitize(e).split.size}.to_i
```

Throne3d (changes requested):
Would you be able to add tests for this? I think I reviewed it without properly reading things and I just noticed that we're probably only sanitizing the top post

l1n:
> Would you be able to add tests for this? I think I reviewed it without properly reading things and I just noticed that we're probably only sanitizing the top post

Makes sense, added a test. It only looked like only the top post was being sanitized - there was an odd special case in the inject, but I've normalized it by passing 0 as the first r. @Throne3d

Throne3d (approved):
Looks good to me!

Throne3d (approved):

Throne3d:
Thanks!

PR #1471, opened 2020-08-21
Site news permalinks

l1n:


Marri (approved):
Moved the permalink higher in the link order and fixed a copy/paste artifact from Delete to Permalink but now it is :turtle:

PR #1474, opened 2020-09-03
Reduce tagged_ats of non-complete posts to the updated_at of their last reply

teceler:
(to prevent unread problems)


Throne3d (approved):

Throne3d:
Is this still an issue? Wondering if we should merge this migration or do something in the console to fix things

PR #1485, opened 2020-09-28
Hotfix total_entries bugs

teceler:


Throne3d (approved):
The new test fails on master (gives 3 rather than 2) and succeeds here (when tested locally). The change  logically makes sense to me. Seems good!

Throne3d:
Updated test gives a failure of 4 on master, but succeeds on the branch! Now tests JOIN and GROUP BY, which was the main issue with the reports page (previously only tested visible_to by accident).

PR #1491, opened 2020-10-27
Don't use default postgres port

echurilov:
I ran into an issue where the postgres in the docker container was conflicting with my local postgres. Updating the glowfic postgres port to something other than the default fixed it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,7 +47,7 @@ services:
     volumes:
       - postgres-data:/var/lib/postgresql/data
     ports:
-      - 5432:5432
+      - 5555:5555
 
 volumes:
   bundler-volume:


docker-compose.yml
```diff
@@ -47,7 +47,7 @@ services:
     volumes:
       - postgres-data:/var/lib/postgresql/data
     ports:
-      - 5432:5432
+      - 5555:5555
```
Throne3d, line 50: I think this'll need to be `5555:5432`, so it still references the right port inside the container, and we may also need to update the `GLOWFIC_DATABASE_HOST` env above to include the correct port.

However, if we're only reaching the port from within Docker, it might be worthwhile using [`expose`](https://docs.docker.com/compose/compose-file/#expose) instead?

l1n: This could also be resolved by letting that mapping be arbitrary on the host and updating `bin-psql` to reflect that port

PR #1493, opened 2020-11-01
Use Newest First as the default sort order for reply search links

Marri:
Searching replies without specifying text means we should specify a sort (here newest first) since 'best match' relies on pgsearch data that won't exist. Currently only fixes various relevant links to reply sort, but possibly we should make the search page itself smarter at some point.


teceler (commented):
It might be better to make the search page smarter, long-term, but this looks fine overall

PR #1501, opened 2021-03-07
Set ETags and Last-Modified

l1n:
Only for `/posts/:id/stats` and `/posts/owed` for now.


app/controllers/posts_controller.rb
```diff
@@ -37,6 +37,7 @@ def owed
     end
 
     @posts = posts_from_relation(@posts.ordered)
+    fresh_when(etag: @posts, last_modified: @posts.max_by(&:tagged_at)&.tagged_at, public: false)
```
Throne3d, line 40: I think this `last_modified` doesn't work in all cases – if you tag the post at the top of your list (or it disappears because it went hiatused or so on), our last_modified value may decrease, and:

> `:last_modified` Sets a “weak” last-update validator on the response. Subsequent requests that set If-Modified-Since may return a 304 Not Modified response if last_modified <= If-Modified-Since.

so the browser may receive a 304 and end up returning a stale page (with a "future" modified date, based on the more recent tag that disappeared from the list) to the user.

I think we could skip the `last_modified` param here.

Looking into ETags a bit more, I think this is a good implementation! When we get new replies on a post, we update both `tagged_at` and `updated_at`, and it seems like that's based on `updated_at` and/or `cache_key_with_version`.

app/controllers/posts_controller.rb
```diff
@@ -175,6 +176,7 @@ def delete_history
   end
 
   def stats
+    fresh_when(etag: @post, last_modified: @post.tagged_at, public: false)
```
Throne3d, line 179: We should use just `@post.updated_at` here, I think – the main post could be updated, changing its word count, without updating `tagged_at`, but changing the contents of this page. The `object:` param should handle both etag and last_modified correctly:

```suggestion
    fresh_when(object: @post, public: false)
```

PR #1502, opened 2021-03-08
Disable CSRF protection on API

l1n:


Throne3d (approved):
This means we won't be able to access the user's session if the CSRF token isn't passed, which seems good for API protection! The user can still send `curl` requests to the API with a token and use that for authentication.

Throne3d:
(Hit into this myself while trying to test out an endpoint with `curl`)

Throne3d:
See, for example:

* https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/request_forgery_protection.rb#L15
* https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/controllers/graphql_controller.rb#L30
* https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/controllers/sessions_controller.rb#L52

for other examples of the difference between UI and API auth

PR #1503, opened 2021-03-13
Reply indicator

### LINKED ISSUE #1154: Show indicator of replies owed
Requested by Sonata

l1n:
Closes #1154. This will have performance impact, probably better to merge after postgres is moved. To mitigate somewhat, the indicator is defaulted to off.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -122,4 +122,21 @@ module ApplicationHelper
     return '(deleted user)'.html_safe if deleted
     link_to username, user_path(user_id)
   end
+
+  # Count of posts the user owes a reply to. Mirrors the logic in
+  # PostsController#owed but returns just the count; used to render the
+  # badge next to the "Replies Owed" nav link when the user has opted
+  # into the indicator.
+  def replies_owed_count(user)
+    return 0 unless user
+    ids = Post::Author.where(user_id: user.id, can_owe: true).group(:post_id).pluck(:post_id)
+    drafts = ReplyDraft.where(post_id: ids, user: user).pluck(:post_id)
+    solo = Post::Author.where(post_id: ids).group(:post_id).having('count(post_id) < 2').pluck(:post_id)
+    posts = Post.where(id: ids)
+      .where.not(last_user: user)
+      .or(Post.where(id: (drafts + solo).uniq))
+      .where.not(status: [:complete, :abandoned])
+    posts = posts.where.not(status: :hiatus).where('tagged_at > ?', 1.month.ago) if user.hide_hiatused_tags_owed?
+    posts.count
+  end
 end
--- a/app/views/layouts/application.haml
+++ b/app/views/layouts/application.haml
@@ -123,7 +123,12 @@
             - if logged_in?
               %li= link_to 'Unread', unread_posts_path
               - unless current_user.read_only?
-                %li= link_to 'Replies Owed', owed_posts_path
+                %li
+                  = link_to 'Replies Owed', owed_posts_path
+                  - if current_user.replies_owed_indicator
+                    - owed_count = replies_owed_count(current_user)
+                    - if owed_count > 0
+                      %span.badge.badge-primary= owed_count
             %li= link_to 'Tags', tags_path
             - day = DailyReport.unread_date_for(current_user)
             - daybadge = DailyReport.badge_for(day)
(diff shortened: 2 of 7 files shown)


app/helpers/application_helper.rb
```diff
@@ -252,4 +252,20 @@ def allowed_boards(obj, user)
...
+    @posts = @posts.where.not(last_user: current_user).or(@posts.where(id: (drafts + solo).uniq))
+    @posts = @posts.where.not(status: [:complete, :abandoned])
+    hiatused = @posts.hiatus.or(@posts.where('tagged_at < ?', 1.month.ago))
+
+    if current_user.hide_hiatused_tags_owed?
+      @posts = @posts.where.not(status: :hiatus).where('tagged_at > ?', 1.month.ago)
+      @hiatused_exist = true if hiatused.count > 0
+    end
+
+    return @posts.count
```
Marri, line 269: None of these should be instance variables; among other reasons, we have a number of controllers which set a different @posts variable and they'll conflict. You're not using them anywhere else, so they can safely drop the @.

app/views/layouts/application.haml
```diff
@@ -111,6 +111,10 @@
             - if logged_in?
               %li= link_to 'Unread', unread_posts_path
               %li= link_to 'Replies Owed', owed_posts_path
+              - if current_user.replies_owed_indicator
+                - $owed = get_owed
```
Marri, line 115: Don't need the $ here; that makes it a global, which we don't especially need it to be doing.

app/helpers/application_helper.rb
```diff
@@ -252,4 +252,20 @@ def allowed_boards(obj, user)
     authored_ids = BoardAuthor.where(user: user).select(:board_id)
     Board.where(id: obj.board_id).or(Board.where(authors_locked: false)).or(Board.where(id: authored_ids)).ordered
   end
+  def get_owed
+    ids = Post::Author.where(user_id: current_user.id, can_owe: true).group(:post_id).pluck(:post_id)
+    @posts = Post.where(id: ids)
+    drafts = ReplyDraft.where(post_id: @posts.select(:id)).where(user: current_user).pluck(:post_id)
```
Marri, line 258: You should be able to use `ids` here instead of `@posts.select(:id)` to make it more efficient

app/helpers/application_helper.rb
```diff
@@ -252,4 +252,20 @@ def allowed_boards(obj, user)
...
+    solo = Post::Author.where(post_id: ids).group(:post_id).having('count(post_id) < 2').pluck(:post_id)
+    @posts = @posts.where.not(last_user: current_user).or(@posts.where(id: (drafts + solo).uniq))
+    @posts = @posts.where.not(status: [:complete, :abandoned])
+    hiatused = @posts.hiatus.or(@posts.where('tagged_at < ?', 1.month.ago))
```
Marri, line 262: You don't need this (I assume it's just a copy/paste artifact)

app/helpers/application_helper.rb
```diff
@@ -252,4 +252,20 @@ def allowed_boards(obj, user)
...
+
+    if current_user.hide_hiatused_tags_owed?
+      @posts = @posts.where.not(status: :hiatus).where('tagged_at > ?', 1.month.ago)
+      @hiatused_exist = true if hiatused.count > 0
```
Marri, line 266: Ditto copy/paste artifact (the rest of the if block is fine)

app/views/layouts/application.haml
```diff
@@ -111,6 +111,10 @@
...
+                - if $owed
```
Marri, line 116: You should probably check > 0 here; despite also being C-based, Ruby (unlike Python) generally treats 0 as truthy.

app/views/users/edit.haml
```diff
@@ -80,6 +80,9 @@
         %td{class: cycle('even', 'odd')}
           = f.check_box :hide_hiatused_tags_owed, class: 'width-15 vmid'
           = f.label :hide_hiatused_tags_owed, "Hide hiatused threads"
+          %br
+          = f.check_box :replies_owed_indicator, class: 'width-15 vmid'
+          = f.label :replies_owed_indicator, "Show replies owed count"
```
Marri, line 85: Might want an extra clarifying "in top menu bar" or something here

Marri (changes requested):
Suggested various tweaks mostly to fix copy paste artifacts; also as a general comment needs tests

PR #1504, opened 2021-03-14
OAuth Provider

### LINKED ISSUE #1049: OAuth support for integrations
Allow applications to authenticate as a user with limited scope and without direct password access using an OAuth2 flow. This also requires a UI for managing API access granted to revoke tokens.
 reported on behalf of Vona ▣ via [Fletcher](fletcher.fun) on Discord (glowfic:constellation-dev)

l1n:
## Summary

Implements OAuth 2.0 provider functionality for the Glowfic application (issue #1049).

### Changes

**New OAuth Controllers:**
- `OauthController` — Token endpoint (client_credentials, password, authorization_code grants), authorize (GET/POST), revoke, test_request
- `OauthClientsController` — CRUD for OAuth client applications
- `ProviderAuthorizer` — Authorization logic for code/token response types

**New OAuth Models:**
- `ClientApplication` — OAuth client registration (key/secret generation via SecureRandom)
- `OauthToken` — Base token model with key generation, invalidation
- `AccessToken` — Authorized access tokens (inherits OauthToken)
- `Oauth2Token` — OAuth 2.0 specific tokens
- `Oauth2Verifier` — Authorization code verifier
- `OauthNonce` — Nonce tracking for replay prevention

**Other Changes:**
- OAuth routes (token, authorize, revoke, test_request, client CRUD)
- OAuth views (authorize pages, client management)
- Migration for OAuth tables (client_applications, oauth_tokens, oauth_nonces)
- API controller: OAuth token authentication support alongside JWT
- User model: `has_many :client_applications` and `:tokens` associations
- Removed `oauth-plugin` gem dependency, replaced with `SecureRandom.hex`
- Fixed `allow_other_host: true` for OAuth redirect (Rails 7+ requirement)

**Specs:**
- Full controller specs for OauthController, OauthClientsController, ProviderAuthorizer
- Model specs for all 6 OAuth models
- OAuth spec helper for test setup

## Test plan
- [x] All CI checks pass (rspec, rubocop, haml-lint, brakeman, bundler-audit, eslint, CodeQL)


app/controllers/api/api_controller.rb
```diff
@@ -15,6 +16,12 @@ class Api::ApiController < ActionController::Base
 
   protected
 
+  def oauth_or_jwt
+    @current_user = @current_user || current_token&.user
+  end
+  def current_user=(user)
```
Throne3d, line 22: (This is strange – `x=` methods should be setters, not equality checkers. Why not just access `current_user` directly where you need to check it?)

app/controllers/oauth_clients_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    if @client_application.save
+      flash[:notice] = "Registered the information successfully"
+      redirect_to :action => "show", :id => @client_application.id
+    else
+      render :action => "new"
+    end
+  end
+
+  def user_params
+      params.require(:client_application).permit(:name, :callback_url, :support_url, :url)
```
Throne3d, line 25: The style's a bit strange on a lot of this code – e.g. here, where the indent doesn't match the rest of the file (4 spaces in from the method, but the method is 2 spaces in from the root), or like where we're using arrow syntax for hashes (`:action => "show"`) and strings for action values – basically all our other controllers use e.g. `render :new` syntax, which is implicitly the action, and hashes are done as e.g. `id: @client_application.id`.

spec/models/oauth_token_spec.rb
```diff
@@ -0,0 +1,311 @@
...
+        end
+      end
+
+      it "should return 1.0 style to_query" do
+        expect(@token.to_query).to eq "oauth_token=#{@token.token}&oauth_token_secret=#{@token.secret}&oauth_callback_confirmed=true"
+      end
+    end
+  end
+
+  if defined? OAUTH_10_SUPPORT && OAUTH_10_SUPPORT
```
Throne3d, line 237: Do we need OAuth 1 support? I don't expect so, and that might be a big simplification to this PR (I see a lot of chunks that are OAuth 1 dedicated)

spec/models/client_application_spec.rb
```diff
@@ -0,0 +1,31 @@
+require File.dirname(__FILE__) + '/../spec_helper'
+RSpec.describe ClientApplication do
+  fixtures  :client_applications, :oauth_tokens
+  before(:each) do
+    @user = User.find_by_id(1) || create(:user)
```
Throne3d, line 5: Mentioned via PM too: it's a bit weird that this code is trying to use User ID 1 by default and falling back to creating a user – you mentioned some of these tests being pulled from the upstream library, which might be why?

spec/helpers/oauth_controller_spec_helper.rb
```diff
@@ -0,0 +1,67 @@
...
+    
+    @access_token=AccessToken.create :user => @user, :client_application => @client_application
+  end
+  
+  def setup_oauth_for_user
+    login_as_application_owner
+    setup_oauth
+    @tokens=[@request_token]
+    @user.reload
+    allow(@user).to receive(:client_applications).and_return(@client_applications)
```
Throne3d, line 34: (Mentioned by PM too) this stubbing is weird and probably the cause of the errors in RSpec – I'm betting that the model the application loads is a new User model that conceptually represents the same ID, so the stub won't go through. We should save these applications in the database where possible – you mentioned some errors with that, and I think it'd be good to get those straightened out

app/models/request_token.rb
```diff
@@ -1,18 +1,17 @@
...
     self.authorized_at = Time.zone.now
-    self.verifier = OAuth::Helper.generate_key(20)[0, 20] unless oauth10?
+    self.verifier = OAuth::Helper.generate_key(20)[0, 20]
     self.save!
   end
 
   def exchange!
     return false unless authorized?
-    return false unless oauth10? || verifier==provided_oauth_verifier
+    return verifier==provided_oauth_verifier
```
Throne3d, line 14:
```
return false unless verifier==provided_oauth_verifier
```

otherwise this returns unconditionally and doesn't execute the transaction below

app/models/oauth_nonce.rb
```diff
@@ -0,0 +1,14 @@
+# Simple store of nonces. The OAuth Spec requires that any given pair of nonce and timestamps are unique.
+# Thus you can use the same nonce with a different timestamp and viceversa.
+class OauthNonce < ApplicationRecord
+  validates :nonce, presence: true
+  validates :timestamp, presence: true
+  validates :nonce, :unique: true, :scope => :timestamp
```
Throne3d, line 6:
```suggestion
  validates :nonce, unique: true, scope: :timestamp
```

compwron (approved):
why do I have this button?

app/controllers/oauth_controller.rb
```diff
@@ -0,0 +1,100 @@
+class OauthController < ApplicationController
+  before_action :login_required, :only => [:authorize, :revoke]
+  oauthenticate :only => [:test_request]
+  oauthenticate :strategies => :token, :interactive => false, :only => [:invalidate]
+  skip_before_action :verify_authenticity_token, :only=>[:invalidate, :test_request, :token]
```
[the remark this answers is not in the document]
l1n: I, did that on purpose it's an API

app/controllers/api/api_controller.rb
```diff
@@ -16,6 +17,10 @@ class Api::ApiController < ActionController::Base
 
   protected
 
+  def oauth_or_jwt
+    @current_user ||= oauth_token_user
```
Marri: This all continues to belong in the Authentication::Api helper, and probably we don't want separate methods, we want current user to ||= oauth_token_user || jwt_user

app/controllers/oauth_clients_controller.rb
```diff
@@ -0,0 +1,57 @@
+class OauthClientsController < ApplicationController
+  before_action :login_required
+  before_action :get_client_application, only: [:show, :edit, :update, :destroy]
+
+  def index
+    @client_applications = @current_user.client_applications
```
Marri: You shouldn't need to use @current_user here, current_user should work fine

app/controllers/oauth_clients_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    @tokens = @current_user.tokens.where('oauth_tokens.invalidated_at is null and oauth_tokens.authorized_at is not null')
```
Marri: optional but we could write a scope for this maybe?

app/controllers/oauth_clients_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    @client_application = @current_user.client_applications.build(user_params)
+    if @client_application.save
+      flash[:notice] = "Registered the information successfully"
+      redirect_to action: :show, id: @client_application.id
+    else
+      render :new
+    end
+  end
+
+  def user_params
```
Marri: This can be private

app/controllers/oauth_clients_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def new
+    @client_application = ClientApplication.new
+  end
+
+  def create
+    @client_application = @current_user.client_applications.build(user_params)
+    if @client_application.save
+      flash[:notice] = "Registered the information successfully"
+      redirect_to action: :show, id: @client_application.id
```
Marri: more stylistically written as
`redirect_to oauth_clients_path(@client_application)`

app/controllers/oauth_clients_controller.rb
```diff
@@ -0,0 +1,57 @@
+class OauthClientsController < ApplicationController
+  before_action :login_required
+  before_action :get_client_application, only: [:show, :edit, :update, :destroy]
```
Marri: not functionality relevant but we usually use find_ instead of get_

app/controllers/oauth_clients_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    if @client_application.update(user_params)
+      flash[:notice] = "Updated the client information successfully"
+      redirect_to action: :show, id: @client_application.id
+    else
+      render :edit
+    end
+  end
+
+  def destroy
+    @client_application.destroy!
```
Marri, line 42: Do you want to catch / display errors here if something goes wrong?

app/controllers/oauth_clients_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+    redirect_to action: :index
+  end
+
+  private
+
+  def get_client_application
+    @client_application = @current_user.client_applications.find_by_id(params[:id])
+    return if @client_application
+    flash.now[:error] = "Wrong application id"
+    raise ActiveRecord::RecordNotFound
```
Marri: If you raise here it won't actually display helpful information like the error message you just specified, it'll just throw an error page. If you want to render the error on another page, maybe pick another location - `oauth_clients_path` to show the index page is probably a good default - and redirect there instead of raising. CharactersController#find_model has an example you can follow, sans logged out handling. If a before_action redirects or renders it'll automatically stop executing subsequent methods so you don't need to worry about breaking the chain.

Marri (commented):
Not done yet but the smol is coughing, leaving what comments I have for now and will return later.

app/controllers/oauth_clients_controller.rb
l1n:
```
  rescue StandardError
    flash[:error] = "Failed to destroy the client application"
    redirect_to oauth_clients_path```

File: app/controllers/oauth_clients_controller.rb (hunk 11 above)
Comment (l1n) [reply app/controllers/oauth_clients_controller.rb]: `scope :authorized, -> { where.not(authorized_at: nil).where(invalidated_at: nil) }`

File: app/controllers/oauth_clients_controller.rb [hunk 17]
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def new
+    @client_application = ClientApplication.new
+  end
+
+  def create
+    @client_application = current_user.client_applications.build(user_params)
+    if @client_application.save
+      flash[:notice] = "Registered the information successfully"
```
Marri: I don't think we have a style defined for flash[:notice], only success or error. I'd either switch to :success or define your own style for this (recommend the first, since we have a lot of layouts involved in new styles)

app/controllers/oauth_clients_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+      redirect_to oauth_client_path(@client_application)
+    else
+      render :new
```
Marri, line 21: May want to flash[:error] the error here as well

app/controllers/oauth_clients_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def show
+  end
+
+  def edit
+  end
+
+  def update
+    if @client_application.update(user_params)
+      flash[:notice] = "Updated the client information successfully"
```
Marri: I will not repeat my previous comment every flash[:notice] I see, but note that there's a bunch

app/controllers/oauth_controller.rb
```diff
@@ -0,0 +1,110 @@
+class OauthController < ApplicationController
+  before_action :login_required, only: [:authorize, :revoke]
+  before_action :authenticate_token, only: [:test_request, :invalidate]
+  protect_from_forgery with: :exception, unless: -> { api_request? }
```
Marri, line 4: I thought this was on by default, what's the goal here? Also do we hit any of these endpoints as an API request?

config/application.rb
```diff
@@ -11,7 +11,7 @@ module Glowfic
   ALLOWED_TAGS = %w(b i u sub sup del ins hr p br div span pre code h1 h2 h3 h4 h5 h6 ul ol li dl dt dd a img blockquote q table tbody td th thead tr
                     strike s strong em big small font cite abbr var samp kbd mark ruby rp rt bdo wbr details summary)
   ALLOWED_ATTRIBUTES = {
-    :all         => %w(xml:lang class style title lang dir),
+    all: %w(xml:lang class style title lang dir),
```
Marri: Let's not, I like the whitespace being aligned, and it's not a functional difference.

app/models/oauth_token.rb
```diff
@@ -0,0 +1,29 @@
+class OauthToken < ApplicationRecord
+  belongs_to :client_application
+  belongs_to :user
+  validates :token, uniqueness: true
+  validates :token, presence: true
+  before_validation :generate_keys, on: :create
+
+  scope :authorized, -> { where.not(authorized_at: nil).where(invalidated_at: nil) }
+  attr_accessor :expires_at
```
Marri: Confused by this, if it's a database column it should already be available

app/controllers/oauth_clients_controller.rb
```diff
@@ -0,0 +1,59 @@
+class OauthClientsController < ApplicationController
```
Marri, line 1: Any reason we're not versioning the API for these like the rest of the API?

app/controllers/oauth_clients_controller.rb
l1n: Good catch, will switch all flash[:notice] to flash[:success].

app/controllers/oauth_clients_controller.rb
l1n: Will do — I'll add something like flash[:error] = @client_application.errors.full_messages.to_sentence before the render.

app/controllers/oauth_clients_controller.rb
l1n: Ack, will do a pass to change them all.

app/controllers/oauth_controller.rb
l1n: The `token` endpoint receives POSTs from external OAuth clients exchanging credentials for tokens, so it needs CSRF exempted. `protect_from_forgery` is on by default from ApplicationController, but the `unless: -> { api_request? }` clause is what exempts those external requests. Previously this was `skip_before_action :verify_authenticity_token` which the CodeQL bot kept flagging — this is the same thing just expressed differently. I can add a comment to clarify the intent.

config/application.rb
l1n: Will revert.

app/models/oauth_token.rb
l1n: You're right, it's a column in the migration. The attr_accessor was left over from the upstream library code and shadows the DB attribute — will remove.

app/controllers/oauth_clients_controller.rb
l1n: OauthClientsController is a user-facing web UI (flash messages, redirects, HTML views) for managing registered apps, not a programmatic API — so I didn't put it under api/v1. The OAuth protocol endpoints (token, authorize) follow the OAuth 2.0 spec URLs which clients expect at fixed paths. Happy to move it if you think it should live elsewhere though.

Marri: No, you're right, I misread it as being under controllers/api but that was just the api controller changes.

PR #1508, opened 2021-04-24
Disable audit display on replies#search

teceler:
Until we have audit count caching, at least

(This also disables display of updated_at on reply#search)


Throne3d (approved):
Looks good to me!

PR #1517, opened 2021-05-22
Don't display facecast field on character#show unless the character has one

teceler:


Throne3d:
I'm not really seeing why this change?

Throne3d (approved):
Seems okay per discussion in #devs!

PR #1522, opened 2021-05-24
Add Github Actions workflow

Throne3d:
Hopefully this will run successfully first try! Here goes...


.github/workflows/rails.yml
```diff
@@ -0,0 +1,56 @@
...
+# - https://docs.github.com/en/actions/guides/building-and-testing-ruby
+# - https://github.com/niclin/rails-github-actions-example/blob/master/.github/workflows/rails.yml
+
+name: Rails
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
teceler, line 11:
```suggestion
    branches: [ master ]
  pull_request:
    branches: [ master ]
```
(possibly we should change our main branch name but that's out of scope)

Throne3d: \o/

docker-compose.yml
```diff
@@ -34,7 +34,7 @@ services:
       RESQUE_TERM_TIMEOUT: 7
       QUEUES: mailer,notifier,high,*
   redis:
-    image: redis:4.0-alpine
+    image: redis:6.2-alpine
```
Throne3d, line 37: (Heroku has us on 6.2 already)

.github/workflows/rspec.yml
```diff
@@ -0,0 +1,56 @@
...
+      GLOWFIC_DATABASE_USER: postgres
+      GLOWFIC_DATABASE_PASS: postgres
+      GLOWFIC_DATABASE_PORT: 5432
+      CC_TEST_REPORTER_ID: 1e0c6dba9930e839038860b6d73301226c821937f57ed35d06fc0e4b7bddf5f6
+      RAILS_ENV: test
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Chrome
+        run: sudo apt-get update && sudo apt-get install -y chromium-browser
```
teceler, line 45: Did you find a lighter-weight way to do this?

.github/workflows/rspec.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.6.8
+          bundler-cache: true  # runs bundle install (with version based on Gemfile.lock) and automatically caches result
+      - name: Load database schema
+        run: bundle exec rails db:schema:load --trace
```
teceler, line 52: why trace?

.github/workflows/rspec.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Run rspec tests and report coverage
+        uses: paambaati/codeclimate-action@v3.0.0
+        with:
+          coverageCommand: bundle exec rspec spec
```
teceler, line 56: Is there a difference between `rspec spec` and `rspec`?

teceler (approved):
Overall this looks good! 

I assume the postgres and redis options do useful things.

.github/workflows/rspec.yml
Throne3d: We had this in the Travis one! I think it adds the full output of the schema load rather than skipping some parts but I'm not sure honestly – Rails says:

```
   -t, --trace=[OUT]                Turn on invoke/execute tracing, enable full backtrace. OUT can be stderr (default) or stdout.
```

.github/workflows/rspec.yml
Throne3d: Not sure – this is again from Travis, but I think it should work with or without?

.github/workflows/rspec.yml
teceler: Well, 'we had it on travis' is a good reason tbh

.github/workflows/rspec.yml
Throne3d: Not yet unfortunately! We might be able to load a docker image of it maybe, and hook up to that, but that might require some awkward X server forwarding.

Marri (approved):
:turtle:

PR #1525, opened 2021-05-24
Upgrade to sprockets 4

Throne3d:
Migration guide available here: https://github.com/rails/sprockets/blob/master/UPGRADING.md

Basically, they're moving away from the practice of auto-compiling everything under assets/, and instead expect a manifest.js file to know what to compile. It also seems like it's a bit picky with some unexpected asset files like .xml or .json, so we had to reference the relevant favicon files directly to have them compiled.


Throne3d:
Weird - the test seems to have flaked and died here? Retrying

PR #1527, opened 2021-05-26
Use stylelint rather than sass-lint for scss linting

teceler:
Sass-lint is unmaintained


teceler:
Sass-lint rules I cannot currently replicate in stylelint:
- no-misspelled-properties
- bem-depth
- border-zero
- property-sort-order (disabled anyway)
- shorthand-values (shorthand-property-no-redundant-values is stricter than the setting we were using)

Sass-lint rules I have partially replicated in stylelint:
- pseudo-element (sass-lint rule prevents single-colon classes)
- space-around-operator (in most places but not in all functions)

Rules I think I have successfully replicated, but am not entirely sure
- brace-style

Throne3d:
This is done now we've migrated to GitHub Actions, right? Should we close this? 🙌

teceler:
> This is done now we've migrated to GitHub Actions, right? Should we close this? raised_hands

We're still using sass-lint and -- I think on codeclimate? I do have the reviewdog PR in here though I think yes

Throne3d:
Oh, oops, I think I misread where we had the stylelint check and thought it was already merged! If we rebase this maybe it can merge? I don't remember if this was blocked on something (since it's marked as Draft).

Throne3d:
About rules we can't replicate:

- no-misspelled-properties - should be possible with [property-no-unknown](https://stylelint.io/user-guide/rules/list/property-no-unknown/)
- [bem-depth](https://github.com/sasstools/sass-lint/blob/develop/docs/rules/bem-depth.md) - BEM seems to be a special way of writing selectors, which we don't use, so I'm fine ignoring this
- [border-zero](https://github.com/sasstools/sass-lint/blob/develop/docs/rules/border-zero.md) - don't personally mind about this
- property-sort-order (disabled anyway) - seems fine to ignore
- shorthand-values - I don't mind what we do here!

I think given the property-no-unknown lint we should be good? We could add it to the config and see how it goes!

Throne3d (approved):

PR #1528, opened 2021-06-07
Logout confirmation button

l1n:


app/views/layouts/application.haml
```diff
@@ -86,7 +86,7 @@
                 %li= link_to 'Favorites', favorites_path
                 %li= link_to "Post", new_post_path
               = link_to logout_path, method: :delete, id: :"header-logout" do
-                = button_tag "Log out", class: 'button'
+                = button_tag "Log out", class: 'button', :onclick => 'return confirm("Press OK to confirm logout");'
```
Throne3d, line 89: It looks like [`button_tag` accepts a simple `confirm:` param](https://apidock.com/rails/v6.0.0/ActionView/Helpers/FormTagHelper/button_tag), like `confirm: "Press OK to confirm logout"` instead of this onclick param – do you know if that would work?

l1n: Might, I'm not sure if the tests will work though - `page.accept_confirm` is specific to the confirm dialogue and I'm not sure if this actually creates a dialogue

app/views/layouts/application.haml
```diff
@@ -87,7 +87,7 @@
                 %li= link_to 'Favorites', favorites_path
                 %li= link_to "Post", new_post_path
               = link_to logout_path, method: :delete, id: :"header-logout" do
-                = button_tag "Log out", class: 'button'
+                = button_tag "Log out", class: 'button', :confirm: 'Press OK to confirm logout'
```
Throne3d, line 90: (Shouldn't have the leading `:`)

```suggestion
                = button_tag "Log out", class: 'button', confirm: 'Press OK to confirm logout'
```

app/views/layouts/application.haml
```diff
@@ -87,7 +87,7 @@
                 %li= link_to 'Favorites', favorites_path
                 %li= link_to "Post", new_post_path
               = link_to logout_path, method: :delete, id: :"header-logout" do
```
Throne3d, line 89: Ooh, okay, so apparently this confirm helper has to (1) be inside a `data: { ... }` object, and (2) works on the link since that's where we actually define the logout redirect (this button is a no-op). This should work (apparently I can't do a bulk suggestion on GitHub because there's a deleted line here, which is a bit weird, but applying both suggestions seemed to work locally).

```suggestion
              = link_to logout_path, method: :delete, id: :"header-logout", data: { confirm: 'Press OK to confirm logout' } do
```

app/views/layouts/application.haml
```diff
@@ -87,7 +87,7 @@
...
-                = button_tag "Log out", class: 'button'
+                = button_tag "Log out", class: 'button', confirm: 'Press OK to confirm logout'
```
Throne3d, line 90:
```suggestion
                = button_tag "Log out", class: 'button'
```

Throne3d (commented):
My bad, after this final one and a rebase we should be good to merge 🤦

app/views/layouts/application.haml
```diff
@@ -86,7 +86,7 @@
                 %li= link_to "Galleries", user_galleries_path(current_user)
                 %li= link_to 'Favorites', favorites_path
                 %li= link_to "Post", new_post_path
-              = link_to logout_path, method: :delete, id: :"header-logout" do
+              = link_to logout_path, method: :delete, id: :"header-logout", data: { confirm: 'Press OK to confirm logout' do
```
Throne3d, line 89: (Missing the trailing curly brace)

```suggestion
              = link_to logout_path, method: :delete, id: :"header-logout", data: { confirm: 'Press OK to confirm logout' } do
```

Throne3d: I couldn't seem to push commits when I tried this before, otherwise I'd be pushing to the branch and trying to get this merged XD I might have something misconfigured, not sure if I'm meant to be able to edit it directly

Throne3d (approved):
Looks like the checks all ran fine! Nice, thanks 🙌

PR #1532, opened 2021-06-25
Fix board search by author UI

teceler:
It did not have javascript, and it should have, not sure how that got missed but here we are.


teceler:
Tested this locally, boards (and posts) search by author seem to work fine.

Marri (approved):
:turtle: lint check to be fixed separately

PR #1541, opened 2021-06-26
Cache badge counts

### LINKED ISSUE #1533: Cache news/message/report per user
(so we're not hitting the db for them every page load)

teceler:
Cache unread news count and message count, and last daily report read
Fixes #1533
Depends on #1537

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -11,10 +11,13 @@ class Message < ApplicationRecord
   before_validation :set_thread_id, :remove_deleted_recipient
   before_create :check_recipient
   after_create :notify_recipient
+  after_commit :invalidate_caches
 
   scope :ordered_by_id, -> { order(id: :asc) }
   scope :ordered_by_thread, -> { order(thread_id: :asc, id: :desc) }
   scope :unread, -> { where(unread: true) }
+
+  CACHE_VERSION = 1
 
   def visible_to?(user)
     user_ids.include?(user.id)
@@ -58,6 +61,16 @@ class Message < ApplicationRecord
     msg.save
   end
 
+  def self.unread_count_for(user)
+    Rails.cache.fetch(Message.cache_string_for(user.id), expires_in: 1.day) do
+      user.messages.unread.count
+    end
+  end
+
+  def self.cache_string_for(user_id)
+    "#{Rails.env}.#{CACHE_VERSION}.unread_message_count.#{user_id}"
+  end
+
   private
 
   def set_thread_id
@@ -89,5 +102,9 @@ class Message < ApplicationRecord
   def remove_deleted_recipient
     return unless recipient&.deleted?
     self.recipient = nil
+  end
+
+  def invalidate_caches
+    Rails.cache.delete(Message.cache_string_for(recipient.id))
   end
 end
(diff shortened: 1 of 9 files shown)


app/models/message.rb
```diff
@@ -58,6 +61,16 @@ def self.send_site_message(user_id, subject, message)
     msg.save
   end
 
+  def self.unread_count_for(user)
+    Rails.cache.fetch(Message.cache_string_for(user.id), expires_in: 1.month) do
```
Throne3d, line 65: So, I know we should invalidate this correctly with callbacks and everything, buuuut maybe we should reduce this to like `expires_in: 1.day` so that if something does go wrong with invalidation it'll be resolved on a short time scale? We should still get a good cache hit rate, since a typical user will load the site a lot more than just once in a single day, but I'm just a bit wary of such a long cache expiry here.

Throne3d: (Same for news)

app/models/news.rb
```diff
@@ -30,9 +32,15 @@ def mark_read(user)
...
+      view = NewsView.find_by(user: user)
+      view ? News.where('id > ?', view.news_id).count : News.count
+    end
+  end
+
+  private
 
-    News.where('id > ?', view.news_id).count
+  def invalidate_caches
+    Rails.cache.delete_matched("#{Rails.env}.#{NewsView::CACHE_VERSION}.unread_news_count.*")
```
Throne3d, line 44: Could we keep this on `NewsView` as a class method that we call from here? That way we can keep the patterns for each cache string centralized on a single model where possible (making it easier to spot accidental divergences).

I think a `NewsView.invalidate_all_caches` would be good!

app/models/message.rb
teceler: hmm, yeah, reducing that seems reasonable since it should be cleared within a day anyway.

teceler: the same *doesn't* apply to news, though, I'm more uncertain on that.

app/views/layouts/application.haml
```diff
@@ -79,8 +79,8 @@
                 %li
                   = link_to messages_path do
                     Inbox
-                    - if current_user.messages.unread.exists?
-                      %span.badge.badge-primary= current_user.messages.unread.count
+                    - if Message.unread_count_for(current_user) > 0
+                      %span.badge.badge-primary= Message.unread_count_for(current_user)
```
Throne3d, line 83: Since we hit the cache for this (an external service call), we should probably just do it once:
```suggestion
                    - unread_count = Message.unread_count_for(current_user)
                    - if unread_count > 0
                      %span.badge.badge-primary= unread_count
```

Throne3d (approved):
LGTM!

app/models/message.rb
Throne3d: The main metric we probably care about for this is cache hit rate – how many times are we doing queries unnecessarily after this change? Even if we reduce the caching expiry to 1 day, we'll still only load this query once per day per user, which is a _lot_ less than our current number of page loads, so the caching still performs a great job if we drop it down to expiring daily. We could reduce it further, to once per month per user, by using this longer expiry, but as these are cheap queries, it probably doesn't help much on this margin, right?

It's not a huge deal either way, I don't think, since the invalidation is tested and looks good, though 🙂

PR #1542, opened 2021-06-27
Improve NotifyFollowersOfNewPostJob

teceler:
- [x] Notify of PostAuthors on post creation
- [ ] Notify when access is granted
- [ ] Notify when a post becomes public/visible to registered users
- [ ] Notify when added as a coauthor


Throne3d:
Hm, looks like this also addresses #732 - could we rebase it + see what's necessary to get it over the finish line? =)

PR #1545, opened 2021-07-05
Add more tests

teceler:
This leaves the only sections of our code uncovered (other than probably some inline ifs; the only sections simplecov detects, at least) some of the branches in GalleriesController#icon

### DIFF
--- a/spec/models/reply_spec.rb
+++ b/spec/models/reply_spec.rb
@@ -186,4 +186,22 @@ RSpec.describe Reply do
       expect(post.reload.last_reply_id).to eq(replies[1].id)
     end
   end
+
+  describe "Writable" do
+    describe "#name" do
+      let(:user) { create(:user) }
+      let(:character) { create(:character, user: user) }
+      let(:reply) { create(:reply, character: character, user: user) }
+
+      it "works without alias" do
+        expect(reply.name).to eq(character.name)
+      end
+
+      it "works with alias" do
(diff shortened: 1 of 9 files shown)


spec/models/reply_spec.rb
```diff
@@ -173,4 +173,35 @@
...
+  describe "#destroy_subsequent_replies" do
+    it "works" do
+      post = create(:post)
+      replies = create_list(:reply, 2, post: post)
+      reply = create(:reply, post: post)
+      create_list(:reply, 2, post: post)
+      expect { reply.send(:destroy_subsequent_replies) }.to change { Reply.count }.by(-3)
+      expect(Reply.where(id: replies.map(&:id)).count).to eq(2)
+      expect(Reply.find_by(id: reply.id)).not_to be_present
+      expect(post.reload.last_reply_id).to eq(reply.id)
```
Throne3d, line 186: This seems strange, but maybe I'm misreading something – the reply with `id: reply.id` doesn't exist (per line 185), but here, the `last_reply_id` of the post is this reply? Isn't that a broken foreign key?

Throne3d (commented):
Looks good beyond that question, I think!

spec/models/reply_spec.rb
teceler: Er. yeah it should not in fact do that, which makes me wonder why tests are passing!

Throne3d (approved):

PR #1552, opened 2021-07-12
Add ReportsHelper

teceler:
* Moves helper-only methods from ReportsController to ReportsHelper
* Tests those methods

(extracted from #1549)


Marri (approved):
:turtle: assuming the ReportsHelper is included automatically

PR #1556, opened 2021-07-14
Have rubocop.yml and .slugignore ignore backstop rather than wraith

teceler:
And clean up wraith in .gitignore


Marri (approved):
:turtle:

PR #1561, opened 2021-07-16
Add PaperTrail

teceler:
- [x] Add tables
- [x] Make post#delete_history work
- [x] Make reply#restore work
- [x] Make audit counts work
- [x] Make read_time_for work
- [ ] Make written#history work
- [ ] Test written#history
- [ ] Write conversion script
- [ ] Test conversion script


spec/controllers/posts_controller_spec.rb
```diff
@@ -1139,21 +1139,18 @@
...
-
-      replies = Audited.audit_class.as_user(post.user) do
+    it "calculates audits", versioning: true do
+      replies = Version.as_user(post.user) do
         create_list(:reply, 6, post: post, user: post.user)
       end
 
-      Audited.audit_class.as_user(post.user) do
+      Version.as_user(post.user) do
         replies[1].touch # rubocop:disable Rails/SkipsModelValidations
```
teceler, line 1148: Note to future self: this is the reason for the test failure, right here, Papertrail seems to be creating an audit for the touch. possibly an after_touch callback?

PR #1578, opened 2021-10-19
force logo img height to stop reflow on page load

l1n:


app/views/layouts/application.haml
```diff
@@ -71,7 +71,7 @@
                 %span &nbsp;|&nbsp;
                 = link_to "Sign up", new_user_path
             = link_to root_path do
-              = image_tag "layouts/logo.png", alt: 'Glowfic Constellation home'
+              = image_tag "layouts/logo.png", alt: 'Glowfic Constellation home', height: '20px'
```
Throne3d, line 74: It looks like this is 30px when I load up the site itself – was this a typo? Also, when I try this myself, even with `30`, it seems to change the size of the logo unless we change the `padding` CSS declaration to `margin`.

(Aside: MDN says this should be declared without `px`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img#attr-height)

Throne3d: Ooh, it's rendered as 20px on mobile due to the media query. Even so, this attribute seems like it's meant to be based on the intrinsic size of the image, which seems to be 30px tall? I think that and the padding -> margin CSS change, or putting the padding on the link instead of the image, would solve it, but we might need to check how it renders in practice to be sure.

l1n: Sorry, missed this for ~4 years. Pushed [7743961be](https://github.com/l1n/glowfic/commit/7743961be):

- Logo is intrinsically 90×30, so `height: 30` (no `px` per MDN, you were right) and `width: 90` match the actual file.
- The HTML attr is what reserves the box before the image bytes arrive, so it kills the desktop reflow.
- Mobile still snaps from 30 → 20 when the media-query CSS kicks in, but that's a small visual change inside an already-painted box, not the multi-line shift the original PR was trying to address.

Re padding → margin: with `height` and `width` set, the browser allocates the correct box up front; the existing `#logo img { padding: 0.5em 0.5em 0.3em 0 }` shifts content within that box rather than changing its outer size, so I left it. Happy to swap to margin if you'd rather a clean separation; minor either way.

PR #1580, opened 2021-11-07
Expand ignore function to more post lists

teceler:
Allow ignoring posts and boards to hide them from favorites, daily report, and recently updated

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -11,7 +11,9 @@ class PostsController < WritableController
   before_action :editor_setup, only: [:new, :edit]
 
   def index
-    @posts = posts_from_relation(Post.ordered, show_blocked: !!params[:show_blocked])
+    @posts = Post.ordered
+    @posts = @posts.not_ignored_by(current_user) if current_user&.hide_from_all
+    @posts = posts_from_relation(@posts, show_blocked: !!params[:show_blocked])
     @page_title = 'Recent Threads'
   end
 
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -45,21 +45,18 @@ class PostsController < WritableController
 
   def unread
     @started = (params[:started] == 'true') || (params[:started].nil? && current_user.unread_opened)
-    @posts = Post.joins("LEFT JOIN post_views ON post_views.post_id = posts.id AND post_views.user_id = #{current_user.id}")
-    @posts = @posts.joins("LEFT JOIN board_views on board_views.board_id = posts.board_id AND board_views.user_id = #{current_user.id}")
+    @posts = Post.not_ignored_by(current_user)
     @posts = @posts.where.not(post_views: { read_at: nil }) if @started
 
-    # post view does not exist and (board view does not exist or post has updated since non-ignored board view read_at)
-    no_post_view = @posts.where(post_views: { user_id: nil })
+    # post view does not exist and (board view does not exist or post has updated since board view read_at)
+    no_post_view = @posts.where(post_views: { id: nil })
     updated_since_board_read = no_post_view.where(board_views: { read_at: nil })
       .or(no_post_view.where("date_trunc('second', board_views.read_at) < date_trunc('second', posts.tagged_at)"))
-      .where(board_views: { ignored: false })
     no_post_view = no_post_view.where(board_views: { user_id: nil }).or(updated_since_board_read)
 
-    # post view exists and post has updated since non-ignored post view read_at and (board view does not exist or is not ignored)
-    with_post_view = @posts.where(post_views: { ignored: false }) # non-existant post-views will return nil here
-    with_post_view = with_post_view.where(board_views: { user_id: nil }).or(with_post_view.where(board_views: { ignored: false }))
-    with_post_view = with_post_view.where(post_views: { read_at: nil })
+    # post view exists and post has updated since post view read_at
+    with_post_view = @posts.where.not(post_views: { id: nil })
+    with_post_view = with_post_view.where(post_views: { read_at: nil }) # possible if someone ignores and then unignores a post
       .or(with_post_view.where("date_trunc('second', post_views.read_at) < date_trunc('second', posts.tagged_at)"))
 
     @posts = with_post_view.or(no_post_view)
--- a/app/views/users/edit.haml
+++ b/app/views/users/edit.haml
@@ -69,6 +69,9 @@
           %br
           = f.check_box :visible_unread, class: 'width-15 vmid'
           = f.label :visible_unread, "Outline when jumping to unread"
+          %br
+          = f.check_box :hide_from_all, class: 'width-15 vmid'
+          = f.label :hide_from_all, "Hide hidden posts from favorites and daily report as well"
       %tr
         %th.sub Daily Report
         %td{class: cycle('even', 'odd')}
--- a/app/views/users/edit.haml
+++ b/app/views/users/edit.haml
@@ -71,7 +71,7 @@
           = f.label :visible_unread, "Outline when jumping to unread"
           %br
           = f.check_box :hide_from_all, class: 'width-15 vmid'
-          = f.label :hide_from_all, "Hide hidden posts from favorites and daily report as well"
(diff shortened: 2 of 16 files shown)


app/controllers/posts_controller.rb
```diff
@@ -45,20 +47,17 @@ def owed
...
     updated_since_board_read = no_post_view.where(board_views: { read_at: nil })
       .or(no_post_view.where("date_trunc('second', board_views.read_at) < date_trunc('second', posts.tagged_at)"))
-      .where(board_views: { ignored: false })
     no_post_view = no_post_view.where(board_views: { user_id: nil }).or(updated_since_board_read)
 
-    # post view exists and post has updated since non-ignored post view read_at and (board view does not exist or is not ignored)
-    with_post_view = @posts.where(post_views: { ignored: false }) # non-existant post-views will return nil here
-    with_post_view = with_post_view.where(board_views: { user_id: nil }).or(with_post_view.where(board_views: { ignored: false }))
+    # post view exists and post has updated since post view read_at
+    with_post_view = @posts.where.not({ user_id: nil })
```
Throne3d, line 60: I'm not sure I understand this one - posts where the post has an author? 🤔

Throne3d (approved):
Don't think I understand the posts#unread bit, but the rest of this looks good!

app/controllers/posts_controller.rb
teceler: good... question. it's been a while since I wrote this and I do *not* remember.

Throne3d: todo: need to look into this in more detail

Throne3d (changes requested):
(see comment, also needs a rebase)

app/controllers/posts_controller.rb
teceler: okay so I *think* what this is *trying* to do is select posts that have a post view and should be `@posts.where.not(post_views: { user_id: nil })`. I'm not sure that actually does the right thing but I think that's what happened there.

app/controllers/posts_controller.rb
```diff
@@ -45,22 +47,19 @@ def owed
...
 
-    # post view exists and post has updated since non-ignored post view read_at and (board view does not exist or is not ignored)
-    with_post_view = @posts.where(post_views: { ignored: false }) # non-existant post-views will return nil here
-    with_post_view = with_post_view.where(board_views: { user_id: nil }).or(with_post_view.where(board_views: { ignored: false }))
-    with_post_view = with_post_view.where(post_views: { read_at: nil })
-      .or(with_post_view.where("date_trunc('second', post_views.read_at) < date_trunc('second', posts.tagged_at)"))
+    # post view exists and post has updated since post view read_at
+    with_post_view = @posts.where.not(post_views: { id: nil })
+    with_post_view = with_post_view.where(post_views: { read_at: nil }) # possible if someone ignores and then unignores a post, I think
+    with_post_view = with_post_view.or(@posts.where("date_trunc('second', post_views.read_at) < date_trunc('second', posts.tagged_at)"))
```
Throne3d, line 62:
```suggestion
    with_post_view = with_post_view.where(post_views: { read_at: nil }) # possible if someone ignores and then unignores a post
      .or(with_post_view.where("date_trunc('second', post_views.read_at) < date_trunc('second', posts.tagged_at)"))
```
This should be equivalent (the `@posts.where(date_trunc...)` will give NULL if post_views doesn't exist, which is falsy) but makes more sense in my brain, and is less of a change from how the code was before - of posts that have a view, we want either the ones where there's no read_at date, or the ones where they were read before the latest tag.

With the condition changed like this, `@posts` is all not-ignored-posts, not just the ones with views, so the logic of the condition is more complex to be sure of

app/controllers/posts_controller.rb
```diff
@@ -45,22 +47,19 @@ def owed
 
   def unread
     @started = (params[:started] == 'true') || (params[:started].nil? && current_user.unread_opened)
-    @posts = Post.joins("LEFT JOIN post_views ON post_views.post_id = posts.id AND post_views.user_id = #{current_user.id}")
-    @posts = @posts.joins("LEFT JOIN board_views on board_views.board_id = posts.board_id AND board_views.user_id = #{current_user.id}")
+    @posts = Post.not_ignored_by(current_user)
```
Throne3d, line 50: This helps simplify things a lot! Nice =)

db/migrate/20211023154550_add_hide_from_all_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddHideFromAllToUser < ActiveRecord::Migration[5.2]
```
Throne3d, line 1:
```suggestion
class AddHideFromAllToUser < ActiveRecord::Migration[7.1]
```
Wow, we added this a long time ago :eyes: 7.1 should be a valid Migration config here

spec/controllers/posts/unread_spec.rb
```diff
@@ -91,42 +91,56 @@ def setup_posts
     end
 
     it "manages board/post read time mismatches" do
```
Throne3d, line 93: I don't think this changes the behavior here, does it? It just adjusts the timestamps (presumably to make the logic be less confusing)

spec/models/post_spec.rb
```diff
@@ -1330,6 +1330,33 @@
...
+      read_post.mark_read(reader)
+
+      unread_ignored.ignore(reader)
+      read_ignored.ignore(reader)
+      read_ignored.mark_read(reader)
+
+      ignored_board.ignore(reader)
+
+      expect(Post.count).to eq(5)
+      expect(Post.not_ignored_by(reader).count).to eq(2)
```
Throne3d, line 1355:
```suggestion
```
I think this line is redundant with the next one, and if it ends up being false (e.g. we actually see 3), it'll obscure the actual error a bit, right?

spec/models/post_spec.rb
```diff
@@ -1330,6 +1330,33 @@
...
+      expect(Post.not_ignored_by(reader)).to match_array([unread_post, read_post])
```
Throne3d, line 1356: IMO it'd be good to add
```suggestion
      expect(Post.not_ignored_by(create(:user)).count).to eq(5)
```
to validate the not_ignored_by join is properly scoping based on user (and not ignoring it for everyone). The code logic is definitely good, just nice to add it in a test

Throne3d, line 1356: IMO it'd be good to add
```suggestion
      expect(Post.not_ignored_by(reader)).to match_array([unread_post, read_post])
      expect(Post.not_ignored_by(create(:user)).count).to eq(5)
```
to validate the not_ignored_by join is properly scoping based on user (and not ignoring it for everyone). The code logic is definitely good, just nice to add it in a test

app/views/users/edit.haml
```diff
@@ -69,6 +69,9 @@
           %br
           = f.check_box :visible_unread, class: 'width-15 vmid'
           = f.label :visible_unread, "Outline when jumping to unread"
+          %br
+          = f.check_box :hide_from_all, class: 'width-15 vmid'
+          = f.label :hide_from_all, "Hide hidden posts from favorites and daily report as well"
```
Throne3d, line 74:
```suggestion
          = f.label :hide_from_all, "Hide hidden posts from favorites, recently updated and daily report as well"
```

Throne3d (approved):

app/controllers/posts_controller.rb
Throne3d: Aha, that makes sense! Nice

Throne3d:
A few small comments, but nothing that's a blocker (afaict). Looks good!

spec/controllers/posts/unread_spec.rb
teceler: Yeah it *should* just timecop things to make them not happen at the same time (or manually use `at_time`), unless i messed up somewhere.

Throne3d (approved):
LGTM! But branch coverage is unhappy now apparently 🫠

Throne3d (approved):

Throne3d (approved):

PR #1583, opened 2021-12-16
Organize helper methods

teceler:
* Creates PostHelper and IconHelper
* Moves icon display methods from ApplicationHelper and WritableHelper to IconHelper
* Moves methods specific to posts from WritableHelper and ApplicationHelper to PostHelper
* Moves methods for handling post/reply text from ApplicationHelper to WritableHelper


app/helpers/icon_helper.rb
```diff
@@ -0,0 +1,63 @@
+module IconHelper
+  ICON = 'icon'.freeze
+  NO_ICON = 'No Icon'.freeze
+  NO_ICON_URL = 'icons/no-icon.png'.freeze
+  TIME_FORMAT = '%b %d, %Y %l:%M %p'.freeze
```
Throne3d, line 5: (is this used here?)
```suggestion
```

app/helpers/writable_helper.rb
```diff
@@ -7,69 +7,6 @@ def unread_warning
     tag.a('(New tab)', href: unread_path(@post), class: 'unread-warning', target: '_blank')
   end
```
Throne3d, line 8: This can still call things from icon_helper, right? (Like unread_path)

(Edit: Answered in Discord - it can!)

app/mailers/user_mailer.rb
```diff
@@ -1,4 +1,6 @@
 class UserMailer < AsyncMailer
+  add_template_helper WritableHelper
```
Throne3d, line 2: (apparently this is getting removed in Rails 6.1 and should be done with `helper WritableHelper` in the future? I wonder if that already works for us - https://stackoverflow.com/questions/65528227/the-method-add-template-helper-does-not-seem-to-work-in-rails-6-1)

Throne3d (approved):

PR #1584, opened 2021-12-16
Test helpers more

teceler:
Test uncovered helper branches

### DIFF
--- a/app/helpers/icon_helper.rb
+++ b/app/helpers/icon_helper.rb
@@ -6,12 +6,10 @@ module IconHelper
   CHAR_ICON_FAKE = 'char-access-icon char-access-fake pointer'.freeze
 
   def icon_tag(icon, **args)
-    return '' if icon.nil?
     icon_mem_tag(icon.url, icon.keyword, **args)
   end
 
   def icon_mem_tag(url, keyword, **args)
-    return '' if url.nil?
     klass = ICON
     klass += ' pointer' if args.delete(:pointer)
     if (supplied_class = args.delete(:class))
--- a/app/helpers/icon_helper.rb
+++ b/app/helpers/icon_helper.rb
@@ -6,10 +6,12 @@ module IconHelper
   CHAR_ICON_FAKE = 'char-access-icon char-access-fake pointer'.freeze
 
   def icon_tag(icon, **args)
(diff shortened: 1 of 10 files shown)


app/helpers/icon_helper.rb
```diff
@@ -6,12 +6,10 @@ module IconHelper
   CHAR_ICON_FAKE = 'char-access-icon char-access-fake pointer'.freeze
 
   def icon_tag(icon, **args)
-    return '' if icon.nil?
```
Throne3d, line 9: Seems this is necessary when doing character replaces - do we need to remove these guards we have here + below as part of this MR?

app/helpers/writable_helper.rb
```diff
@@ -44,7 +44,6 @@ def split_paragraphs_largebreak(text)
   # https://apidock.com/rails/ActionView/Helpers/TextHelper/simple_format
   def simple_format_largebreak(text, options={})
     wrapper_tag = options.fetch(:wrapper_tag, :p)
-    text = sanitize(text) if options.fetch(:sanitize, true)
```
Throne3d, line 47: Can we remove `sanitize: false` from the `sanitize_written_content` thing below accordingly?

Seems to be the only place we use this, so makes sense to me

We can probably remove all the options here as a result actually, and just hard-code wrapper_tag as :p

Throne3d (approved):

app/helpers/writable_helper.rb
teceler: Wait why did I even remove that in the first place

Throne3d: I'd guess because this is only called in one place, and always with the same param, but we can add it back to get this merged! =)

app/helpers/writable_helper.rb
```diff
@@ -44,7 +44,6 @@ def split_paragraphs_largebreak(text)
...
     paragraphs = split_paragraphs_largebreak(text)
```
Throne3d, line 47:
```suggestion
    text = sanitize(text) if options.fetch(:sanitize, true)
    paragraphs = split_paragraphs_largebreak(text)
```

Throne3d (approved):

PR #1587, opened 2021-12-18
Fix IconHelper#dropdown_icons test non-determinism

Throne3d:


teceler (approved):
Looks good assuming it fixes the bug! :turkey:

PR #1592, opened 2021-12-19
Use Selenium Docker image for CI

Throne3d:
Builds on #1522, trying to remove the Chromium requirement


Throne3d:
Nah, this'll have to connect back to the host for the HTTP connection. Sounds awkward!

PR #1594, opened 2021-12-20
Fix minor rubocop lints

teceler:
- Adds whitespace
- Uses update! in tests
- Disables warning on post split script
- Uses expand_path __dir__ pattern
- Declares `dependent: :destroy` on board writers and cameos
- Uses `Rails.env.test?` rather than `Rails.env == 'test'`

### DIFF
--- a/app/models/board.rb
+++ b/app/models/board.rb
@@ -13,9 +13,9 @@ class Board < ApplicationRecord
 
   has_many :board_authors, inverse_of: :board, dependent: :destroy
   has_many :authors, class_name: 'User', through: :board_authors, source: :user, dependent: :destroy
-  has_many :board_writers, -> { where(cameo: false) }, class_name: 'BoardAuthor', inverse_of: :board
+  has_many :board_writers, -> { where(cameo: false) }, class_name: 'BoardAuthor', inverse_of: :board, dependent: :destroy
   has_many :writers, class_name: 'User', through: :board_writers, source: :user, dependent: :destroy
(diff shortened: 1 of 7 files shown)


app/models/board.rb
```diff
@@ -13,9 +13,9 @@ class Board < ApplicationRecord
 
   has_many :board_authors, inverse_of: :board, dependent: :destroy
   has_many :authors, class_name: 'User', through: :board_authors, source: :user, dependent: :destroy
-  has_many :board_writers, -> { where(cameo: false) }, class_name: 'BoardAuthor', inverse_of: :board
+  has_many :board_writers, -> { where(cameo: false) }, class_name: 'BoardAuthor', inverse_of: :board, dependent: :destroy
   has_many :writers, class_name: 'User', through: :board_writers, source: :user, dependent: :destroy
-  has_many :board_cameos, -> { where(cameo: true) }, class_name: 'BoardAuthor', inverse_of: :board
+  has_many :board_cameos, -> { where(cameo: true) }, class_name: 'BoardAuthor', inverse_of: :board, dependent: :destroy
```
Throne3d, line 18: I'm a little curious if this changes anything – they should be through the `:board_authors` class so I wouldn't expect this to do anything, probably?

But seems OK!

Throne3d (approved):

app/models/board.rb
teceler: It probably doesn't, although we've had some weirdness with through joins and dependent destroy before, so who knows. it's clearer than setting dependent: false on them tho.

PR #1598, opened 2021-12-22
Clean up javascript lints

teceler:


app/assets/javascripts/galleries/add_existing.js
```diff
@@ -47,7 +47,7 @@ $(document).ready(function() {
 });
 
 $(window).on('beforeunload', function() {
-  if (skipWarning || imageIds.length === 0) return;
+  if (skipWarning || imageIds.length === 0) { return false; }
```
Throne3d, line 50: Not clear on the motivation here (either for the braces or the false value). What was the original complaint?

app/assets/javascripts/galleries/add_new.js
```diff
@@ -1,44 +1,38 @@
...
 var done = 0;
 var total = 0;
 var failed = 0;
 
 $(document).ready(function() {
   fixButtons();
   $(".icon-row td:has(input)").each(function() {
-    $(this).keydown(processDirectionalKey);
+    $(this).keydown(function(event) {
+      if (event.which < 37 || event.which > 40) { return; } // skip if not a directional key
```
Throne3d, line 12: (Forced braces are weird and these are magic numbers – can we use the .indexOf thing like before?)

app/assets/javascripts/galleries/add_existing.js
teceler: complaint was about returning both a value and not returning something from a function, brackets are because I find that more readable on javascript ifs. ...I think that's the one I was most concerned it would break something tho.

app/assets/javascripts/galleries/add_new.js
```diff
@@ -52,6 +46,18 @@ function processDirectionalKey(event) {
...
+  $(this).closest('td').prev().find('input').focus();
+  return true;
+}
+
+function processKeyRight(caret, length) {
+  if (caret <= length) { return false; }
+  $(this).closest('td').next().find('input').focus();
+  return true;
+}
+
```
Throne3d, line 60: I'd ignore the linter on extracting these

app/assets/javascripts/galleries/add_new.js
```diff
@@ -117,6 +123,7 @@ function cleanUpRows() {
         anySet = true;
         return false;
       }
+      return false;
```
Throne3d, line 126: Changes the behavior – $.each stops when you return false, and we want to intentionally return undefined (or true if we really need to return something) to make sure it continues.

app/assets/javascripts/galleries/add_new.js
teceler: yeah I pulled these to here because of the complexity complaint and didn't want to redeclare the variables but I could make a [37, 38, 39, 40] list probably?

Or, yes, we could ignore it, but it wasn't *wrong* that that function was hideously complicated.

app/assets/javascripts/galleries/add_new.js
teceler: *huh*. Okay.

app/assets/javascripts/galleries/uploader.js
```diff
@@ -129,3 +130,8 @@ function deleteUnusedIcons(keys) {
     $.authenticatedPost('/api/v1/icons/s3_delete', {s3_key: key});
   });
 }
+
+function isPolicyExpired(responseText) {
+  if (responseText.includes("Invalid according to Policy: Policy expired.")) { return true; }
+  return responseText.includes("Idle connections will be closed.");
```
Throne3d, line 136:
```suggestion
  return responseText.includes("Invalid according to Policy: Policy expired.") || responseText.includes("Idle connections will be closed.");
```
(can be split across lines if we have to, but I forget what JavaScript's rules are for continuation)

teceler: Yeah I split it like that because the line was long and I was extracting it anyway :+1:

app/assets/javascripts/posts/editor.js
```diff
@@ -406,23 +412,28 @@ function setGalleriesAndDefault(galleries, defaultIcon) {
...
-    iconSelectBox.append(galleryString(galleries[j], multiGallery));
-  }
+  setGalleries(galleries);
+
+  if (defaultIcon && shownIcons.indexOf(defaultIcon.id) < 0) { iconSelectBox.append(iconString(defaultIcon)); }
 
-  // If no default and no icons in any galleries, remove pointer
-  if (!defaultIcon && shownIcons.length === 0) {
+  // If no icons, remove pointer
+  if (shownIcons.length === 0) {
```
Throne3d, line 420: Not clear on this change 🤔 If there's a default icon and no gallery, won't this now fail to trigger?

teceler: It wouldn't trigger with a default icon before, either, though

Throne3d: The inverse, rather, there's _no_ default icon and no gallery, my bad

teceler: If there's no default icon the length should be 0? I think?

app/assets/javascripts/galleries/add_existing.js
Throne3d: Yeah, events like this are always awkward and returning a string is even now deprecated (but the alternatives aren't well supported). Returning false might do something weird in some browsers. I'd leave this and ignore the linter on it.

app/assets/javascripts/galleries/add_new.js
Throne3d: I'd move the 37, 38, 39, 40 magic numbers to a global constant and refer to them from both functions

app/assets/javascripts/galleries/add_existing.js
Throne3d: (also FWIW I never see braces on single-line if statements in JavaScript - it's actually kinda distracting to me)

app/assets/javascripts/galleries/uploader.js
Throne3d:
```suggestion
  return responseText.includes("Invalid according to Policy: Policy expired.")
    || responseText.includes("Idle connections will be closed.");
```
then? having an extra if takes a lot more time for me to parse here

app/assets/javascripts/posts/editor.js
Throne3d: Maaaybe? shownIcons is a global and so I'm not sure where all we touch it from =( this might be fine but I don't know for sure

Throne3d: Actually, why do we even remove the pointer here if there _is_ a default icon? This code is weird – I'd probably leave it until we can be confident in testing it

teceler: Reverted pending What Is This Code Even

app/assets/javascripts/galleries/add_new.js
```diff
@@ -52,6 +47,18 @@ function processDirectionalKey(event) {
...
 }
 
+function processKeyLeft(caret) {
+  if (caret !== 0) { return false; }
+  $(this).closest('td').prev().find('input').focus();
+  return true;
+}
+
+function processKeyRight(caret, length) {
+  if (caret <= length) { return false; }
```
Throne3d, line 57: Think this should be `caret < length` based on the previous condition being `caret >= length`

app/assets/javascripts/galleries/uploader.js
```diff
@@ -17,8 +18,9 @@ $(document).ready(function() {
     var usedUrls = $.map($('form.icon-upload').find('input[id$=_url]'), function(input) { return $(input).val(); });
     var uploadedUrls = $.map(uploadedIcons, function(value, key) { return key; });
     var unusedUrls = uploadedUrls.filter(function(x) { return usedUrls.indexOf(x) < 0; });
-    if (unusedUrls.length < 1) return true;
+    if (unusedUrls.length < 1) { return true; }
     deleteUnusedIcons($.map(unusedUrls, function(url) { return uploadedIcons[url]; }));
+    return false;
```
Throne3d, line 23: I think `return false` changes behavior here so the submit doesn't finish: https://api.jquery.com/on/#event-handler suggests that this will call stopPropagation + preventDefault, whereas the default return (undefined) doesn't have that behavior - https://github.com/jquery/jquery/blob/ace646f6e83e653f666ba715c200739f1cbdba52/src/event.js#L326

teceler: i... what... but... why...

Throne3d: Not sure! jQuery seems to use undefined as equivalent to true in this situation, and return false is a different behavior.

PR #1605, opened 2021-12-25
Enforce rubocop on more files

teceler:
Now covered:
* Seed files
* Most of the config files (but not the ones where we're monkey patching)
* Some bin files


bin/yarn
```diff
@@ -4,8 +4,8 @@ Dir.chdir(APP_ROOT) do
   begin
     exec "yarnpkg", *ARGV
   rescue Errno::ENOENT
-    $stderr.puts "Yarn executable was not detected in the system."
-    $stderr.puts "Download Yarn at https://yarnpkg.com/en/docs/install"
+    warn "Yarn executable was not detected in the system."
+    warn "Download Yarn at https://yarnpkg.com/en/docs/install"
```
Throne3d, line 8: Weird, I think this is Rails default

teceler: I can revert it and tell rubocop to ignore it if we object, but I figured it wasn't a big deal given we don't... use... yarn... anyway...

Throne3d: Probably OK, we just might end up reverting it with the next `rails app:update`

Throne3d (approved):
(Needs a rebase but) I forgot to approve this last time! Approved

PR #1607, opened 2021-12-25
Standardize spacing

teceler:
Enforces:
* Spaces inside hash literals
* Spaces around operators
* Spaces before block braces
* Spaces inside block braces
* Spaces inside range literals
* No space between -> and open parens in lambdas with conditions
* No space between hash/array/etc and left square bracket
* Aligned multiline if/unless conditions
* Symmetrical braces on multiline method calls


app/controllers/uploading_controller.rb
```diff
@@ -16,6 +16,7 @@ def set_s3_url
       success_action_status: '201',
       acl: 'public-read',
       content_type_starts_with: 'image/',
-      cache_control: 'public, max-age=31536000')
+      cache_control: 'public, max-age=31536000',
+    )
```
Throne3d, line 20: Think this is from the comma PR but it's not like it matters :D

Throne3d (approved):

app/controllers/uploading_controller.rb
teceler: The comma must have snuck in when I was rebasing xD

PR #1611, opened 2021-12-25
Enforce good practices and consistent style on tests

teceler:
* Uses Time.use_zone block rather than setting Time.zone
* Replaces instance variables in tests
* Uses numbers for http statuses
* Avoids expect(...).and_return(...) where it's not necessary

### DIFF
--- a/spec/jobs/application_job_spec.rb
+++ b/spec/jobs/application_job_spec.rb
@@ -29,6 +29,7 @@ RSpec.describe ApplicationJob do
 
     job = StubJob.new(2, :test)
     allow(job).to receive(:perform).and_raise(exc)
+    expect(job).to receive(:perform)
     begin
       job.perform_now
     rescue Exception # rubocop:disable Lint/RescueException
--- a/spec/jobs/application_job_spec.rb
+++ b/spec/jobs/application_job_spec.rb
@@ -28,7 +28,7 @@ RSpec.describe ApplicationJob do
     expect(ExceptionNotifier).to receive(:notify_exception).with(exc, data: { job: StubJob.name, args: [2, :test] })
 
     job = StubJob.new(2, :test)
-    expect(job).to receive(:perform).and_raise(exc)
+    allow(job).to receive(:perform).and_raise(exc)
     begin
       job.perform_now
     rescue Exception # rubocop:disable Lint/RescueException
--- a/spec/script/post_split_spec.rb
+++ b/spec/script/post_split_spec.rb
@@ -13,6 +13,7 @@ RSpec.describe "post_split" do # rubocop:disable RSpec/DescribeClass
 
     it "requires valid subject" do
       allow(STDIN).to receive(:gets).and_return('')
+      expect(STDIN).to receive(:gets)
       expect { split_post }.to raise_error(RuntimeError, 'Invalid subject')
     end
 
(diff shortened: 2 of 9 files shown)


spec/services/post_scraper_spec.rb
```diff
@@ -305,7 +305,7 @@
 
     html = "<!DOCTYPE html>\n<html></html>\n"
     stub_with_body = double
-    expect(stub_with_body).to receive(:body).and_return(html)
+    allow(stub_with_body).to receive(:body).and_return(html)
```
Throne3d, line 308: What's the difference between allow and expect here? I thought expect would require a call, but allow would just stub it out (without any expectations).

Throne3d (approved):
Beyond the comment, LGTM!

spec/services/post_scraper_spec.rb
teceler: Basically that, allow stubs and expect does an expectation. (Expect also stubs In An Order which came up in one of the tests xD) Rubocop thinks you shouldn't mix expecting something and stubbing it and honestly I think it's probably right. ...possibly some of these should *also* be expected :thinking:

Throne3d: How do we expect without a stub in context?

teceler: I think expect to receive doesn't stub but I have to check. that or you do the stubbing desperately which is, at least, clearer as to what the heck you're doing?

Throne3d (changes requested):
Spoke about this in the discord - would be good to turn the `expect(blah).to receive(thing).and_do_thing()`s into `allow(...).to receive(thing).and_to_thing()` followed by an expect on it too

spec/jobs/application_job_spec.rb
```diff
@@ -28,7 +28,7 @@ def perform(*_args)
     expect(ExceptionNotifier).to receive(:notify_exception).with(exc, data: { job: StubJob.name, args: [2, :test] })
 
     job = StubJob.new(2, :test)
-    expect(job).to receive(:perform).and_raise(exc)
+    allow(job).to receive(:perform).and_raise(exc)
```
Throne3d, line 31:
```suggestion
    allow(job).to receive(:perform).and_raise(exc)
    expect(job).to receive(:perform)
```

spec/script/post_split_spec.rb
```diff
@@ -12,7 +12,7 @@
     let(:reply) { create(:reply) }
 
     it "requires valid subject" do
-      expect(STDIN).to receive(:gets).and_return('')
+      allow(STDIN).to receive(:gets).and_return('')
```
Throne3d, line 15:
```suggestion
      allow(STDIN).to receive(:gets).and_return('')
      expect(STDIN).to receive(:gets)
```

spec/services/post_scraper_spec.rb
Throne3d, line 308:
```suggestion
    allow(stub_with_body).to receive(:body).and_return(html)
    expect(stub_with_body).to receive(:body)
```

Throne3d (commented):
I think this is all!

Throne3d (approved):
Looks good!

PR #1615, opened 2022-01-17
Author search improvements

### LINKED ISSUE #1540: Use AND rather than OR for multiauthor continuity search

teceler:
Fixes #1540, as the actual problem seems to have been cameo inclusion


app/controllers/posts_controller.rb
```diff
@@ -268,17 +268,11 @@ def search
...
-        else
-          post_ids &= author_posts
-        end
-        break if post_ids.empty?
-      end
-      @search_results = @search_results.where(id: post_ids.uniq)
+      # get author matches for posts that have at least one
+      author_posts = Post::Author.where(user_id: params[:author_ids]).group(:post_id)
+      # select posts that have all of them
+      author_posts = author_posts.having('COUNT(post_authors.user_id) = ?', params[:author_ids].length).pluck(:post_id)
```
Throne3d, line 274: Ooh, this is interesting. If we weren't so reliant on Rails for generating the SQL here, I'd probably do some condition like plucking an array of distinct user IDs and checking it's equal to the ones we want, but I'm having trouble thinking how to translate that into Rails. Since PostAuthor is probably unique enough over (post, user), this should be fine!

Throne3d: (I think we can technically `.select(:post_id)` here and use a subquery but I know we're worried about introducing too many of those =P just saying)

teceler: it should be unique on that scope, yeah, on the rails level if not the db level

app/services/board/searcher.rb
```diff
@@ -10,7 +10,10 @@ def search(params)
   end
 
   def search_authors(author_ids)
-    author_boards = author_ids.map { |author_id| BoardAuthor.where(user_id: author_id).pluck(:board_id) }.reduce(:&).uniq
+    # get author matches for boards that have at least one
+    author_boards = BoardAuthor.where(user_id: author_ids, cameo: false).group(:board_id)
```
Throne3d, line 14: (We don't want to include cameos? Huh, I kinda figured that might be something people wanted, but if that was the original issue, then it seems OK)

app/controllers/posts_controller.rb
Throne3d: Oh, we could maybe do `COUNT(DISTINCT post_authors.user_id)` to guarantee this is true even if we don't have a unique constraint over (post, user), but I think we do, so it shouldn't be an issue

Throne3d (approved):

app/services/board/searcher.rb
teceler: I'd be fine with writing a way *to* include cameos but it was a source of confusion especially since the cameo authors weren't displayed in results

PR #1616, opened 2022-01-27
Read-only account type

Marri:
How horrified will our tests and linters be with me? Tbd.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -165,6 +165,13 @@ class ApplicationController < ActionController::Base
   end
   helper_method :generate_short
 
+  def readonly_forbidden
+    return unless logged_in?
+    return unless current_user.readonly?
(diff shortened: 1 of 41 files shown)


app/controllers/application_controller.rb
```diff
@@ -165,6 +165,13 @@ def generate_short(msg)
   end
   helper_method :generate_short
 
+  def readonly_forbidden
+    return unless logged_in?
+    return unless current_user.readonly?
+    flash[:error] = "This feature is not available to read-only accounts."
+    redirect_to continuities_path
```
teceler, line 172: Hmm on one hand a better redirect might be good, on the other that sounds potentially tricky

teceler:
going over controllers quickly -- do we want reader accounts to have block access -- I think yes? indexes currently require a mod to create but what about those? they specifically shouldn't have access to messages -- news creation is already locked -- tags should be?

...also we might want a better header for readers

PR #1620, opened 2022-02-13
Update Ruby version

Marri:


teceler:
So as far as deprecation warnings we have:

ActiveSupport (6.0.4.4):
lib/active_support/cache.rb:476: warning: Passing the keyword argument as the last hash parameter is deprecated
lib/active_support/cache.rb:487: warning: Passing the keyword argument as the last hash parameter is deprecated
lib/active_support/cache.rb:739: warning: The called method `initialize' is defined here
this file doesn't seem to... actually... exist... in activesupport... so I suspect this is in fact a redis-activesupport problem

ActionPack (6.0.4.4):
lib/action_controller/test_case.rb:460: warning: The called method `process' is defined here

Apipie-Rails (0.5.18) (fix seems to be merged but unreleased):
lib/apipie/extractor/recorder.rb:154: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call

Redis (4.2.5):
lib/redis.rb:836: warning: The called method `set' is defined here

Redis-Namespace (1.8.0):
lib/redis/namespace.rb:476: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call

Redis-ActiveSupport (5.2.0) (unmaintained):
lib/active_support/cache/redis_store.rb:85: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
lib/active_support/cache/redis_store.rb:286: warning: The called method `read_entry' is defined here
lib/active_support/cache/redis_store.rb:305: warning: The called method `delete_entry' is defined here

Throne3d:
Currently depending on: #1621 (can be removed), #1622 (important for removing deprecation warnings)

Throne3d (approved):
👀

PR #1622, opened 2022-02-13
Switch to native rails redis support

teceler:
And remove dependency on unmaintained redis-activesupport gem

Will require a new redis server for caching because this does not well support storing caches and lock keys on the same database.


Throne3d:
We'll keep using REDIS_URL for the non-cache version, e.g. for background jobs, but then REDIS_CACHE_URL specifically for cache items?

teceler:
> We'll keep using REDIS_URL for the non-cache version, e.g. for background jobs, but then REDIS_CACHE_URL specifically for cache items?

Yeah, I left the uses of REDIS_URL that seemed to be pointing at that (the ones for `$redis`) alone.

Throne3d (approved):
Sounds good to me then! We just need to verify we add the new cache on deploy

config/environments/development.rb
```diff
@@ -18,7 +18,7 @@
     config.action_controller.perform_caching = true
     # config.action_controller.enable_fragment_cache_logging = true
 
-    config.cache_store = :redis_cache_store, {url: ENV['REDIS_URL']}
+    config.cache_store = :redis_cache_store, url: ENV['REDIS_CACHE_URL']
```
Marri, line 21: Probably we need to set this somewhere for development? Docker-compose or something?

Throne3d: Ooh, yeah – we have

```yaml
REDIS_URL: redis://redis
```
in there

teceler: Yeah, dev doesn't use caching *by default* but we probably still want something it works with for testing caching, and similar.

docker-compose.yml
```diff
@@ -2,6 +2,7 @@ version: "3.7"
 
 x-env: &env
   REDIS_URL: redis://redis
+  REDIS_CACHE_URL: redis://redis
```
teceler, line 5: That puts them both on the same redis instance, but given how rarely dev is expected to use this that's probably fine, and can be tweaked per case if it's not.

PR #1624, opened 2022-02-16
Fix development redis URL

Throne3d:


teceler (approved):
Whoops! Looks good as a fix!

PR #1627, opened 2022-02-23
Add codeql-analysis.yml, address lints

Throne3d:
Auto-generated by GitHub!

### DIFF
--- a/app/assets/javascripts/posts/editor.js
+++ b/app/assets/javascripts/posts/editor.js
@@ -152,7 +152,7 @@ function setupWritableEditor() {
     // Set the ID
     var id = $(this).val();
     $("#reply_character_alias_id").val(id);
-    $("#post-editor .post-character #name").html($('#character_alias option:selected').text());
+    $("#post-editor .post-character #name").text($('#character_alias option:selected').text());
     $('#alias-selector').hide();
     $("#post-editor .post-character").data('alias-id', id);
   });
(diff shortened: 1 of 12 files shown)


Marri (approved):
Approving on the assumption we'll fix the new warnings separately

app/controllers/application_controller.rb
```diff
@@ -206,8 +206,8 @@ def set_timezone(&block)
   def require_glowfic_domain
     return unless Rails.env.production? || params[:force_domain] # for testability
     return unless standard_request?
-    return if request.host.include?('glowfic.com')
-    return if request.host.include?('glowfic-staging.herokuapp.com')
+    return if request.host.match?(/(^|\.)glowfic\.com$/)
+    return if request.host.match?(/(^|\.)glowfic-staging\.herokuapp\.com$/)
```
Marri, line 210: These should be optional, since `request.host` is not user generated/provided

app/assets/javascripts/posts/editor.js
```diff
@@ -253,29 +253,29 @@ function bindIcon() {
...
   var imgId = icon.id;
   var imgUrl = icon.url;
   var imgKey = icon.keyword;
   shownIcons.push(icon.id);
 
-  if (!icon.skip_dropdown) $("#icon_dropdown").append($("<option>").attr({value: imgId}).append(imgKey));
+  if (!icon.skip_dropdown) $("#icon_dropdown").append($("<option>").attr({value: imgId}).text(imgKey));
   var iconImg = $("<img>").attr({src: imgUrl, alt: imgKey, title: imgKey, 'class': 'icon img-'+imgId, 'data-icon-id': imgId});
-  return $("<div>").attr('class', 'gallery-icon').append(iconImg).append("<br />").append(imgKey)[0].outerHTML;
+  return $("<div>").attr('class', 'gallery-icon').append(iconImg).append("<br />").append(document.createTextNode(imgKey));
```
Marri, line 278: Do not totally understand what's up with the text node thing here

app/assets/javascripts/posts/editor.js
```diff
@@ -253,29 +253,29 @@ function bindIcon() {
...
+    iconNodes.push(iconNode(icons[i]));
   }
 
-  if (!multiGallery) return iconsString;
+  if (!multiGallery) return iconNodes;
 
-  var nameString = "<div class='gallery-name'>" + gallery.name + "</div>";
-  return "<div class='gallery-group'>" + nameString + iconsString + "</div>";
+  var nameNode = $("<div class='gallery-name'></div>").text(gallery.name);
+  return $("<div class='gallery-group'>").append(nameNode).append(iconNodes);
```
Marri, line 267: do we need a </div> here?

app/controllers/application_controller.rb
Throne3d: Cool! Which would you prefer, keeping this change or undoing it and marking false positive?

app/assets/javascripts/posts/editor.js
Throne3d: We can't use `.text`, because that overrides all of the contents - we just want to add text to it. `.text` basically wraps the text in a DOM text node, which makes sure it isn't interpreted as HTML, but instead as text - this should let us append text safely here

app/assets/javascripts/posts/editor.js
Throne3d: Shouldn't need to! I'll add it anyway but I think jQuery does that automatically

Throne3d:
```suggestion
  return $("<div class='gallery-group'></div>").append(nameNode).append(iconNodes);
```

Marri: Just wasn't sure because the line above had one; don't really care which way we make it consistent but they should probably match

PR #1628, opened 2022-02-25
Add background-color back to login warning

Throne3d:
When we removed the inbox flash, we also incidentally
removed the background from the login warning, where you
log in/out in another tab!

This adds it back, piggybacking off the breadcrumbs color.
We could instead use `success`, or `error`, or create a new
separate category for this - I went with a small change
here that hopefully wouldn't use a confusing color, but am
open to thoughts!

Here's how it looks on starrylight before the change:

image

And here's how it looks afterwards:

image


teceler (approved):
Looks good, if not the syntax I'm used to seeing for `.flash.error`s

PR #1629, opened 2022-02-25
Increase colspan to tags > characters header

Throne3d:
Seems to be consistently colspan=8 with the space for the
edit buttons! We're just down by one.

See e.g. https://glowfic.com/tags/322?view=characters

image


teceler (approved):
:green_book:

PR #1631, opened 2022-02-25
Address Bullet complaints

Throne3d:
Separated from #1306

### DIFF
--- a/app/controllers/characters_controller.rb
+++ b/app/controllers/characters_controller.rb
@@ -65,6 +65,7 @@ class CharactersController < ApplicationController
 
   def show
     @page_title = @character.name
+    @posts = posts_from_relation(@character.recent_posts) if params[:view] == 'posts'
     @meta_og = og_data
     use_javascript('characters/show') if @character.user_id == current_user.try(:id)
   end
--- a/app/controllers/characters_controller.rb
+++ b/app/controllers/characters_controller.rb
@@ -65,7 +65,6 @@ class CharactersController < ApplicationController
 
   def show
     @page_title = @character.name
-    @posts = posts_from_relation(@character.recent_posts)
     @meta_og = og_data
     use_javascript('characters/show') if @character.user_id == current_user.try(:id)
   end
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -6,16 +6,22 @@ class MessagesController < ApplicationController
 
   def index
     blocked_ids = Block.where(blocking_user: current_user).pluck(:blocked_user_id)
+
+    includes = []
     if params[:view] == 'outbox'
       @page_title = 'Outbox'
       from_table = current_user.sent_messages.where(visible_outbox: true).ordered_by_thread.select('distinct on (thread_id) messages.*')
       from_table = from_table.where.not(recipient_id: blocked_ids).joins(:recipient).where.not(users: { deleted: true })
+      includes << :recipient
     else
       @page_title = 'Inbox'
       from_table = current_user.messages.where(visible_inbox: true).ordered_by_thread.select('distinct on (thread_id) messages.*')
       from_table = from_table.where.not(sender_id: blocked_ids).left_outer_joins(:sender).where('users.deleted IS NULL OR users.deleted = false')
+      includes << :sender
     end
-    @messages = Message.from(from_table).select('*').order('subquery.id desc').paginate(page: page)
+    @messages = Message.from(from_table, "messages").joins(:first_thread)
+      .select('*', 'first_threads_messages.subject as thread_subject')
+      .includes(*includes).order('messages.id desc').paginate(page: page)
     @view = @page_title.downcase
   end
 
@@ -74,7 +80,7 @@ class MessagesController < ApplicationController
     @page_title = message.unempty_subject
     @box = message.box(current_user)
 
-    @messages = Message.where(thread_id: message.thread_id).ordered_by_id
+    @messages = Message.where(thread_id: message.thread_id).includes(:sender).ordered_by_id
     if @messages.any? { |m| m.recipient_id == current_user.id && m.unread? }
       @messages.each do |m|
         next unless m.unread?
--- a/app/helpers/post_helper.rb
+++ b/app/helpers/post_helper.rb
@@ -13,7 +13,9 @@ module PostHelper
       return safe_join([joined_links, "#{num_deleted} #{deleted}"], ' and ')
     end
 
-    first_author = post.user.deleted? ? authors.first : post.user
+    # filter post.user from post.authors to avoid n+1 query over post.user
+    first_author = authors.find { |x| x.id == post.user_id } || authors.first
+
     first_link = linked ? user_link(first_author, colored: colored) : first_author.username
     hovertext = safe_join((authors - [first_author]).map(&:username), ', ')
     others = linked ? link_to("#{total - 1} others", stats_post_path(post), title: hovertext) : "#{total - 1} others"
--- a/app/views/characters/show.haml
+++ b/app/views/characters/show.haml
@@ -78,7 +78,7 @@
             Delete Character
 
 - if params[:view] == 'galleries'
-  - galleries = @character.galleries.ordered
+  - galleries = @character.galleries.select('galleries.*').with_gallery_groups.ordered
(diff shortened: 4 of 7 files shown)


app/controllers/characters_controller.rb
```diff
@@ -61,7 +61,6 @@ def create
 
   def show
     @page_title = @character.name
-    @posts = posts_from_relation(@character.recent_posts)
```
Throne3d, line 64: Replaced by fetching this relation only when we need it in the relevant view

app/controllers/messages_controller.rb
```diff
@@ -5,16 +5,22 @@ class MessagesController < ApplicationController
...
     else
       @page_title = 'Inbox'
       from_table = current_user.messages.where(visible_inbox: true).ordered_by_thread.select('distinct on (thread_id) messages.*')
       from_table = from_table.where.not(sender_id: blocked_ids).left_outer_joins(:sender).where('users.deleted IS NULL OR users.deleted = false')
+      includes << :sender
     end
-    @messages = Message.from(from_table).select('*').order('subquery.id desc').paginate(page: page)
+    @messages = Message.from(from_table, "messages").joins(:first_thread)
+      .select('*', 'first_threads_messages.subject as thread_subject')
+      .includes(*includes).order('messages.id desc').paginate(page: page)
```
Throne3d, line 23: The includes here are for displaying the other participant in messages on message list views, which is currently a n+1

app/controllers/messages_controller.rb
```diff
@@ -5,16 +5,22 @@ class MessagesController < ApplicationController
...
+      includes << :recipient
     else
       @page_title = 'Inbox'
       from_table = current_user.messages.where(visible_inbox: true).ordered_by_thread.select('distinct on (thread_id) messages.*')
       from_table = from_table.where.not(sender_id: blocked_ids).left_outer_joins(:sender).where('users.deleted IS NULL OR users.deleted = false')
+      includes << :sender
     end
-    @messages = Message.from(from_table).select('*').order('subquery.id desc').paginate(page: page)
+    @messages = Message.from(from_table, "messages").joins(:first_thread)
+      .select('*', 'first_threads_messages.subject as thread_subject')
```
Throne3d, line 22: thread_subject here lets us find the first thread title for display of the thread, which right now involves a n+1 query for each message - used for unempty_subject

app/controllers/characters_controller.rb
teceler, line 64: Is this not used in the html?

app/controllers/messages_controller.rb
```diff
@@ -72,7 +78,7 @@ def show
     @page_title = message.unempty_subject
     @box = message.box(current_user)
 
-    @messages = Message.where(thread_id: message.thread_id).ordered_by_id
+    @messages = Message.where(thread_id: message.thread_id).includes(:sender).ordered_by_id
```
Throne3d, line 81: Each message in a thread on messages#show loads its user separately right now

app/helpers/post_helper.rb
```diff
@@ -13,7 +13,9 @@ def author_links(post, linked: true, colored: false)
       return safe_join([joined_links, "#{num_deleted} #{deleted}"], ' and ')
     end
 
-    first_author = post.user.deleted? ? authors.first : post.user
+    # filter post.user from post.authors to avoid n+1 query over post.user
+    first_author = authors.find { |x| x.id == post.user_id } || authors.first
```
Throne3d, line 17: Any posts with "<user> and 3 others" currently use an n+1 to select the first post author

app/views/characters/show.haml
```diff
@@ -78,7 +78,7 @@
             Delete Character
 
 - if params[:view] == 'galleries'
-  - galleries = @character.galleries.ordered
+  - galleries = @character.galleries.select('galleries.*').with_gallery_groups.ordered
```
Throne3d, line 81: Character galleries use an n+1 to fetch gallery groups

app/views/posts/stats.haml
```diff
@@ -40,7 +40,7 @@
     %tr
       %th.sub.width-150 Authors
       %td{class: cycle('odd', 'even')}
-        - post_authors = @post.post_authors.joins(:user).order(Arel.sql('LOWER(users.username) asc'))
+        - post_authors = @post.post_authors.joins(:user).includes(:user).order(Arel.sql('LOWER(users.username) asc'))
```
Throne3d, line 43: Post stats page uses an n+1 to figure out how to render users - joins doesn't do an includes for purposes of the associations, so we still do an n+1 fetch afterwards

app/controllers/characters_controller.rb
Throne3d: (Yes, but only in some cases, so I embedded it directly)

teceler: ah, I didn't get to that yet

app/controllers/messages_controller.rb
```diff
@@ -5,16 +5,22 @@ class MessagesController < ApplicationController
...
     if params[:view] == 'outbox'
       @page_title = 'Outbox'
       from_table = current_user.sent_messages.where(visible_outbox: true).ordered_by_thread.select('distinct on (thread_id) messages.*')
       from_table = from_table.where.not(recipient_id: blocked_ids).joins(:recipient).where.not(users: { deleted: true })
+      includes << :recipient
     else
       @page_title = 'Inbox'
       from_table = current_user.messages.where(visible_inbox: true).ordered_by_thread.select('distinct on (thread_id) messages.*')
       from_table = from_table.where.not(sender_id: blocked_ids).left_outer_joins(:sender).where('users.deleted IS NULL OR users.deleted = false')
+      includes << :sender
```
teceler, line 19: not sure why you're doing this as array appending rather than a ternary after this bit, but w/e, that's not important

app/controllers/messages_controller.rb
teceler: :deer:

app/controllers/characters_controller.rb
Throne3d: Test expectations wanted it to be assigned anyway, so I've moved it back with an if!

app/helpers/post_helper.rb
teceler: you lost the deletion check there
```suggestion
    first_author = authors.active.find { |x| x.id == post.user_id } || authors.first
```

app/controllers/messages_controller.rb
Throne3d: Oh, I was originally using `:first_thread` - I would've also just done a variable here like `include` but that's a keyword so I didn't fix it up.

I could use `inc = :recipient` and `inc = :sender`

bin-docker/rspec
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-docker-compose run --rm web bundle exec rspec "$@"
+docker-compose run --rm web /bin/bash -c "bundle exec rspec \"$@\"; export tempcode=\$? && rm -Rf tmp/ log/* && exit \$tempcode"
```
teceler, line 3: What is this doing?

app/helpers/post_helper.rb
Throne3d: `authors` is defined above and already rejects deleted, actually!

```
authors = post.authors.reject(&:deleted?).sort_by { |a| a.username.downcase }
```

teceler: oh, excellent, carry on then. -- possibly we should check if `post.user_deleted` exists before we do the more computationally expensive thing but it's not a big deal

bin-docker/rspec
Throne3d: I do not know how this got committed e.e but it's the hack I'm doing locally because Docker keeps writing tmp/ or log/ files with root permissions.

Let me delete it, whoops.

app/controllers/messages_controller.rb
Marri: why'd we change the order statement here?

Throne3d: It should be doing the same thing - we just renamed the name Rails will give to the subquery within the `.from` call. I think it was necessary so it knew which field to join on when I added `.joins(:first_thread)`.

Marri (approved):

PR #1634, opened 2022-02-27
Disable Rails/SaveBang cop on app/ and config/

teceler:
Since we seem to be ignoring the warnings anyway.

Leaving it on on spec which we want to fail noisily.

I'm open to instead putting this in a rubocop-todo file if we want to actually check return values of stuff, but.


Throne3d (approved):

teceler:
Superseded by #1639

PR #1638, opened 2022-03-17
Add admin service to transfer characters & related between accounts

### LINKED ISSUE #1496: Transfer character ownership admin tool
Done this once, only managed it kind of, have gotten another request, can we make a properly tested thing somewhere to magic this doable

teceler:
For eventual integration into the admin control panel, but should work from the command line in the interim.

Fixes #1496

### DIFF
--- a/app/services/character/relocator.rb
+++ b/app/services/character/relocator.rb
@@ -1,0 +1,120 @@
+# Steps:
+# validate single user, unshared icons, galleries, and templates (if included), lack of character groups
+# collect: templates (if included), characters, posts, replies, drafts, galleries, icons
+# update all on unaudited (templates, aliases, drafts, galleries, icons)
+# UpdateModelJob on audited (characters, posts, replies)
+
+class Character::Relocator < Object
+  def initialize(admin_id)
+    @audited_user_id = admin_id
+  end
+
+  def transfer(character_ids, new_user_id, include_templates: false)
+    @characters = Character.where(id: character_ids)
+    raise CharacterGroupError.new('Characters must not have groups') if @characters.where.not(character_group: nil).exists?
+
+    User.find(new_user_id) # fail with invalid target user
+
+    @user_id = validate_user!
+    @user_other_characters = Character.where(user_id: @user_id).where.not(id: @characters.ids)
+    @other_character_galleries = get_galleries(@user_other_characters)
+    @other_character_icon_ids = get_icon_ids(@user_other_characters, @other_character_galleries)
+
+    @galleries = validate_galleries!
+    @icons = validate_icons!
+    @templates = validate_templates! if include_templates
+
+    @posts = Post.where(character_id: character_ids, user_id: @user_id)
+    @replies = Reply.where(character_id: character_ids, user_id: @user_id)
+    @drafts = ReplyDraft.where(character_id: character_ids, user_id: @user_id)
+
+    @new_authors = calc_new_authors(new_user_id)
+    @rem_authors = calc_rem_authors
+
+    update_characters(new_user_id, include_templates)
+  end
+
+  private
+
+  def validate_user!
+    user_id = @characters.select(:user_id).distinct.pluck(:user_id)
+    raise RequireSingleUser.new('Characters must all have the same original user') if user_id.length > 1
+    User.find(user_id.first).id
+  end
+
+  def validate_galleries!
+    galleries = get_galleries(@characters)
+    intersection = @other_character_galleries.ids & galleries.ids
+    msg = "Galleries for characters which are being moved must not be the same as ones not being moved; gallery ids #{intersection} intersect"
+    raise OverlappingGalleriesError.new(msg) unless intersection.empty?
+    galleries
+  end
+
+  def validate_icons!
+    icon_ids = get_icon_ids(@characters, @galleries)
+    intersection = @other_character_icon_ids & icon_ids
+    msg = "Icons for characters which are being moved must not be the same as ones not being moved; icon ids #{intersection} intersect"
+    raise OverlappingIconsError.new(msg) unless intersection.empty?
+    Icon.where(id: icon_ids)
+  end
+
+  def validate_templates!
+    template_ids = @characters.select(:template_id).distinct.pluck(:template_id)
+    templates = Template.where(id: template_ids)
+    raise ActiveRecord::RecordInvalid if templates.where.not(user_id: @user_id).exists?
+    msg = 'When moving templates all characters in them must be moved'
+    raise OverlappingTemplatesError.new(msg) if @user_other_characters.where(template_id: template_ids).exists?
+    templates
+  end
+
+  def calc_new_authors(new_user_id)
+    post_ids = (@replies.select(:post_id).distinct.pluck(:post_id) + @posts.ids).uniq
+    existing_authors = Post::Author.where(post_id: post_ids, user_id: new_user_id).pluck(:post_id)
+    (post_ids - existing_authors)
+  end
+
+  def calc_rem_authors
+    post_ids = (@replies.select(:post_id).distinct.pluck(:post_id) + @posts.ids).uniq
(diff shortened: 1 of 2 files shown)


app/services/character/relocater.rb
```diff
@@ -0,0 +1,97 @@
...
+    # rubocop:disable Rails/SkipsModelValidations
+    @templates.update_all(user_id: new_user_id) if include_templates
+    @galleries.update_all(user_id: new_user_id)
+    @icons.update_all(user_id: new_user_id)
+    @drafts.update_all(user_id: new_user_id)
+    # rubocop:enable Rails/SkipsModelValidations
+
+    UpdateModelJob.perform_later(Post.to_s, { user_id: @user_id }, { user_id: new_user_id }, @audited_user_id)
+    UpdateModelJob.perform_later(Reply.to_s, { user_id: @user_id }, { user_id: new_user_id }, @audited_user_id)
+    UpdateModelJob.perform_later(Character.to_s, { user_id: @user_id }, { user_id: new_user_id }, @audited_user_id)
```
Throne3d, line 39: Should we handle how this could impact PostAuthor too? e.g. if a character is transferred in a thread that didn't previously have this new author

app/services/character/relocater.rb
```diff
@@ -0,0 +1,97 @@
+# Steps:
+# validate single user, unshared icons, galleries, and templates (if included), lack of character groups
+# collect: templates (if included), characters, posts, replies, drafts, galleries, icons
+# update all on unaudited (templates, aliases, drafts, galleries, icons)
+# UpdateModelJob on audited (characters, posts, replies)
+
+class Character::Relocater < Object
```
Throne3d, line 7: (Relocator)

spec/services/character/relocater_spec.rb
```diff
@@ -0,0 +1,133 @@
+RSpec.describe Character::Relocater do
+  include ActiveJob::TestHelper
+
+  let(:relocator) { Character::Relocater.new(create(:admin_user)) }
```
Throne3d, line 4: It looks like the relocator is meant to take an ID, not the user object - does this work right?

teceler: That is weird! why are they passing!

teceler: ...apparently `Model.find_by(id: instance_of_that_model)` Works By Rails Magic

app/services/character/relocater.rb
teceler: Fine.

app/services/character/relocator.rb
```diff
@@ -0,0 +1,115 @@
...
+    @new_authors.each { |post_id| Post::Author.create!(post_id: post_id, user_id: new_user_id) }
+    @rem_authors.each { |post_id| Post::Author.find_by!(post_id: post_id, user_id: @user_id).destroy! }
+  end
+
+  private
+
+  def validate_user!
+    user_id = @characters.select(:user_id).distinct.pluck(:user_id)
+    raise RequireSingleUser.new('Characters must all have the same original user') if user_id.length > 1
+    user_id = user_id.first
```
Marri, line 53:
```suggestion
    User.find(user_id.first).id
```

app/services/character/relocator.rb
```diff
@@ -0,0 +1,115 @@
...
+    User.find(user_id)
+    user_id
+  end
+
+  def validate_galleries!
+    galleries = get_galleries(@characters)
+    intersection = @other_character_galleries.ids & galleries.ids
+    msg = "Galleries for characters which are being moved must not be the same as ones not being moved; gallery ids #{intersection} are"
```
Marri, line 61: are what?

app/services/character/relocator.rb
```diff
@@ -0,0 +1,115 @@
...
+    raise OverlappingGalleriesError.new(msg) unless intersection.empty?
+    galleries
+  end
+
+  def validate_icons!
+    icon_ids = get_icon_ids(@characters, @galleries)
+    intersection = @other_character_icon_ids & icon_ids
+    msg = "Icons for characters which are being moved must not be the same as ones not being moved; icon ids #{intersection} are"
```
Marri, line 69: ditto

app/services/character/relocator.rb
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  def transfer(character_ids, new_user_id, include_templates: false)
+    @characters = Character.where(id: character_ids)
+    @user_id = validate_user!
+    @user_other_characters = Character.where(user_id: @user_id).where.not(id: @characters.ids)
+    @other_character_galleries = get_galleries(@user_other_characters)
+    @other_character_icon_ids = get_icon_ids(@user_other_characters, @other_character_galleries)
+
+    raise CharacterGroupError.new('Characters must not have groups') if @characters.where.not(character_group: nil).exists?
```
Marri, line 19: maybe do this earlier to fail quickly?

app/services/character/relocator.rb
```diff
@@ -0,0 +1,115 @@
...
+    @galleries = validate_galleries!
+    @icons = validate_icons!
+    @templates = validate_templates! if include_templates
+
+    @posts = Post.where(character_id: character_ids, user_id: @user_id)
+    @replies = Reply.where(character_id: character_ids, user_id: @user_id)
+    @drafts = ReplyDraft.where(character_id: character_ids, user_id: @user_id)
+
+    User.find(new_user_id)
```
Marri, line 28: ?

app/services/character/relocator.rb
```diff
@@ -0,0 +1,115 @@
...
+    @icons.update_all(user_id: new_user_id)
+    @drafts.update_all(user_id: new_user_id)
+    # rubocop:enable Rails/SkipsModelValidations
+
+    UpdateModelJob.perform_later(Post.to_s, { character_id: @characters.ids }, { user_id: new_user_id }, @audited_user_id)
+    UpdateModelJob.perform_later(Reply.to_s, { character_id: @characters.ids }, { user_id: new_user_id }, @audited_user_id)
+    UpdateModelJob.perform_later(Character.to_s, { id: @characters.ids }, { user_id: new_user_id }, @audited_user_id)
+
+    @new_authors.each { |post_id| Post::Author.create!(post_id: post_id, user_id: new_user_id) }
+    @rem_authors.each { |post_id| Post::Author.find_by!(post_id: post_id, user_id: @user_id).destroy! }
```
Marri, line 45: do we need ~all of this in a db transaction

app/services/character/relocator.rb
teceler: no, that doesn't work b/c then you -- well, you'd have to put the first on the return, too, I guess?

app/services/character/relocator.rb
teceler: Making sure the user actually exists, iirc

app/services/character/relocator.rb
teceler: ...yeah this should probably be a transcation

app/services/character/relocator.rb
teceler: ...the same?

app/services/character/relocator.rb
Marri: This should be able to replace this and the following couple lines

app/services/character/relocator.rb
Marri: Yeah but we do this in validate_user! as well

app/services/character/relocator.rb
teceler: -- oh, because you're asking for the id off it -- yeah, that could work

app/services/character/relocator.rb
teceler: I think that's the old user though?

teceler: yeah validate_user! is checking the characters all have one user who exists, this is checking that the user_id the service was given corresponds to an actual user

Marri (approved):

PR #1641, opened 2022-03-28
Add legacy privacy setting

teceler:


app/concerns/concealable.rb
```diff
@@ -7,13 +7,16 @@ module Concealable
...
       registered: 3,
+      legacy: 4,
     }, _prefix: true
 
     def visible_to?(user)
       # does not support access lists at this time
       return true if privacy_public?
       return false unless user
       return true if privacy_registered?
+      return false if user.read_only?
```
Marri, line 18: Do we ever now, or plan to, use visible_to? on something that read-only users could own? (e.g. if we someday let them make indexes or something, iunno). read only cannot presently coexist with admin so that part at least seems fine.

app/jobs/notify_followers_of_new_post_job.rb
```diff
@@ -49,7 +49,9 @@ def filter_users(post, user_ids)
     user_ids -= post.author_ids
     user_ids -= blocked_user_ids(post)
     return [] unless user_ids.present?
-    User.where(id: user_ids, favorite_notifications: true)
+    users = User.where(id: user_ids, favorite_notifications: true)
+    users = users.where.not(role_id: Permissible::READONLY) if post.privacy_legacy?
```
Marri, line 53: I think we have a scope for this

app/models/post/view.rb
```diff
@@ -13,7 +13,7 @@ def mark_favorite_read
     favorited_users = user.favorites.where(favorite: post.joined_authors).exists?
     return unless favorited_continuity || favorited_users
 
-    message = NotifyFollowersOfNewPostJob.notification_about(post, user, unread_only: true)
+    message = NotifyFollowersOfNewPostJob.notification_about(post, user, unreadonly: true)
```
Marri, line 16: Is this on purpose?

Marri (commented):
Still need to review the tests but have some comments for now

app/concerns/concealable.rb
teceler: I could conceivably see letting them make indexes but I don't expect it? I think you'd just have to put the owner condition earlier if we did that, though.

app/models/post/view.rb
teceler: ...I thought I fixed that. no, it's not on purpose.

app/concerns/concealable.rb
```diff
@@ -6,14 +6,17 @@ module Concealable
       public: 0,
       private: 1,
       access_list: 2,
-      registered: 3,
+      full_accounts: 3,
+      registered: 4,
```
teceler, line 10: I think possibly we want some comments here

Marri (approved):

PR #1647, opened 2022-05-05
Update flatposts on more changes

### LINKED ISSUE #252: Editing icons/characters does not update flat views

### LINKED ISSUE #1460: Deleting replies does not update flat posts
Because after_save is stupid and doesn't cover deletes

### LINKED ISSUE #1645: More things should queue GenerateFlatPostJob
* Reply deletion
* Character replace affecting the post
* Icon replace affecting the post
* Icon swap (replacing the image) on posts where that icon is used

teceler:
Fixes #1645 
Fixes #1460 
Fixes #252

* Regenerates flatpost when reply is destroyed
* Regenerates flatpost when icon or icon keyword is changed
* Regenerates flatpost when character name or screenname is changed
* Regenerates flatpost when username is changed or user is archived

### DIFF
--- a/app/models/icon.rb
+++ b/app/models/icon.rb
@@ -22,6 +22,7 @@ class Icon < ApplicationRecord
   before_validation :use_icon_host
   before_save :use_https
   before_update :delete_from_s3
+  after_update :update_flat_posts
   after_destroy :clear_icon_ids, :delete_from_s3
 
   scope :ordered, -> { order(Arel.sql('lower(keyword) asc'), created_at: :asc, id: :asc) }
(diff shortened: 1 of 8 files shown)


app/models/icon.rb
```diff
@@ -73,6 +74,12 @@ def clear_icon_ids
     UpdateModelJob.perform_later(Reply.to_s, { icon_id: id }, { icon_id: nil }, audited_user_id)
   end
 
+  def update_flat_posts
+    return unless saved_change_to_url? || saved_change_to_keyword
```
teceler, line 78: Pretty sure that should have a question mark, which raises the question of why tests passed...

Marri: I think Rails magic things often have functions with and without the keyword (possibly if they're using a variable with a getter that returns the same value?) but support changing it to use the ?

app/models/user.rb
```diff
@@ -159,4 +160,10 @@ def blocked_or_hidden_posts(keyword, post_user_ids, full_user_ids)
       post_ids
     end
   end
+
+  def update_flat_posts
+    return unless saved_change_to_username || saved_change_to_deleted
```
Marri, line 165: ?s for consistency perhaps

app/models/user.rb
```diff
@@ -159,4 +160,10 @@ def blocked_or_hidden_posts(keyword, post_user_ids, full_user_ids)
...
+    post_ids = (Post.where(user_id: id).pluck(:id) + Reply.where(user_id: id).select(:post_id).distinct.pluck(:post_id)).uniq
```
Marri, line 166: Possibly faster to use PostAuthor here

teceler: -- oh, yeah, PostAuthor probably would cover that

Marri (approved):

PR #1649, opened 2022-05-05
Make all extends use placeholders

### LINKED ISSUE #1633: Use placeholders for extends (scss)
See https://github.com/stylelint-scss/stylelint-scss/blob/master/src/rules/at-extend-no-missing-placeholder/README.md, using placeholders results in less css.

teceler:
Fixes #1633, probably needs tested for styling consistency

### DIFF
--- a/app/assets/stylesheets/galleries.scss
+++ b/app/assets/stylesheets/galleries.scss
@@ -33,7 +33,7 @@
   align-items: stretch;
   input { height: 23px; }
 
-  .icons-box {
+  .icons-box, .replace-icon {
     display: flex;
     align-items: center;
     justify-content: center;
@@ -126,7 +126,7 @@
 
 /* The preview icon boxes on replace icon and replace character */
 .replace-icon {
-  @extend .icons-box;
+  background-color: $bg_color_icon;
   width: 150px;
   text-align: center;
 }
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/global.scss
```diff
@@ -84,8 +84,10 @@ button, .button {
   }
 }
 
+.spacer { @extend %shared-spacer; }
+
 .spacer-alt {
-  @extend .spacer;
+  @extend %spacer;
```
Marri, line 90: Should this be `%shared-spacer` not `%spacer`?

app/assets/stylesheets/galleries.scss
```diff
@@ -126,7 +126,7 @@
 
 /* The preview icon boxes on replace icon and replace character */
 .replace-icon {
-  @extend .icons-box;
+  background-color: $bg_color_icon;
```
Marri, line 129: Was this on purpose? Doesn't seem related to the @extend at least.

app/assets/stylesheets/global.scss
teceler: yes, which is probably why tests are crashing

app/assets/stylesheets/galleries.scss
teceler: I'd have to hunt for the icons-box definition but I suspect I just copied that from it

app/assets/stylesheets/global.scss
teceler:
```suggestion
  @extend %shared-spacer;
```

PR #1659, opened 2022-06-23
Preserve reply.created_at on restore

### LINKED ISSUE #1657: Restoring deleted reply messes with unread marker
Ezra — Today at 9:16 PM
I'm getting the behavior where a Read marker is staying at a specific edited reply, even when specifically clicking the eye on another reply. (Clicking the eye gives the “Post has been marked as read until...” message for the reply I clicked, so it is maybe prematurely announcing its success.)
I'm gonna try the workaround of clearing my read marker for the post and seeing if I can set it after that.
I can clear it, and I can set it before that point, but it still gets stuck. I'm assuming that the issue will no longer affect me after I “catch up” to the time of the edited reply, so absent a fix I can just try to get that far before losing my place.
maybe that's Feedback
—the offending reply is this (run of) edited reply(ies) in “crisis of faith”
https://glowfic.com/replies/1814219#reply-1814219

Manya — Today at 9:47 PM
oh yeah, my read marker went crazy for a bit there
and yeah, it gets fixed when you get up to the tags that were posted after that one

Marri — Today at 9:51 PM
hmmmm I vaguely recall reordering some replies once, I wonder if that's where that happened? unread markers and reordering do not yet exactly work together which is why it's still manual

Manya — Today at 9:53 PM
it's a recent one, I think it was deleted and then undeleted?
https://glowfic.com/replies/1814219/history
it looks like undeleting set the _post_ date to the date it was restored, and not just the updated date*

*n.b. this means reply.created_at not post.tagged_at

Marri:
Fixes #1657 (I hope)

### DIFF
--- a/app/controllers/replies_controller.rb
+++ b/app/controllers/replies_controller.rb
@@ -235,6 +235,7 @@ class RepliesController < WritableController
     end
 
     new_reply = Reply.new(audit.audited_changes)
+    new_reply.created_at = audit.audited_changes[:created_at]
     unless new_reply.editable_by?(current_user)
       flash[:error] = "You do not have permission to modify this post."
       redirect_to post_path(new_reply.post) and return
--- a/app/controllers/replies_controller.rb
+++ b/app/controllers/replies_controller.rb
@@ -235,7 +235,7 @@ class RepliesController < WritableController
     end
 
     new_reply = Reply.new(audit.audited_changes)
-    new_reply.created_at = audit.audited_changes[:created_at]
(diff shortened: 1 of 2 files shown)


teceler (approved):
Assuming tests pass, looks good!

app/controllers/replies_controller.rb
```diff
@@ -235,6 +235,7 @@ def restore
     end
 
     new_reply = Reply.new(audit.audited_changes)
+    new_reply.created_at = Audited::Audit.find_by(action: 'create', auditable_id: params[:id]).created_at
```
Marri, line 238: Is it possible we'd have more than one of these (if e.g. we'd restored it before?)

PR #1662, opened 2022-06-24
Clean up some minor lints

teceler:
* Use Enumerable#index_with rather than map + hash
* Use use_zone block for setting time zone, more firmly this time
* Use filter_map rather than map + compact (note: this will filter out things that return 'false' as well as 'nil'
* Add spaces to a config file

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -103,13 +103,13 @@ class PostsController < WritableController
 
   def unhide
     if params[:unhide_boards].present?
-      board_ids = params[:unhide_boards].map(&:to_i).compact.uniq
+      board_ids = params[:unhide_boards].filter_map(&:to_i).uniq
       views_to_update = BoardView.where(user_id: current_user.id).where(board_id: board_ids)
       views_to_update.each { |view| view.update(ignored: false) }
     end
 
     if params[:unhide_posts].present?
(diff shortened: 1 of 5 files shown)


app/controllers/posts_controller.rb
```diff
@@ -103,13 +103,13 @@ def hidden
 
   def unhide
     if params[:unhide_boards].present?
-      board_ids = params[:unhide_boards].map(&:to_i).compact.uniq
+      board_ids = params[:unhide_boards].filter_map(&:to_i).uniq
```
Throne3d, line 106: Interesting - I think `.map(&:to_i).compact` wasn't doing anything beyond `.map(&:to_i)` before – see this example, where to_i never seems to return nil:

```rb
irb(main):006:0> ['1', '2', '', 'hello', nil].map(&:to_i).compact
=> [1, 2, 0, 0, 0]
irb(main):007:0> ['1', '2', '', 'hello', nil].filter_map(&:to_i)
=> [1, 2, 0, 0, 0]
```
Doesn't look like a change of behavior, so seems fine - just wanted to check!

Throne3d (approved):
LGTM!

app/controllers/posts_controller.rb
teceler: Oh, yeah, to_i shouldn't return false, so the behavior change between map+compact and filter map I noted in the comment is probably fine.

PR #1666, opened 2022-06-27
Include icon helper methods in UserMailer

Marri:
Fixes a crash where icon_tag is not found


teceler (approved):
This should work? The main question is if you need PostHelper as well, and iirc you checked that. ...that and 'why didn't our tests catch this', I guess.

PR #1674, opened 2022-07-26
Fix/paste bug

Marri:
We added the paste plugin to TinyMCE in this commit https://github.com/Marri/glowfic/commit/c5cb4bf94e9144de9d0d1a3fd0fff3ef319338e5 as an attempt to try and fix the font size fuckery we'd been seeing as noted here https://github.com/Marri/glowfic/pull/959 but I think we fixed the font size issue here https://github.com/Marri/glowfic/pull/1370 and we're now seeing issues with users using our default layout unable to paste in the RTF error because it causes type errors. Why? WHO KNOWS. But if we don't need the plugin anymore and it's causing issues, might as well :fire:

### DIFF
--- a/app/assets/javascripts/posts/editor.js
+++ b/app/assets/javascripts/posts/editor.js
@@ -287,10 +287,10 @@ function setupTinyMCE() {
     tinyMCE.init({
       selector: selector,
       menubar: false,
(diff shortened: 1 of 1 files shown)


teceler (approved):
Looks good, assuming it fixes the bug on staging and etc.

PR #1677, opened 2022-08-27
Link character from reply editor

Marri:
All hail @andaisq

### DIFF
--- a/app/assets/javascripts/posts/editor.js
+++ b/app/assets/javascripts/posts/editor.js
@@ -152,7 +152,7 @@ function setupWritableEditor() {
     // Set the ID
     var id = $(this).val();
     $("#reply_character_alias_id").val(id);
-    $("#post-editor .post-character #name").html($('#character_alias option:selected').text());
+    $("#post-editor .post-character #name").text($('#character_alias option:selected').text());
     $('#alias-selector').hide();
     $("#post-editor .post-character").data('alias-id', id);
   });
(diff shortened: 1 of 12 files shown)


app/views/posts/_editor.haml
```diff
@@ -12,7 +12,7 @@
     #current-icon-dropdown= dropdown_icons(item, character_galleries)
     .post-info-text
       .post-character{class: ('hidden' unless item.character), data: { character_id: item.character_id || '', alias_id: item.character_alias_id || '' }}
-        %span#name= item.name
+        %span#name= link_to item.character_name, character_path(item.character_id)
```
Marri, line 15: You do in fact need to use `item.name` here, it handles aliases gracefully (it's defined here: https://github.com/Marri/glowfic/blob/master/app/concerns/writable.rb#L47)

app/assets/javascripts/posts/editor.js
```diff
@@ -152,7 +152,8 @@ function setupWritableEditor() {
     // Set the ID
     var id = $(this).val();
     $("#reply_character_alias_id").val(id);
-    $("#post-editor .post-character #name").html($('#character_alias option:selected').text());
+    var correctID = $("input#reply_character_id")[0].value
+    $("#post-editor .post-character #name").html(`<a href="/characters/${correctID}">${$('#character_alias option:selected').text()}</a>`);
```
Marri, line 156: Per discussion in Discord these would be safer as JQuery nodes, e.g. `$("<a>").text(name)` etc

app/assets/javascripts/posts/editor.js
```diff
@@ -381,7 +382,8 @@ function setAliases(aliases, name) {
 
 function setAliasFromID(selectedAliasID) {
   var correctName = $("#character_alias option[value=\""+selectedAliasID+"\"]").text();
-  $("#post-editor .post-character #name").html(correctName);
+  var correctID = $("input#reply_character_id")[0].value
```
Marri, line 385: We can probably pass this as an argument to the function instead; it's available everywhere we call this method from.

app/assets/javascripts/posts/editor.js
```diff
@@ -152,7 +152,8 @@ function setupWritableEditor() {
     // Set the ID
     var id = $(this).val();
     $("#reply_character_alias_id").val(id);
-    $("#post-editor .post-character #name").html($('#character_alias option:selected').text());
+    var correctID = $("input#reply_character_id")[0].value
```
Marri, line 155: I think you can do `.val()` instead of `[0].value`?

app/assets/javascripts/posts/editor.js
Marri:
```suggestion
    var nameNode = $("<a>").attr("href", "/characters/"+correctID).text($('#character_alias option:selected').text());
    $("#post-editor .post-character #name").html(nameNode);
```

app/assets/javascripts/posts/editor.js
```diff
@@ -353,7 +355,9 @@ function setInfoBoxFields(characterId, name, screenname) {
     characterNameBox.hide();
   }
   characterNameBox.data('character-id', characterId);
-  $("#post-editor .post-character #name").html(name);
```
Marri, line 356: Just delete this one, it's the one we found wasn't doing anything :fire:

PR #1689, opened 2022-10-09
Update bundler, nodejs and eslint for equality between envs

Throne3d:
Heroku uses fixed versions of bundler (2.3.10) and nodejs (16.x), which we were outdated on. Our GitHub Actions check uses a different version of eslint from the one installed in our Dockerfile, so I've also bumped that!

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,3 +1,8 @@
+extends:
+  - eslint:recommended
+
+root: true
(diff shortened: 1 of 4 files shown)


teceler (approved):
I don't seem to get different warnings or etc between any of the eslint versions here, and keeping things synced seems in general good

PR #1690, opened 2022-10-10
Convert CharacterGroups to tags

teceler:
* Adds TemplateTags
* Adds a CharacterGroup tag type
* Renames old CharacterGroup model to OldCharacterGroup
* Adds model connections
* Migrates old-style character groups to tags
* Adds scopes to interact with groups
* Rewrites part of characters#index and most of the group partial to handle new underlying infra
* Updates factories and model setup to handle new underlying parts
* Updates character relocator to check for new-style groups


teceler:
This is below coverage because OldCharacterGroup is still in (for migration reasons) and untested (because nothing calls it). Unsure whether to drop the coverage or make a dummy test or what.

PR #1691, opened 2022-10-10
Standardize flash messages between controllers

teceler:
Derived from #1359 but without the additional delegation and logging.


spec/controllers/users_controller_spec.rb
```diff
@@ -448,7 +448,7 @@
       }
 
       expect(response).to redirect_to(edit_user_url(user))
-      expect(flash[:success]).to eq('Changes saved successfully.')
+      expect(flash[:success]).to eq('Changes saved.')
```
Throne3d, line 451: There's one more example in this file that needs updating (I can't seem to comment on it to fix it) but then this LGTM!

Throne3d (approved):

PR #1692, opened 2022-10-12
Test Presenters

teceler:
* Creates hash matching test helper
* Tests a number of untested presenter / as json branches
* Cleans up checking if presented item is nil
* Cleans up some presenter return conditions
(Split out from #1549)

### DIFF
--- a/spec/support/spec_test_helper.rb
+++ b/spec/support/spec_test_helper.rb
@@ -32,6 +32,8 @@ module SpecTestHelper
         transform(ele)
       elsif ele.is_a?(ActiveRecord::Base)
         ele
+      elsif ele.is_a?(ActiveSupport::TimeWithZone)
+        ele.in_time_zone.to_s(:iso8601)
       else
         ele.to_s
       end
--- a/spec/support/spec_test_helper.rb
+++ b/spec/support/spec_test_helper.rb
@@ -33,7 +33,7 @@ module SpecTestHelper
       elsif ele.is_a?(ActiveRecord::Base)
         ele
       elsif ele.is_a?(ActiveSupport::TimeWithZone)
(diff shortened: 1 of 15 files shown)


spec/support/spec_test_helper.rb
```diff
@@ -6,6 +6,40 @@ def login_as(user)
   def login
     login_as(create(:user))
   end
+
+  RSpec::Matchers.define :match_hash do |expected|
```
Throne3d, line 10: Maybe we should name this `match_presented_hash` or something, since it does more complex transformations?

spec/support/spec_test_helper.rb
```diff
@@ -6,6 +6,40 @@ def login_as(user)
...
+        ele.to_a
+      elsif ele.is_a?(Array)
+        ele = ele.sort unless ele[0].is_a?(Hash)
+        ele.map { |e| transform_element(e) }
+      elsif ele.is_a?(Hash)
+        transform(ele)
+      elsif ele.is_a?(ActiveRecord::Base)
+        ele
+      else
+        ele.to_s
```
Throne3d, line 36: I guess we might need to edit this to support timestamps based on the current diff we're seeing. I'd prefer the format with the `T`, especially if that's the one we're currently returning, since it's RFC 3339 compliant, and I think datetime.to_s doesn't follow that by default - not sure.

Throne3d (commented):
Nice! Lots of additional tests \o/

spec/support/spec_test_helper.rb
teceler: idr exactly how much match_array did when I looked at it. I could use that if you think it's important?

Throne3d (approved):

PR #1698, opened 2022-11-22
Terminology fix in replies/edit.haml

teceler:
For some reason the page header is 'edit post' rather than 'edit reply'


Throne3d (approved):

Throne3d:
(Looks like rspec needs some expectation updates here)

PR #1709, opened 2023-01-28
Fix discord link

teceler:


app/views/sessions/index.haml
```diff
@@ -9,6 +9,5 @@
     Join our community on:
   %ul
     %li= link_to "Alicorn's forum", 'http://alicorn.elcenia.com/board/index.php', target: '_blank', rel: 'noopener'
-    %li= link_to "IRC", 'https://client00.chat.mibbit.com/?channel=%23glowfic&server=irc.psigenix.net', target: '_blank', rel: 'noopener'
-    %li= link_to "Glowfic Community Discord", Glowfic::DISCORD_LINK_GLOWFIC, target: '_blank', rel: 'noopener'
+    %li= link_to "Glowfic Community Discord", ENV.fetch('DISCORD_LINK_GLOWFIC'), target: '_blank', rel: 'noopener'
```
Marri, line 12: Should we have it display some sort of note if we just delete the link? Or at least hide the row if the env variable doesn't exist? That way we have a fallback if we ever need to disable invites for a while again.

Marri (approved):

PR #1711, opened 2023-01-29
Upgrade to Ruby 3.0.5

Throne3d:
Relies on our Rails 6.1 upgrade (to make gem changes conflict less). Includes an update to our pg gem to remove a deprecation warning, and makes some of our version constraints more liberal.

Initially, it seems like this is just ... working? Created a post, tested out the rich text editor, things just seem to naively work!


Marri (changes requested):
.rubocop.yml needs updating too

Marri (approved):
:turtle:

Gemfile
```diff
@@ -19,18 +19,18 @@ gem 'barnes' # heroku ruby-specific metrics
 gem 'bootstrap', '~> 4.5' # pin until major version is handled
 gem 'exception_notification'
 gem 'get_process_mem'
-gem 'gon', '~> 6.4.0'
+gem 'gon', '~> 6.4'
```
teceler, line 22: I think this will match any minor version of 6.0, which may or may not be what you intended?

Throne3d: It is! They should all be compatible, and most of the other constraints allow minor upgrades (the exceptions are rails and audited, where we have some custom patches / logic depending on internals, afaik, which we don't here)

teceler: That makes sense! just wanted to make sure it was the intended behavior, that's tripped me up before.

Throne3d: Makes sense! Thanks =)

PR #1712, opened 2023-01-29
Bump redis and redis-namespace

Throne3d:
Solves a deprecation warning:

```
glowfic-worker-1    | Pipelining commands on a Redis instance is deprecated and will be removed in Redis 5.0.0.
glowfic-worker-1    | 
glowfic-worker-1    | redis.pipelined do
glowfic-worker-1    |   redis.get("key")
glowfic-worker-1    | end
glowfic-worker-1    | 
glowfic-worker-1    | should be replaced by
glowfic-worker-1    | 
glowfic-worker-1    | redis.pipelined do |pipeline|
glowfic-worker-1    |   pipeline.get("key")
glowfic-worker-1    | end
glowfic-worker-1    | 
glowfic-worker-1    | (called from /usr/local/bundle/gems/redis-namespace-1.8.2/lib/redis/namespace.rb:530:in `namespaced_block'}
```

(Bumped redis-namespace intentionally and the Redis change came with, but seems harmless enough)


teceler (approved):
Both changelogs look safe to me, assuming we aren't using the affected add/remove methods. (for future reference, [redis-namespace changelog](https://github.com/resque/redis-namespace/blob/master/CHANGELOG.md#182))

PR #1714, opened 2023-02-14
Fix Rubocop lints

Throne3d:
Fixes some cops that were apparently introduced in the Rails 6.1 upgrade! Supersedes #1713, since it doesn't actually rely on the Ruby upgrade at all.


Marri (approved):
:turtle: reasonably confident the mailer name is irrelevant yes

PR #1736, opened 2023-09-09
Update nodejs installation process

Throne3d:
Previous process is deprecated: https://github.com/nodesource/distributions#new-update-%EF%B8%8F

Blocked on #1735 for the selenium issue!


Marri (commented):
We're still good with 16 right?

Throne3d:
Yep! It still seems to be on 16.x (kinda surprised me, tbh) - https://devcenter.heroku.com/articles/ruby-support#installed-binaries

teceler (approved):
This looks fine, other than 'possibly we should be using the new apt key syntax for pgdg as well'

Throne3d:
The postgres installation instructions haven't updated for this yet, but I'll try it locally and add a change if it works out!

PR #1738, opened 2023-09-09
Bump Ruby to 3.2.2

Throne3d:


teceler (approved):
[Ruby changelog for 3.2](https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/)
It looks like the main thing that affects us is improved regex matching, but possibly we want Marri to check this too, given how much relies on ruby working.

PR #1780, opened 2023-09-23
Upgrade Rails config defaults to 6.1

Throne3d:


config/initializers/new_framework_defaults_6_0.rb
```diff
@@ -1,45 +0,0 @@
...
-# Return false instead of self when enqueuing is aborted from a callback.
-# Rails.application.config.active_job.return_false_on_aborted_enqueue = true
-
-# Send Active Storage analysis and purge jobs to dedicated queues.
-Rails.application.config.active_storage.queues.analysis = :active_storage_analysis
-Rails.application.config.active_storage.queues.purge    = :active_storage_purge
-
-# When assigning to a collection of attachments declared via `has_many_attached`, replace existing
-# attachments instead of appending. Use #attach to add new attachments without replacing existing ones.
-Rails.application.config.active_storage.replace_on_assign_to_many = true
```
Throne3d, line 31: Not sure that all of these are safe (e.g. this one). It's important for us to upgrade configs over time, though. I'll try digging into each item and validating that it all makes sense.

config/application.rb
```diff
@@ -53,7 +53,7 @@ def self.full(text)
 
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 5.2
+    config.load_defaults 6.1
```
Throne3d, line 56: Turns out active_record.has_many_inversing isn't enabled until this is enabled (the config initializer doesn't take effect properly - https://github.com/rails/rails/issues/45683), and it also seems to be messing with our reply.rb item:

```rb
belongs_to :post, inverse_of: :replies, optional: false
```
somehow making the "Post must exist" validation stop executing. For now I'll keep the has_many_inversing config reverted and we can try figuring out why that's breaking the `optional: false` validation later.

PR #1783, opened 2023-09-24
Add account config to default hide retired characters

Throne3d:


teceler (approved):
Looks good! Other than the horrifying param bit but I'm not sure how fixable that is. Params are just terrible sometimes.

PR #1809, opened 2023-10-12
NPC Box

### LINKED ISSUE #425: NPC box
Probably done with aliases on the main user account that then display. This makes it lower barrier to add accounts for NPCs, since they won't show up in your large list of characters and you don't need to facecast them or anything, but you can still name them to make it clear who's who.

There should be a way to then convert your NPC posts into an actual character (characters#replace on main user account in specific posts with a specific alias?).

This likely relies on alias creation being made easier (in the reply editor or at least faster on the character page) – #171.

Throne3d:
Fixes #425.

Basically, users want a way to name the NPCs they write with, without creating full character accounts for them that pollute their Characters list (esp if the NPC is only used in a single thread, and e.g. uses the author character icon, and so on).

Here we enable that by adding a new "NPC" category to the character selection box when writing a post or reply, and letting you select either from your existing list of NPCs (disambiguated by thread name, hacked into the "nickname" section, and incl. separate "thread NPCs" in case that's easier), or create a new NPC via the select2 tags process (this works fine for single selects, too: it just creates a new `<option>` based on your input). These are persisted as characters with a new is_npc flag behind the scenes, which hides them from relevant pages but still allows us to collect all the metadata on where this character is used or e.g. allows the user to provide a bit more information on the NPC if they desire.

Due to how ReplyDrafts work, in case we create them (reply#create preview + draft, but not reply#update preview nor on posts), we also create the character just prior, which may leave orphan characters lying around if the draft is never posted. This should be relatively rare - and in other cases, we only persist the character when its reply/post is created (at which point they could be orphaned by deleting the post/reply anyway), so it doesn't seem particularly problematic.

Whichever icon is used for the first reply/post with the new NPC is treated as its default icon, though that can be tweaked on the character edit page for the NPC, and similarly you can add other galleries. Otherwise it behaves similarly to the author character, letting you select either your author icon or no icon: NPC icons can be temporarily moved to your avatar if you need to set an icon for a new NPC, which is a bit awkward in terms of UX, or you can just post with another icon, edit the NPC to add the right gallery and default icon, and edit the reply to have the right icon after that. This seems like a decent compromise for this initial iteration!

These NPC characters also show up in the quick switcher, which was an important consideration for the end users!

The character edit page disables the screenname, template and cluster inputs when NPC is selected, as those (per users) don't make much sense, but it doesn't clear out the values, so users can enter data there if they really want by temporarily disabling the NPC config! =) Additionally, a character can be transitioned to/from NPC state if necessary, which should make NPC "promotions" straightforward if an NPC starts getting fleshed out. NPCs should also appear in the character replace menu, at least in this first iteration, which should make it easy to transition to/from this new approach.

Currently there's no way to forcibly show all NPCs for a user on their character page - NPCs are ignored, similar to retired characters, but with no toggle. This didn't seem obligatory to provide in a first release, and we can add this functionality at a later point if we need to.

UX:

image
image
image
image
image
image
image
image

WIP:

- [x] fix linting
- [x] reply drafts don't work with it yet (I'm thinking I might just persist the character and use the ID when you click preview / draft, to make things easier, but I'm not sure - maybe I'll store the NPC name on the draft so we can create it only when you submit the reply?)
- [x] if you have multiple NPCs with the same name, right now there's no way to properly disambiguate them: thinking of adding recent thread names to the dropdown to help here, but a little concerned about the SQL involved
- [x] fix tests (esp the character presenter tests are probably surprised by the new is_npc field)
- [x] add tests (esp UI tests, I want to make sure this works properly with JavaScript in a way we won't break)
- [x] a missing variable around the quick switcher, still need to figure out how I'm gonna solve that
- [x] display on a character's page that it's an NPC (and maybe hide nickname?)
- [x] consider disabling fields for screenname, template, cluster when enabling NPC on the character edit page
- [x] ~~make it possible to show NPC characters on a user's page? but don't show it by default~~ -> later follow-up


db/migrate/20230930053037_add_is_npc_to_characters.rb
```diff
@@ -0,0 +1,5 @@
+class AddIsNpcToCharacters < ActiveRecord::Migration[6.1]
+  def change
+    add_column :characters, :is_npc, :boolean
```
Throne3d, line 3:
```suggestion
    add_column :characters, :is_npc, :boolean, default: false, null: false
```

app/assets/javascripts/posts/editor.js
```diff
@@ -186,6 +186,8 @@ function hideSelect(target, selectBox, selectHolder) {
 
 function fixWritableFormCaching() {
   // Hack to deal with Firefox's "helpful" caching of form values on soft refresh (now via IDs)
+  // TODO: fix caching for is_npc + the input field for creating it?
+  // TODO: Retrigger select2 filter based on the input?
```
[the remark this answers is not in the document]
Throne3d:
```suggestion
```

app/assets/javascripts/posts/editor.js
```diff
@@ -306,33 +311,38 @@ function setupTinyMCE() {
...
   setGalleriesAndDefault(resp.galleries, resp.default_icon);
   setIcon('');
-  if (restoreIcon)
+  if (restoreIcon) {
     setIconFromId(selectedIconID);
-  else if (resp.default_icon)
+  } else if (resp.default_icon) {
     setIcon(resp.default_icon.id, resp.default_icon.url, resp.default_icon.keyword, resp.default_icon.keyword);
+  }
 }
```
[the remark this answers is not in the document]
Throne3d: don't think I changed this tbh, ignoring

app/controllers/writable_controller.rb
```diff
@@ -170,6 +177,19 @@ def og_data_for_post(post, page: 1, total_pages:, per_page: 25)
     }
   end
 
+  def process_npc(writable, permitted_character_params)
+    return unless writable.character.nil?
+    return unless permitted_character_params[:is_npc] == 'true'
+
+    post_name = if writable.is_a? Post
```
Throne3d, line 184:
```suggestion
    # we take the NPC's first post's subject as its nickname, for disambiguation in dropdowns etc
    post_name = if writable.is_a? Post
```

app/controllers/posts_controller.rb
```diff
@@ -329,6 +331,10 @@ def preview
     @post.assign_attributes(permitted_params(false))
     @post.board ||= Board.find_by_id(3)
 
+    if permitted_character_params[:is_npc] && @post.character_id.nil?
+      @post.build_character(permitted_character_params.merge(default_icon_id: @post.icon_id))
+    end
```
Throne3d, line 336:
```suggestion
    process_npc(@post, permitted_character_params)
```

app/controllers/writable_controller.rb
```diff
@@ -7,19 +7,29 @@ def build_template_groups(user=nil)
...
     if @post
       uniq_chars_ids = @post.replies.where(user_id: user.id).where.not(character_id: nil).group(:character_id).pluck(:character_id)
       uniq_chars_ids << @post.character_id if @post.user_id == user.id && @post.character_id.present?
-      uniq_chars = Character.where(id: uniq_chars_ids).ordered.pluck(pluck)
+      uniq_chars = Character.where(id: uniq_chars_ids, is_npc: false).ordered.pluck(Template::CHAR_PLUCK)
       threadchars = faked.new('Thread characters', nil, uniq_chars)
       @templates.insert(0, threadchars)
+
+      uniq_npcs = Character.where(id: uniq_chars_ids, is_npc: true).ordered.pluck(Template::NPC_PLUCK)
+      threadnpcs = faked_npcs.new('Thread NPCs', nil, uniq_npcs)
```
teceler, line 28:
```suggestion
      threadnpcs = faked_npcs.new('Post NPCs', nil, uniq_npcs)
```

app/controllers/writable_controller.rb
```diff
@@ -7,19 +7,29 @@ def build_template_groups(user=nil)
...
     @templates = templates + [templateless]
+    all_npcs = faked_npcs.new('All NPCs', nil, user.characters.where(retired: false, is_npc: true).ordered.pluck(Template::NPC_PLUCK))
+    @npcs = [all_npcs]
 
     if @post
       uniq_chars_ids = @post.replies.where(user_id: user.id).where.not(character_id: nil).group(:character_id).pluck(:character_id)
       uniq_chars_ids << @post.character_id if @post.user_id == user.id && @post.character_id.present?
-      uniq_chars = Character.where(id: uniq_chars_ids).ordered.pluck(pluck)
+      uniq_chars = Character.where(id: uniq_chars_ids, is_npc: false).ordered.pluck(Template::CHAR_PLUCK)
       threadchars = faked.new('Thread characters', nil, uniq_chars)
```
Throne3d, line 24:
```suggestion
      threadchars = faked.new('Post characters', nil, uniq_chars)
```

app/assets/stylesheets/replies.scss
```diff
@@ -40,7 +40,7 @@
   padding: 5px 20px 10px 20px;
   z-index: 2;
   background-color: $bg_color_selectionpopup;
-  min-width: 200px;
+  min-width: 250px;
```
teceler, line 43: Why was this changed?

app/views/posts/_editor.haml
```diff
@@ -26,11 +26,18 @@
...
+              = select_tag :active_character,
+              option_groups_from_collection_for_select(@templates, :plucked_characters, :name, :first, :last, item.character_id),
+              { prompt: current_user.username, class: 'chosen-select' }
+            #swap-character-npc
+              %h4 Choose NPC
+              = select_tag :active_npc,
+              option_groups_from_collection_for_select(@npcs, :plucked_npcs, :name, :first, :last, item.character_id),
+              { prompt: "Select NPC or type to create" }
+            %button.view-button#select-character{type: 'button'} Character
+            %button.view-button#select-npc{type: 'button'} NPC
```
teceler, line 40: What is this doing?

spec/controllers/characters_controller_spec.rb
```diff
@@ -74,9 +74,11 @@
       it "skips retired characters when specified" do
```
teceler, line 74: probably want to rename this and also test that it hides npcs even when retired status isn't specified? might be better to just make a test for that tbh

teceler:
Notes that aren't linked to a specific code point:
* maybe scope `.where(is_npc: false)`
* test character#show for npc
* test adding an existing npc to a new post
* test adding an existing npc to the same post

app/assets/stylesheets/replies.scss
Throne3d: Oh! With the previous value and the new buttons, the UI sometimes collapsed weirdly. Let me find a screenshot quickly!

app/views/posts/_editor.haml
Throne3d: This lets you select whether you're seeing the dropdown for NPCs or Characters in the select box - here:

image

The dropdown functions differently on the NPCs page, letting you create a new one, but this also creates a clear separation between the two types of characters, which seemed very desirable based on the user feedback!

app/assets/stylesheets/replies.scss
Throne3d: 200px minimum
image

250px minimum
image

(This way we always keep the NPC + Character buttons on the same line, rather than wrapping them)

I can go more aggressive on the minimum, down to 220px, which maintains that property and has a lower chance of messing with smaller UIs - that might be good?

image

Throne3d:
> test character#show for npc

I can add a test explicitly for this if we want, especially if we want a controller spec for it, but the `spec/features/characters/edit_spec.rb` and `spec/features/characters/new_spec.rb` tests are actually already validating the character#show rendering after updating NPCs!

> maybe scope .where(is_npc: false)

That could be good - do we have similar for retired? I don't think I want to make it a default scope (like I mentioned to Marri, ideally this initial step will leave the new NPC-type characters around places that we haven't evaluated we want to hide them from, so people have more flexibility rather than less), but I think it makes sense to add it as a quick scope somewhere.

> test adding an existing npc to a new post
> test adding an existing npc to the same post

Ooh, right, good idea - not sure why I was only testing new NPCs (I guess old NPCs should fall into the basic character_id flow, and it was safe in manual testing, but there's no guarantee it remains that way). I'll add tests for this!

app/views/posts/_editor.haml
teceler: Oh! I see. ...I'm not sure how I feel about having npcs being that obtrusive but it does look like it would work

spec/controllers/characters_controller_spec.rb
Throne3d: Not sure why I combined this, it really should've been separate - thanks!

app/views/posts/_editor.haml
Throne3d: Yeah, I wasn't sure where else to put it but I think this menu doesn't get used as much as the quick switcher so it should help hide it from most flows, at least? Red seems to think he'll use NPCs a lot and liked the clear separation (rather than polluting the existing characters list with it), as otherwise they'd be more obtrusive elsewhere. Maybe someone will come up with a better UI suggestion when it's released and we can move it? But I think it'll probably take some time for people to get familiar with it and know how often it should show up / where we should move it to based on experience with it - I didn't have many other ideas for how to incorporate it, really.

app/assets/stylesheets/replies.scss
Throne3d: Do you think we should drop this to 220px, or is 250px ok?

PR #1811, opened 2023-10-12
Fixes after CodeClimate lints

Throne3d:
A few different issues from https://github.com/glowfic-constellation/glowfic/pull/1598:

* after 1727b67c8d83414e80c1aa75aaf23930a11da648, we're trying to call .hide on a plain string, rather than a jQuery object! Broke the post/reply editor character box disappearing when you click outside its box.
* after ec0a965b52c611a5ff29304fac6beb48c3be28f4, keybindings to quickly go through forms in the galleries#add_new page were broken by some scoping rules around $(this)
* (Unrelated issue: galleries#add_new's keybindings weren't working for new rows properly, because they were being inserted in the tfoot instead of the tbody)
* after 76546f877c2d133daa8528df42225e2e53dfe5f6, icon submit was broken after removing an uploaded icon and going through the deleteUnusedIcons flow, because jQuery event propagation was being blocked by `return false`

### DIFF
--- a/app/assets/javascripts/galleries/add_new.js
+++ b/app/assets/javascripts/galleries/add_new.js
@@ -17,6 +17,7 @@ $(document).ready(function() {
   });
 });
 
+// eslint-disable-next-line complexity
 function processDirectionalKey(event) {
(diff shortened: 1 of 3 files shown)


teceler (approved):
That looks like it should fix the problem! I'm a little concerned there might be similar instances though.

PR #1812, opened 2023-10-12
Fix minio icon upload with Docker approach

Throne3d:
The minio:9000 guest isn't usually available from the Docker host under that name, but instead binds to localhost with the given port. This allows the webserver to connect to minio directly inside the Docker network, and also allows the host to connect to the minio instance from the UI.


Throne3d:
Coverage 😤

PR #1819, opened 2023-10-14
Initial attempt at parallel tests

Throne3d:
Seems flaky with parallelism >4 in CI (and maybe locally too). Should investigate, but for now this is probably blocked?

Does speed things up with parallelism 2 though - 6m21s without, 4m9s with.


Throne3d:
Seems flaky with parallelism >4 in CI (and maybe locally too). Should investigate, but for now this is probably blocked?

Does speed things up with parallelism 2 though - 6m21s without, 4m9s with.

teceler:
We now have parallel tests

PR #1822, opened 2023-10-15
Use case insensitive searches by default

### LINKED ISSUE #1247: Exact match reply search is case sensitive
Might need a check box, might need to just not do the thing (currently "How were you a detective" matches but "how were you a detective" does not)

Throne3d:
Uses ILIKE for case insensitive searches (via database locale) rather than forcing downcase matches

Fixes #1247, and possibly other situations - though this might have undesirable effects in some situations if people are expecting case sensitive matches? I think all of these situations make sense to use case insensitive matches - including the API - but it's possible there are cases where we want case sensitivity that I'm not considering.

### DIFF
--- a/app/controllers/api/v1/posts_controller.rb
+++ b/app/controllers/api/v1/posts_controller.rb
@@ -10,7 +10,7 @@ class Api::V1::PostsController < Api::ApiController
   param :q, String, required: false, desc: 'Subject search term'
   def index
     queryset = Post.order(Arel.sql('LOWER(subject) asc'))
-    queryset = queryset.where('subject ILIKE ?', "%#{params[:q].downcase}%") if params[:q].present?
+    queryset = queryset.where('subject ILIKE ?', "%#{params[:q]}%") if params[:q].present?
 
     posts = paginate queryset, per_page: 25
     posts = posts.visible_to(current_user)
--- a/app/controllers/api/v1/posts_controller.rb
+++ b/app/controllers/api/v1/posts_controller.rb
@@ -10,7 +10,7 @@ class Api::V1::PostsController < Api::ApiController
   param :q, String, required: false, desc: 'Subject search term'
   def index
     queryset = Post.order(Arel.sql('LOWER(subject) asc'))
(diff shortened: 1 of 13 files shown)


app/controllers/api/v1/posts_controller.rb
```diff
@@ -10,7 +10,7 @@ class Api::V1::PostsController < Api::ApiController
   param :q, String, required: false, desc: 'Subject search term'
   def index
     queryset = Post.order(Arel.sql('LOWER(subject) asc'))
-    queryset = queryset.where('LOWER(subject) LIKE ?', "%#{params[:q].downcase}%") if params[:q].present?
+    queryset = queryset.where('subject ILIKE ?', "%#{params[:q].downcase}%") if params[:q].present?
```
Throne3d, line 13:
```suggestion
    queryset = queryset.where('subject ILIKE ?', "%#{params[:q]}%") if params[:q].present?
```

app/controllers/posts_controller.rb
```diff
@@ -286,10 +286,10 @@ def search
     @search_results = @search_results.where(id: Setting.find(params[:setting_id]).post_tags.pluck(:post_id)) if params[:setting_id].present?
     if params[:subject].present?
       if params[:abbrev].present?
-        search = params[:subject].downcase.chars.join('% ')
-        @search_results = @search_results.where('LOWER(subject) LIKE ?', "%#{search}%")
+        search = params[:subject].chars.join('% ')
+        @search_results = @search_results.where('subject ILIKE ?', "%#{search}%")
       else
-        @search_results = @search_results.search(params[:subject]).where('LOWER(subject) LIKE ?', "%#{params[:subject].downcase}%")
+        @search_results = @search_results.search(params[:subject]).where('subject ILIKE ?', "%#{params[:subject].downcase}%")
```
Throne3d, line 292:
```suggestion
        @search_results = @search_results.search(params[:subject]).where('subject ILIKE ?', "%#{params[:subject]}%")
```

PR #1823, opened 2023-10-15
Clarify empty icons aren't broken

### LINKED ISSUE #1191: Clarify that the empty boxes in icon view are filler for no icon, rather than being a broken icon
As reported in #1189 it's not clear from the UI if the empty boxes represent a broken image or not; we should make this clearer

Throne3d:
Fixes #1191 

Update, now looks like this:

image

Previously just shows an empty box:

image

Temporary intermediate one we discarded:

image

### DIFF
--- a/app/assets/stylesheets/characters.scss
+++ b/app/assets/stylesheets/characters.scss
@@ -40,7 +40,12 @@
   border: 1px $border_color_character_iconless solid;
   width: 102px;
   height: 102px;
+  padding: 20px 5px;
   text-align: center;
+
+  color: $font_color_page;
+  font-size: 90%;
+  font-style: italic;
 
   &.icon {
     max-width: 102px;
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/characters.scss
```diff
@@ -40,8 +40,13 @@
   border: 1px $border_color_character_iconless solid;
   width: 102px;
   height: 102px;
+  padding: 20px 5px;
   text-align: center;
 
+  color: $font_color_page;
```
Throne3d, line 46: Without this, on styles that have a specific color for link hover, we see the text inside the icon change color on hover, which seems strange (it seems like it should be "fixed" as a property of the image).

teceler:
Oh. don't like that. it looks incredibly Wrong xD

Throne3d:
Any other suggestions? :D

Throne3d:
Updated look:

image

PR #1824, opened 2023-10-15
Convert #loading and #saveerror on reorder to classes

### LINKED ISSUE #1151: Boards#edit invalid html
When it has both sections and sectionless posts (example: effugence in new seeds), the `#loading` and `#saveerror` ids exist twice.

Throne3d:
Fixes an issue with duplicate IDs on this page when handling both sections + sectionless posts.

Fixes #1151. Page still loads and the loading indicator shows up as expected - we were even already filtering our $("#loading, ...) scope to within a specific box, so it's not like it was expecting this to be globally distinct anyway.

### DIFF
--- a/app/assets/javascripts/reorder.js
+++ b/app/assets/javascripts/reorder.js
@@ -119,7 +119,7 @@ function syncRowOrders(orderBox, path, param) {
   // Reduce race conditions by only allowing one update at a time
   unbindArrows(orderBox);
   disableSortable(orderBox);
-  $("#loading", orderBox).show();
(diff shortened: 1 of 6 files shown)


teceler (approved):
Looks good, assuming you checked for instances of loading and saveerror that were using a different way to indicate they were looking for an id?

Throne3d:
Yep! I wasn't able to see `loading` many other places (except in another `#loading` context - `#loading`) and `saveerror` is found nowhere else =D

PR #1825, opened 2023-10-15
Log gallery load error for user

### LINKED ISSUE #1073: Indicate gallery loading failure more obviously
https://github.com/Marri/glowfic/blob/ad00696b4d2eeb0a3e19aec9fe6adf270a45fd79/app/assets/javascripts/galleries/expander.js#L90

Throne3d:
Fixes #1073

We log errors like this elsewhere in an `alert` - seems decent, and network issues should be more clear this way, if loading something fails due to an issue on the user's end, rather than just mysteriously not showing up.


teceler (approved):
:green_book:

PR #1826, opened 2023-10-15
Tweak default select2 height

### LINKED ISSUE #835: Increase the height of the character selector dropdown by half an item
On some platforms, the scrollbar for the character selector doesn't show. If we can modify the height of the select2 control, we should increase the height of the box by half an item. This might be problematic since the text in items wraps and could mess with the relevant heights.

Example of a failure to scroll (definitely has more items below this set):

Throne3d:
Increasing this from 200px to 220px makes it clear that partial results are available, even on platforms without scrollbars (etc)

Before:

image

After:

image

Fixes #835


teceler (approved):
:green_book:

PR #1827, opened 2023-10-15
Fix searching posts in Site testing

### LINKED ISSUE #1566: Don't filter site testing from search results if that's the searched board

Throne3d:
Fixes #1566


Throne3d:
Seems like the only other search page that takes `board` is replies, seem like there are any other search endpoints that take board, but that doesn't seem to be using the same posts_from_relation logic that bans Site testing, so I think we're good to go with just this!

PR #1832, opened 2023-10-21
Show NPCs on characters#index if in a template

Throne3d:
In general they shouldn't be in a template, but just to make sure the interface is consistent with templates#show which currently shows the NPC (and seems better from a discoverability perspective).


Throne3d:
-> not moving forward with this, since users are using this temporarily while we don't have a way to view all of a user's NPCs in a clear central place

PR #1873, opened 2023-11-21
Add sorbet for type checking

Throne3d:
No idea if we'll ever merge this, just wanted to test out how it'd look + what it'd be like!


Throne3d:
There's a truly insane amount of vendored code here - TODO: gitignore sorbet/rbi (against the upstream recommendation) and just tell people to generate these files locally?

PR #1915, opened 2024-01-07
Add delete reply draft button

Throne3d:
Flow:

image

-> enter data and save draft

image

-> click delete draft

image

-> confirm and back to original editor

image


Throne3d:
tests passed but needs better coverage

PR #1916, opened 2024-01-07
Avoid Rails.application.secrets in Rails 7.1

Throne3d:
The development/test key doesn't need to be kept secret, and we can
directly use ENV in other contexts. The replacement for secrets is
credentials, checking in an encrypted version of the secrets into
version control, but I'd prefer to keep using environment variables in
production to avoid checking in those secrets, even encrypted.


Throne3d:
again, coverage :eyes:

PR #1920, opened 2024-01-10
Upgrade cache_format_version to 7.0

Throne3d:
Once we're sure we're not going to roll back 7.1, we should go to 7.1
here too!

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -55,7 +55,7 @@ module Glowfic
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 6.1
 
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -55,6 +55,9 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 6.1
 
+    # use newer 7.1 cache format
```
Throne3d, line 58:
```suggestion
    # use newer 7.0 cache format
```

PR #1944, opened 2024-01-20
Set auth token more correctly, blank session on invalid API requests

Throne3d:
Previously the set_auth_token method was called every time we referenced current_user

Additionally, session[:api_token] is accessed by set_login_gon and similar, but using string keys - if the user didn't have an auth token before, and we generate it on this request, the gon value would be nil incorrectly.

Now that we blank the session on invalid API requests, users can also reload the page to get a new API token when an invalid request is sent. This should help get out of bad states more easily!

### DIFF
--- a/app/concerns/authentication/api.rb
+++ b/app/concerns/authentication/api.rb
@@ -36,10 +36,12 @@ module Authentication::Api
       ActiveSupport::HashWithIndifferentAccess.new(body)
     rescue JWT::ExpiredSignature
       error = { message: "Authorization token has expired." }
+      session[:api_token] = nil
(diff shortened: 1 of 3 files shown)


app/concerns/authentication/api.rb
```diff
@@ -36,10 +36,12 @@ def decode_value(value)
       ActiveSupport::HashWithIndifferentAccess.new(body)
     rescue JWT::ExpiredSignature
       error = { message: "Authorization token has expired." }
+      session[:api_token] = nil
```
Throne3d, line 39: (only has any effect if the user passed a valid CSRF token, after #1502)

PR #1952, opened 2024-01-21
Add ActiveStorage

teceler:


app/jobs/migrate_upload_job.rb
```diff
@@ -0,0 +1,55 @@
+require 'uri'
+require 'digest'
+
+class MigrateUploadJob < ApplicationJob
+  queue_as :low
+
+  def perform(icon_id)
+    icon = Icon.find(icon_id)
+    raise IconNotUploaded, "Icon #{icon.id} is remotely hosted" unless icon.s3_key.present?
```
amandaschloss, line 9: possibly for another PR but: do we want to do this for all icons? we've discussed eliminating remote hosting before, and if this does it anyway?

Marri, line 9: possibly for another PR but: do we want to do this for all icons? we've discussed eliminating remote hosting before, and if this does it anyway?

teceler: I don't *think* this is reuploading them like we'd want for that, iirc it's just entering them in activestorage with their current urls.

PR #1960, opened 2024-01-24
Fix ETag invalidation for posts#owed when reading a post

Throne3d:
`@posts` isn't updated when users read posts, so we need to include the unread_views entry for users' browsers to correctly update the page's arrow colors when they read posts.


Throne3d:
this doesn't solve it entirely - if you read a thread that has a lower updated_at than the latest one you've read then it's still an issue

PR #1964, opened 2024-01-25
Various linting work

teceler:
- Configure some cops that apparently got missed when they were added to the relevant rubocop gems
- Clean up some haml lints from, most likely, that not being checked when we enabled the cops on rubocop
- Make a specific rubocop file for app/views to work with haml-lint's config file, rather than a constantly out of date list of ignored cops

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -730,6 +730,9 @@ Capybara/NegationMatcher: # new in 2.14
   Enabled: true
   EnforcedStyle: have_no
 
+Capybara/RedundantWithinFind: # new in 2.20
+  Enabled: true
+
 Capybara/RSpec/HaveSelector: # new in 2.19
   Enabled: false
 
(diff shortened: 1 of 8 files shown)


app/views/.rubocop.yml
```diff
@@ -0,0 +1,19 @@
+inherit_from: ../../.rubocop.yml
```
Throne3d, line 1: Ooh, nice!

.rubocop.yml
```diff
@@ -753,6 +756,10 @@ FactoryBot/AssociationStyle: # new in 2.23
 FactoryBot/ConsistentParenthesesStyle: # new in 2.14
   Enabled: true
 
+FactoryBot/ExcessiveCreateList: # new in 2.25
```
Throne3d, line 759: Not sure how I missed these. Thanks for catching them!

Throne3d (approved):

PR #1975, opened 2024-02-08
Get haml-lint rubocop to chill

teceler:
I don't know why these passed locally but are having trouble on PRs but there's a simple solution


Throne3d:
probably due to #1971, let's review the changes to see if they make sense and maybe apply them?

but legit!

PR #2004, opened 2024-03-10
Make facecasts tags

teceler:


db/migrate/20240310003914_migrate_facecasts.rb
```diff
@@ -0,0 +1,21 @@
...
+          tag = Facecast.create!(name: fc, user_id: user_id) if tag.nil?
+          CharacterTag.create!(character_id: id, tag_id: tag.id)
+          galleries.each { |gid| GalleryTag.create!(gallery_id: gid, tag_id: id) }
+        end
+      end
+    end
+  end
+
+  def down
+    Facecast.each(&:destroy!)
```
Throne3d, line 19: We're not deleting the `pb` field on Character with this change, then? I think it might be nice to delete the column for cleanup, but if we do that, we need to restore it in `def down`, too, in case we need to revert things in prod 🤔

teceler: ...oh, hm, yeah, if we're doing this as a migration rather than a migration script (which we might not want to!) the logical thing would be to remove that in the migration too

spec/system/characters/new_spec.rb
```diff
@@ -29,7 +29,6 @@
     within('.form-table') do
       fill_in 'Template Nickname', with: 'Example nickname'
       fill_in 'Screen Name', with: 'example_screenname'
-      fill_in 'Facecast', with: 'Example facecast'
```
Throne3d, line 32: I'll see if we can test this with JS tests (per Discord convo)

PR #2033, opened 2024-04-08
Duplicate screenname message

TeegardenMath:
When creating a new character, display a notification if the chosen screenname is already in use (but don't block character creation).

I don't know if anyone else wanted this feature. But I like picking unique screennames like we're still on Dreamwidth.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_09_30_053037) do
+ActiveRecord::Schema[7.1].define(version: 2023_09_30_053037) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "citext"
   enable_extension "plpgsql"
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -154,6 +154,7 @@ RSpec.configure do |config|
       options.add_argument('--no-sandbox')
       options.add_argument("--user-data-dir=#{ENV['CHROMEDRIVER_CONFIG']}") if ENV['CHROMEDRIVER_CONFIG']
       options.add_argument('--window-size=1366,768')
+      options.add_argument('--disable-dev-shm-usage')
     end
   end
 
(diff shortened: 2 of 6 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_09_30_053037) do
+ActiveRecord::Schema[7.1].define(version: 2023_09_30_053037) do
```
compwron, line 13: interesting change - not related to the goal of the PR I think. Possibly a good idea to merge in general tho, it may have been left out of an earlier unrelated PR?

spec/spec_helper.rb
```diff
@@ -154,6 +154,7 @@
       options.add_argument('--no-sandbox')
       options.add_argument("--user-data-dir=#{ENV['CHROMEDRIVER_CONFIG']}") if ENV['CHROMEDRIVER_CONFIG']
       options.add_argument('--window-size=1366,768')
+      options.add_argument('--disable-dev-shm-usage')
```
compwron, line 157: interesting, TIL
https://peter.sh/experiments/chromium-command-line-switches/#disable-dev-shm-usage

> --disable-dev-shm-usage
> The /dev/shm partition is too small in certain VM environments, causing Chrome to fail or crash. Use this flag to work-around this issue (a temporary directory will always be used to create anonymous shared memory files).
> See the bug details: https://bugs.chromium.org/p/chromium/issues/detail?id=715363

spec/spec_helper.rb
```diff
@@ -35,7 +35,7 @@
       end
     end
     enable_coverage :branch
-    minimum_coverage line: 99.98, branch: 94.72
+    minimum_coverage line: 99.98, branch: 93.8
```
teceler, line 38: This seems to have snuck in

TeegardenMath: huh. the thing i did was
`git checkout main -- spec/spec_helper.rb`
which i expected to put it back to the default version (to undo the thing commented on upthread). any idea what's up with that?

teceler: That should! Possibly your local version of main was outdated, if we changed that recently? `git fetch` + `git checkout origin/main -- spec/spec_helper.r` should avoid that, i think

TeegardenMath: hmmm i have done those things and it's telling me nothing changed. baffling

PR #2146, opened 2024-08-11
Add more frozen_string_literal magic comments

Throne3d:


Throne3d:
This is looking fine locally! I'll merge it to staging and we can see how it goes.

PR #2150, opened 2024-08-11
Use pagination more to avoid large loads

teceler:
- Paginates large sectioned boards by section (per page: 25)
- Paginates large character views by template in template view IF the user has more than 50 templates AND the the user does not have character groups (per page: 25)
- Paginates ungrouped character pages IF the user has more than 100 characters (per page: 25 for list view, 50 for icon view)
- Paginates galleries#index (per page: 100)
- Paginates galleries#show (per page: 100)

### DIFF
--- a/app/views/boards/show.haml
+++ b/app/views/boards/show.haml
@@ -55,6 +55,11 @@
         %tr
           %td.continuity-spacer{colspan: 5}
       = render partial: 'posts/list_item', collection: @posts, as: :post, locals: { hide_continuity: true }
+
+    - if @board_sections.methods.include?(:total_pages) && @board_sections.total_pages > 1
+      %tfoot
+        %tr
+          %td= render 'posts/paginator', paginated: @board_sections, no_per: true
 - else
   - if @board.description.present?
     - content_for :post_list_description do
--- a/app/views/characters/index.haml
+++ b/app/views/characters/index.haml
@@ -74,8 +74,10 @@
       - if (group_chars = @user.characters.non_npcs.where(character_group_id: nil)).exists?
         = render 'group', characters: group_chars, group: nil, skip_grouped_templates: true, page_view: page_view, colspan: colspan
     - elsif @user.characters.exists?
-      = render partial: partial_type, collection: @user.templates.ordered, as: :name
-      - if (templateless = @user.characters.non_npcs.where(template_id: nil)).exists?
+      - templates = @user.templates.ordered
+      - templates.paginate(per_page: 25, page: params[:page]) if templates.count > 50
+      = render partial: partial_type, collection: templates, as: :name
+      - if (templateless = @user.characters.non_npcs.where(template_id: nil)).exists? && (templates.methods.exclude?(:total_pages) || templates.total_pages == params[:page])
         = render partial_type, name: "No Template", characters: templateless.ordered, show_new_character_button: @user.id == current_user&.id
     - else
       %tr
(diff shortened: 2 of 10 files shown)


spec/controllers/galleries_controller_spec.rb
```diff
@@ -280,6 +296,14 @@
         expect(meta_og[:title]).to eq("user » gallery")
         expect(meta_og[:description]).to eq("16 icons\nTags: Tag 1, Tag 2")
       end
+
+      it "paginates large galleryless" do
+        user = create(:user)
```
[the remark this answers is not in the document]
teceler: ...but I do use that? twice??

teceler:
This does not affect expandable galleries (I don't have better solution than having it just show an error and say to go to the gallery page) or people with character groups (terrifying).

I could add something to paginate template#show though

app/views/boards/show.haml
```diff
@@ -55,6 +55,11 @@
         %tr
           %td.continuity-spacer{colspan: 5}
       = render partial: 'posts/list_item', collection: @posts, as: :post, locals: { hide_continuity: true }
+
+    - if @board_sections.methods.include?(:total_pages) && @board_sections.total_pages > 1
```
Throne3d, line 59: (I can't comment on the relevant line but) I think I remember there being a bug with board sections where if you stick `?page=2` in the URL then it'd show the second page or something weird like that, but I don't know why I would think that (we don't even paginate the posts inside sections right now, do we? it has `with_pagination: false`)

But just to be sure, when we use `?page=2` on a board with sections, it shows the second page of sections, but all the posts in those sections?

teceler: I don't remember if I tested that particular case on the site! I'll look into that.

Throne3d: Also it's been forever since I've last done Ruby but I think the normal way of asking this (the methods thing) is `.responds_to?(:total_pages)` - do we have this `.methods.include?` pattern somewhere else?

app/views/characters/index.haml
```diff
@@ -74,15 +76,16 @@
...
       %tr
         %td.centered.padding-5{class: cycle('even', 'odd'), colspan: colspan} — No characters yet —
-  - elsif character_split == 'npcs'
-    - characters = @user.characters.npcs.ordered
-    = render partial_type, name: nil, characters: characters, show_template: false
   - else
-    - characters = @user.characters.non_npcs.includes(:template).ordered
-    = render partial_type, name: nil, characters: characters, show_template: true
+    - characters = @user.characters.ordered
+    - characters = character_split == 'npcs' ? characters.npcs : characters.non_npcs
```
Throne3d, line 89: Seems like we've removed the `.includes` here and we're no longer showing templates when there's no character_split - is that intentional?

Throne3d (approved):

Throne3d:
> This does not affect expandable galleries (I don't have better solution than having it just show an error and say to go to the gallery page) or people with character groups (terrifying).

Honestly so legit for character groups

For expandable galleries I think this is fair - maybe in the future we could just load the first N and have a "show more" button redirecting to the full paginated version?

app/views/characters/index.haml
teceler: Ah, I missed that was there.

Throne3d (approved):

app/views/boards/show.haml
Throne3d: (yes to the last question - https://discord.com/channels/521407923629457428/521411396613701662/1273042397470658746)

Throne3d:
Merging some other stuff, then we'll rebase this on top and deploy it to staging to validate the UI

PR #2212, opened 2024-09-30
Fix pagination and notifications

teceler:
- Adds a paginator on template view
- Renames the pluck on characters icon view to avoid the pagination getting overwritten (and making the footer not display)
- Updates some uses of 'messages' around notifications
- Adds a 'none yet' if there are no notifications
- Fixes the notifications header


Marri (approved):
:turtle:

PR #2255, opened 2024-11-09
Use favorite type for notifications

teceler:
Depends on #2213

So currently we always display that a favorited author has created a new post, even if the favorite is of a continuity. This is to fix that.


app/helpers/notification_helper.rb
```diff
@@ -1,13 +1,22 @@
...
-    'import_fail'          => 'Post import failed',
-    'new_favorite_post'    => 'An author you favorited has written a new post',
-    'joined_favorite_post' => 'An author you favorited has joined a post',
+    nil     => {
+      'import_success'       => 'Post import succeeded',
+      'import_fail'          => 'Post import failed',
+      'new_favorite_post'    => 'A once-favorited subject has a new post',
+      'joined_favorite_post' => 'A once-favorited author has joined a post',
+    },
+    'user'  => {
```
Throne3d, line 10:
```suggestion
    'User'  => {
```

app/helpers/notification_helper.rb
```diff
@@ -1,13 +1,22 @@
...
+      'new_favorite_post'    => 'An author you favorited has written a new post',
+      'joined_favorite_post' => 'An author you favorited has joined a post',
+    },
+    'board' => {
```
Throne3d, line 14:
```suggestion
    'Board' => {
```

app/helpers/notification_helper.rb
```diff
@@ -1,13 +1,22 @@
...
+      'new_favorite_post' => 'An continuity you favorited has a new post',
+    },
   }
 
-  def subject_for_type(notification_type)
-    NOTIFICATION_MESSAGES[notification_type]
+  def subject_for_type(notification_type, favorite_type)
+    NOTIFICATION_MESSAGES[favorite_type][notification_type]
```
Throne3d, line 20: Or maybe

```suggestion
    NOTIFICATION_MESSAGES[favorite_type.lower][notification_type]
```
?

PR #2278, opened 2024-11-24
Bump Rails defaults to 7.1

Throne3d:
Recommended once we're stably on an upgraded version.


Throne3d:
Seems like this stops validating the existence of post boards if the value isn't changed:

```rb
let(:post) { create(:post, authors_locked: false) }
...
puts "-A board:#{post.board.inspect}, board_id:#{post.board_id.inspect}, valid:#{post.valid?}, errors:#{post.errors.full_messages}"
post.board_id = 0
puts "-B board:#{post.board.inspect}, board_id:#{post.board_id.inspect}, valid:#{post.valid?}, errors:#{post.errors.full_messages}"
post.update_columns(board_id: 0) # rubocop:disable Rails/SkipsModelValidations
puts "-C board:#{post.board.inspect}, board_id:#{post.board_id.inspect}, valid:#{post.valid?}, errors:#{post.errors.full_messages}"
post.reload
puts "-D board:#{post.board.inspect}, board_id:#{post.board_id.inspect}, valid:#{post.valid?}, errors:#{post.errors.full_messages}"
```
out:
```
-A board:#<Board id: 114, name: "test board 00005", creator_id: 57, created_at: "2024-11-24 15:46:34.611547110 -0500", updated_at: "2024-11-24 15:46:34.611547110 -0500", description: nil, pinned: false, authors_locked: false>, board_id:114, valid:true
# then an exception that `undefined method 'open_to?' for nil` in `post#valid_board`
# using `return if board.nil? || board.open_to?(user)` instead we get:
-A board:#<Board id: 132, name: "test board 00005", creator_id: 66, created_at: "2024-11-24 15:49:16.221155977 -0500", updated_at: "2024-11-24 15:49:16.221155977 -0500", description: nil, pinned: false, authors_locked: false>, board_id:132, valid:true, errors:[]
-B board:nil, board_id:0, valid:false, errors:["Board must exist"]
-C board:nil, board_id:0, valid:true, errors:[]
-D board:nil, board_id:0, valid:true, errors:[]
```

Updating the columns in the database (`post.update_columns` makes it stop performing the "Board must exist" validation.

One of the tests in question (`PostsController PUT update author lock locking handles unexpected failure`) is using this to make the post invalid in the database so a save to an unrelated field will fail. We could look at handling this another way: this is an interesting behavior change, but not one that should impact us too much, since it relies on a field already being invalid in the database.

PR #2287, opened 2024-11-28
Use Strict Locals

teceler:
DO NOT MERGE THIS WITHOUT FURTHER TESTING

### DIFF
--- a/app/views/admin/characters/_relocate.haml
+++ b/app/views/admin/characters/_relocate.haml
@@ -1,3 +1,5 @@
+-# locals: ()
+
 = form_with url: do_relocate_characters_path, method: :post do |f|
   %table.form-table
     %thead
--- a/app/views/boards/_reorder_posts.haml
+++ b/app/views/boards/_reorder_posts.haml
@@ -1,6 +1,4 @@
--# locals: ( unsectioned_posts:, id: :id, klass: '' )
--# TODO: klass is not used but is passed from indexes#edit
->>>>>>> 573889077 (Use default on boards/_reorder_posts)
+-# locals: ( unsectioned_posts:, id: :id )
 
 - reset_cycle
 #reorder-posts-table
--- a/app/views/boards/_reorder_posts.haml
+++ b/app/views/boards/_reorder_posts.haml
@@ -1,3 +1,7 @@
+-# locals: ( unsectioned_posts:, id: :id, klass: '' )
+-# TODO: klass is not used but is passed from indexes#edit
+>>>>>>> 573889077 (Use default on boards/_reorder_posts)
+
 - reset_cycle
 #reorder-posts-table
   .content-header
@@ -12,7 +16,7 @@
       Saved
   %ul.sortable.table-list
     - unsectioned_posts.each do |post|
-      %li.section-ordered{class: cycle('even', 'odd'), data: { id: post.send(local_assigns[:id] || :id), order: post.section_order }}
+      %li.section-ordered{class: cycle('even', 'odd'), data: { id: post.send(id), order: post.section_order }}
         .section-ordered-handle
           = image_tag "icons/arrow_double.png", title: 'Reorder', class: 'disabled-arrow', alt: '↕'
         = link_to post.subject, post_path(post)
--- a/app/views/characters/_icon_view.haml
+++ b/app/views/characters/_icon_view.haml
@@ -1,12 +1,15 @@
-- characters = name.characters.non_npcs.ordered unless local_assigns[:characters]
-- characters = characters.not_retired unless local_assigns.fetch(:show_retired) { show_retired }
+-# locals: ( name: nil, characters: nil, show_template: false, show_user: false, hide_buttons: false, show_new_character_button: false, show_retired: show_retired() )
+-# show_template, show_user, and hide_buttons are to match list_section
+
+- characters ||= name.characters.non_npcs.ordered
+- characters = characters.not_retired unless show_retired
 
 - if @template.nil? && name.present?
   %tr
     %td.subber.padding-10{colspan: 6}
       - if name.is_a? String
         %b= name
-        - if local_assigns[:show_new_character_button]
+        - if show_new_character_button
           = link_to new_character_path do
             .link-box.action-new{style: 'font-size: 14px'} + New Character
       - else
--- a/app/views/characters/_list_section.haml
+++ b/app/views/characters/_list_section.haml
@@ -1,14 +1,16 @@
+-# locals: ( name: nil, characters: nil, show_template: false, show_user: false, hide_buttons: false, show_new_character_button: false, show_retired: show_retired() )
+
 - col_count = 7
-- col_count += 1 if local_assigns[:show_template]
-- characters = name.characters.non_npcs.ordered unless local_assigns[:characters]
-- characters = characters.not_retired unless local_assigns.fetch(:show_retired) { show_retired }
+- col_count += 1 if show_template
+- characters ||= name.characters.non_npcs.ordered
+- characters = characters.not_retired unless show_retired
 
 - if @template.nil? && name.present?
   %tr
     %td.subber.character-list-template{colspan: col_count}
       - if name.is_a? String
         %b= name
-        - if local_assigns[:show_new_character_button]
+        - if show_new_character_button
           = link_to new_character_path do
             .link-box.action-new{style: 'font-size: 14px'} + New Character
       - else
@@ -38,7 +40,7 @@
     - klass = cycle('even', 'odd')
     %td.padding-5{class: klass}
       %b Name
-    - if local_assigns[:show_template]
+    - if show_template
       %td.padding-5{class: klass}
         %b Template
     %td.padding-5{class: klass}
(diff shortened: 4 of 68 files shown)


app/views/boards/_reorder_posts.haml
```diff
@@ -12,7 +15,7 @@
       Saved
   %ul.sortable.table-list
     - unsectioned_posts.each do |post|
-      %li.section-ordered{class: cycle('even', 'odd'), data: { id: post.send(local_assigns[:id] || :id), order: post.section_order }}
+      %li.section-ordered{class: cycle('even', 'odd'), data: { id: post.send(id || :id), order: post.section_order }}
```
Throne3d, line 18: (could probably combine this into `id: :id` above instead?)

teceler: huh, does that work? i'm kind of confused on what this is doing tbh

app/views/admin/characters/_relocate.haml
```diff
@@ -1,3 +1,5 @@
+-# locals: ()
```
Throne3d, line 1: I think we're missing `views/about/_accept_tos.haml`

app/views/boards/_reorder_posts.haml
```diff
@@ -1,3 +1,6 @@
+-# locals: ( unsectioned_posts:, id: nil, klass: '' )
+-# TODO: klass is not used but is passed from indexes#edit
```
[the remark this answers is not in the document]
Throne3d: Looks like `klass` is only passed in by `indexes/edit.haml` (like you said), but it's IndexPost instead of a string / the usual cycle logic. I think it's a mistake, we can get rid of it.

app/views/boards/_reorder_posts.haml
Throne3d: It should set `:id` as the default value for the `id` field, which is what this was doing before - `local_assigns[:id]` if it's not falsy (nil/false), otherwise `:id`

Assuming nothing's explicitly passing in `id: nil`, but instead `id: :some_other_value`, it should be equivalent

app/views/characters/_icon_view.haml
```diff
@@ -1,12 +1,15 @@
-- characters = name.characters.non_npcs.ordered unless local_assigns[:characters]
-- characters = characters.where(retired: false) unless local_assigns.fetch(:show_retired) { show_retired }
+-# locals: ( name: nil, characters: nil, show_template: false, show_user: false, hide_buttons: false, show_new_character_button: false, show_retired: show_retired() )
```
Throne3d, line 1: Huh, I didn't know `show_retired: show_retired` would work. Does it need the parentheses?

app/views/characters/_icon_view.haml
```diff
@@ -1,12 +1,15 @@
...
+-# show_template, show_user, and hide_buttons are to match list_section
```
Throne3d, line 2: (does it need to match list_section? I think we pass in different parameters and it doesn't make sense to have those fields for the icon view given that we don't have those columns)

app/views/characters/_list_section.haml
```diff
@@ -50,14 +52,14 @@
...
         %b User
       - else
         %b Setting
     %td.padding-5{class: klass}
-  - pluck = characters_list(characters.ordered, local_assigns[:show_template])
-  - params = { show_template: local_assigns[:show_template], show_user: local_assigns[:show_user], hide_buttons: local_assigns[:hide_buttons] }
-  - params[:settings] = settings_info(characters) unless local_assigns[:show_user]
+  - pluck = characters_list(characters.ordered, show_template)
+  - params = { show_template: show_template, show_user: show_user, hide_buttons: hide_buttons }
+  - params[:settings] = show_user ? [] : settings_info(characters)
```
Throne3d, line 62:
```suggestion
  - params[:settings] = show_user ? {} : settings_info(characters)
```
(settings_info returns a hash, so this is probably clearer)

app/views/galleries/_icon_item.haml
```diff
@@ -1,3 +1,5 @@
+-# locals: ( icon_item:, is_owner: false, skip_forms: false )
```
Throne3d, line 1: Does it make sense to have defaults in all these fields? I think these are obligatory in the parent

app/views/galleries/_list_item.haml
```diff
@@ -1,3 +1,5 @@
+-# locals: ( list_item:, index: nil, is_owner: false, skip_forms: false )
```
Throne3d, line 1: I didn't find `index: nil` here - is there something passing this in?

app/views/galleries/_single.haml
```diff
@@ -1,14 +1,17 @@
+-# locals: ( gallery:, icons: nil, character_gallery: nil, is_owner: false, skip_forms: false, hide_minmax: false, klass: '' )
```
Throne3d, line 1:
```suggestion
-# locals: ( gallery:, icons: nil, character_gallery: nil, is_owner: false, skip_forms: false, hide_minmax: false, klass: nil )
```
This way we don't change the value we pass to `class:` below, right?

app/views/global/_replace.haml
```diff
@@ -1,3 +1,6 @@
+-# locals: ( obj:, name:, icon:, alt_icon:, is_character: false )
+-# TODO: is_character is not used
```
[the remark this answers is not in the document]
Throne3d: (I'd get rid of is_character it's only used in `characters/replace.haml`, right?)

app/views/posts/_breadcrumbs.haml
```diff
@@ -1,3 +1,5 @@
+-# locals: ( post:, subtitle: '' )
```
Throne3d, line 1: Any reason to prefer `''` and `.blank?` as the default instead of `nil` and checking `.nil?`?

app/views/posts/_list.haml
```diff
@@ -1,12 +1,14 @@
+-# locals: ( posts:, table_class: '', hide_continuity: false, show_unread_count: false, check_box_name: false )
```
Throne3d, line 1:
```suggestion
-# locals: ( posts:, table_class: '', hide_continuity: false, show_unread_count: false, check_box_name: nil )
```
(Seems better to combine nil and string than false and string - this field is a string in the list_item it's passed into)

app/views/posts/_list_item.haml
```diff
@@ -1,5 +1,6 @@
+-# locals: ( post:, index: nil, hide_continuity: false, show_unread_count: false, check_box_name: false, check_box_padding: false, klass: cycle('even', 'odd') )
```
Throne3d, line 1:
```suggestion
-# locals: ( post:, index: nil, hide_continuity: false, show_unread_count: false, check_box_name: nil, check_box_padding: false, klass: cycle('even', 'odd') )
```

app/views/replies/_single.haml
```diff
@@ -1,3 +1,6 @@
+-# locals: ( reply:, hide_footer: false, show_edit: false )
+-# TODO: show_edit is not used but is called from posts#preview -- possibly hide_footer?
```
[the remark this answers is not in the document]
Throne3d: I'd switch it over to hide_footer! Sounds like what was intended

app/views/templates/_editor.haml
```diff
@@ -1,3 +1,5 @@
+-# locals: ( f: )
```
Throne3d, line 1: I think we're missing `writable/_history.haml` too

app/views/boards/_reorder_posts.haml
teceler: oh well then we probably just want the default to be -- ah, :id, yes

app/views/characters/_icon_view.haml
teceler: We do a lot of passing things to partial_type and sometimes those things -- well, i ran into it with show_template but easier just to duplicate them all

app/views/characters/_icon_view.haml
Throne3d: Ooh, ok!

app/views/galleries/_single.haml
teceler: Whoops, I thought I caught all of those

app/views/posts/_breadcrumbs.haml
teceler: Because I was trying to match types at some point in the past xD

app/views/replies/_single.haml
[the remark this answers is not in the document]
teceler: This will make posts and reply preview stop showing the footer (timestamps), which is a functionality change I want to explicitly flag. You're probably right tho.

app/views/galleries/_icon_item.haml
teceler: They have defaults in `_icon_view`

app/views/galleries/_single.haml
teceler: I don't think it makes a difference in this case actually! Because we only use it there and don't further manipulate it

Throne3d (approved):

Throne3d:
Performed some further testing locally! Also corroborated the values by comparing notes when I did the same checks locally. Most pages have render_views enabled. I think we're good to push forward with this and fix any obvious errors that show up, so I'll look at deploying it, since it's a meaningful improvement to development workflows.

PR #2291, opened 2024-11-30
Fix template "All Characters" button with default hidden retired

redmvc:
When your account defaults to hiding retired characters, we need to specify explicitly `retired=true` in the URL to get retired characters to show up. This already worked on the characters page, but not on `template#show`.

### DIFF
--- a/app/views/templates/show.haml
+++ b/app/views/templates/show.haml
@@ -36,7 +36,7 @@
             .view-button
               Hide Retired
         - else
-          = link_to character_menu_link(retired: nil), class: 'view-button-link' do
+          = link_to character_menu_link(retired: true), class: 'view-button-link' do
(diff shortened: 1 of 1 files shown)


app/views/templates/show.haml
```diff
@@ -36,7 +36,7 @@
             .view-button
               Hide Retired
         - else
-          = link_to character_menu_link(retired: nil), class: 'view-button-link' do
+          = link_to character_menu_link(retired: true), class: 'view-button-link' do
```
Throne3d, line 39: Same logic as https://github.com/glowfic-constellation/glowfic/blob/main/app/views/characters/index.haml#L55, sounds good!

Throne3d (approved):

PR #2292, opened 2024-12-01
Add option to retire templates

redmvc:
Add an option to mark an entire template as retired, which causes it to not show up on the characters list if retired characters aren't shown and causes all of its members to not be selectable when writing posts or replies.


teceler (approved):
Looks good!

PR #2298, opened 2024-12-02
Add UI for thread splitting

redmvc:
Create UI for regular users to split their threads in two. The tool is found in the hamburger menu and changes visualisation to "split mode" which allows a user to choose a reply from which to split the thread and redirects them to a page that allows them to choose the new thread's subject and perform the split.


app/controllers/posts_controller.rb
```diff
@@ -321,6 +322,32 @@ def warnings
     redirect_to post_path(@post, params)
   end
 
+  def split
+    @page_title = 'Split Post'
+  end
+
+  def do_split
```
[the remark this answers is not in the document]
redmvc: This code already existed in the base and was just moved.

Throne3d, line 329: Comparison - old requirements:
1\. Reply existed, was part of this post
2\. Subject given
3\. User has `split_posts` permission

New requirements (instead of 3. above):
3\. User can edit the post (either itself or metadata, same as `/posts/x/edit`)
4\. Post is author-locked

app/jobs/split_post_job.rb
```diff
@@ -31,6 +32,9 @@ def create_post(first_reply, old_post:, subject:)
     new_post.skip_edited = true
     new_post.is_import = true
     new_post.assign_attributes(old_post.attributes.slice(*POST_ATTRS))
+    POST_ASSOCS.each do |assoc|
```
Throne3d, line 35: ❤️

app/controllers/posts_controller.rb
teceler: Hm, we might want an 'or user has split_posts permission'?

app/views/posts/show.haml
```diff
@@ -61,7 +69,7 @@
             = image_tag "icons/star_delete.png".freeze, class: 'vmid'.freeze, style: 'margin-bottom: 3px;'.freeze, alt: ''
             Unfavorite
       - else
-        = link_to favorites_path(post_id: @post.id, page: page, per_page: per_page), method: :post do
+        = link_to favorites_path(post_id: @post.id), method: :post do
```
Throne3d, line 72: Drive-by fix, cool

app/controllers/posts_controller.rb
redmvc: So that a mod/admin can do it? I'm not opposed if you guys think it makes sense but I don't particularly see the need.

app/views/replies/_single.haml
```diff
@@ -32,17 +32,25 @@
...
-          = link_to path_for(reply, 'edit_%s') do
-            = image_tag "icons/pencil.png".freeze, title: 'Edit'.freeze, alt: 'Edit'.freeze
-        - if reply.deletable_by?(current_user)
-          = link_to path_for(reply, '%s'), method: :delete, data: { confirm: "Are you sure you wish to delete this #{reply.class.to_s.downcase}?" } do
-            = image_tag "icons/cross.png".freeze, title: 'Delete'.freeze, alt: 'Delete'.freeze
-        - if reply.is_a?(Reply) && logged_in?
-          = link_to post_path(@post, unread: true, at_id: reply.id), method: :put do
-            = image_tag "icons/eye.png".freeze, title: 'Mark Unread Here'.freeze, alt: 'Mark Unread Here'.freeze
+        - can_be_split = reply.is_a?(Reply) && @post.metadata_editable_by?(current_user) # Logic for splitting a thread at a reply
+        - if params[:split_ui] == 'true' && can_be_split
```
Throne3d, line 36: This feels like a lot of logic embedded inside here and spread between different parts of the partial.

Can we set a variable up top in posts#show for whether or not the split_ui is enabled (and accessible by the user), then we just rely on that being passed to the partial here for consistency? That way we can make sure the UI enable/disable button, and hiding the reply writer, and adjusting the buttons, is all consistently set up.

spec/controllers/posts/split_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+    it "requires edit permissions" do
+      login
+      get :split, params: { id: user_post.id }
+      expect(response).to redirect_to(post_url(user_post))
+      expect(flash[:error]).to eq("You do not have permission to modify this post.")
+    end
+
+    it "requires locked authorship" do
+      login_as(user)
+      put :update, params: { id: user_post.id, authors_locked: 'false' }
```
Throne3d, line 22: Usually controller specs only do one request - I'd do `user_post.update(authors_locked: false)` here instead of submitting another request.

Alternatively, we could put this in `spec/requests/` and test the full flow from a `cURL` perspective? =)

app/controllers/posts_controller.rb
Throne3d: Yeah, I think we discussed this and decided it shouldn't be necessary with end users having access. Possible, but not necessary

app/views/posts/show.haml
```diff
@@ -127,15 +146,15 @@
...
 - else
   .post-expander
     .info + Show First Post
     .hidden
-      = render 'replies/single', reply: @post
+      = render 'replies/single', reply: @post, split_ui: split_ui
 - if @replies.present?
   = render 'posts/paginator', paginated: @replies
-= render partial: 'replies/single', collection: @replies, as: :reply
+= render partial: 'replies/single', collection: @replies, as: :reply, locals: {split_ui: split_ui}
```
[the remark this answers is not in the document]
Throne3d:
```suggestion
= render partial: 'replies/single', collection: @replies, as: :reply, locals: { split_ui: split_ui }
```

spec/controllers/posts/split_spec.rb
redmvc: I did not realise `user_post.update(authors_locked: false)` was a thing, can do

Throne3d (approved):
LGTM assuming it passes tests!

app/controllers/posts_controller.rb
redmvc: @teceler is that alright?

teceler: Yeah, I can't think of many cases where we'd need that, unless we got called on to split a megacont post or something, and even then  you can probably just add more steps

PR #2308, opened 2024-12-10
Bump Rails to 7.2 in README

Throne3d:


Throne3d:
docs-only, doesn't need to pass tests

PR #2311, opened 2024-12-12
Fix psql command

redmvc:
The PSQL command wasn't running correctly to access the database (was an old version or sth?)


Throne3d:
I think we just never used `psql` before tbh, and replicated the command in a similar format to other commands without considering the specific differences here!

LGTM

PR #2315, opened 2024-12-12
Add author content warnings

redmvc:
Add the possibility for authors to create default content warnings that may apply to any post of theirs, whether the post is individually warned or not.


app/views/users/_list.haml
```diff
@@ -0,0 +1,11 @@
+%tbody
```
Throne3d, line 1: Do our `/_list.haml`s usually not contain the header? (If they don't, interesting! I thought they did.)

Throne3d: Btw, can we pass in `users` as a local variable instead of using `@users`? Would be nice to start relying on strict locals more for our partials!

app/views/users/_list.haml
```diff
@@ -0,0 +1,11 @@
...
+  - @users.each do |user|
```
Throne3d, line 2: Can we pass in `users` as a local variable instead of using `@users`? Would be nice to start relying on strict locals more for our partials!

db/migrate/20241209191600_add_user_tags.rb
```diff
@@ -0,0 +1,11 @@
+class AddUserTags < ActiveRecord::Migration[7.2]
+  def change
+    create_table :user_tags do |t|
+      t.integer :user_id, null: false
+      t.integer :tag_id, null: false
+      t.timestamps null: true
```
Throne3d, line 6: Huh, do timestamps need to be nullable? (It seems like we should specify this, which makes sense, but https://guides.rubyonrails.org/active_record_migrations.html#migration-overview suggests that `null: false` should work for timestamps too, and seems preferable)

spec/models/tag_spec.rb
```diff
@@ -20,6 +20,31 @@
...
+      user1 = create(:user)
+      user1.update!(content_warning_ids: [good_tag.id])
+      user2 = create(:user)
+      user2.update!(content_warning_ids: [good_tag.id])
+      user3 = create(:user)
+      user3.update!(content_warning_ids: [good_tag.id])
+      user4 = create(:user)
+      user4.update!(content_warning_ids: [bad_tag.id])
+      user5 = create(:user)
+      user5.update!(content_warning_ids: [bad_tag.id])
```
Throne3d, line 37: I think you could write this:

```suggestion
      user1, user2, user3 = create_list(:user, 3, content_warning_ids: [good_tag.id])
      user4, user5 = create_list(:user, 2, content_warning_ids: [bad_tag.id])
```

spec/system/users/edit_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    # TODO all fields
+    # within("#edit_user_#{user.id}") do
+    #   expect(page).to have_field('Username', with: 'John Doe')
+    #   expect(page).to have_field('Email address', with: 'dummy@example.com')
+    #   fill_in 'Username', with: 'Jane Doe'
+    #   fill_in "Email address", with: "dummy2@example.com"
+    # end
+
+    expect(page).to have_no_selector('.select2-selection__choice')
```
Throne3d, line 31: Wow, you're right, this is horrible :D 

I'll see if we can do some better helpers for this kind of selection logic

app/views/users/_list.haml
redmvc: Some partials do, some don't

app/views/users/_list.haml
redmvc: Would kind of prefer not to do that in this PR? Since it's a separate topic

db/migrate/20241209191600_add_user_tags.rb
Throne3d: I think most of our other timestamps _are_ nullable, but they also have `precision: nil` defined on them - my guess is some defaults may have changed in different Rails versions.

Did you use a specific rationale to pick this config? Wondering if I'm missing something

redmvc: No clue, I just c/p the other _tags

Throne3d: Ah ok, I think it's ok then!

spec/system/users/edit_spec.rb
redmvc: :)

app/views/users/_list.haml
Throne3d: Can we include it here if possible? I think it makes sense to consistently include it, where possible

app/views/users/_list.haml
Throne3d: Hm, ok. Clean-up PR next? :D

app/views/users/_list.haml
Throne3d: Awkward to combine because of differing header formats (search bar vs other things); nm!

app/views/users/_list.haml
Throne3d: -> handled

Throne3d (approved):

PR #2319, opened 2024-12-13
Add "Check All" checkboxes to all pages

redmvc:
Extend the functionality of the "select all" checkboxes that were already available on the inbox to cover posts, continuities, and notifications where relevant.


app/assets/javascripts/global.js
```diff
@@ -13,8 +13,14 @@ $(document).ready(function() {
...
 
-  $("#select_all").click(function() {
-    $(".select-all-box").prop('checked', this.checked);
+  $('.check-all').on('change', function() {
+    // The check-all's checkbox value will contain the name of the checkboxes which are meant to be checked
+    $(`.checkbox[name="${this.value}"]`).prop('checked', this.checked);
+  });
+
+  $('.checkbox').on('change', function() {
+    const allChecked = $(`.checkbox[name="${this.name}"]`).filter(':checked').length === $(`.checkbox[name="${this.name}"]`).length;
```
Throne3d, line 22:
```suggestion
    const checkboxes = $(`.checkbox[name="${this.name}"]`)
    const allChecked = checkboxes.filter(':checked').length === checkboxes.length;
```
seems a bit clearer

Throne3d, line 22: Are all `.checkboxes` part of a series for `.check-all`? It seems like it would be safer to scope this JavaScript only to checkboxes that we know are part of check-all lists. It likely doesn't do anything in the case there isn't a corresponding `.check-all`, but I'm thinking if `this.name` might be empty for some checkboxes or if there are some other weird edge cases here.

app/controllers/notifications_controller.rb
```diff
@@ -10,6 +10,8 @@ def index
 
     post_ids = @notifications.map(&:post_id).compact_blank
     @posts = posts_from_relation(Post.where(id: post_ids), with_pagination: false).index_by(&:id)
+
+    use_javascript('global')
```
Throne3d, line 14: Huh, the global javascript doesn't load everywhere? Interesting. It's been a while since I touched how we load the javascript but this surprises me

app/assets/javascripts/global.js
```diff
@@ -13,8 +13,14 @@ $(document).ready(function() {
     return false;
   });
 
-  $("#select_all").click(function() {
-    $(".select-all-box").prop('checked', this.checked);
+  $('.check-all').on('change', function() {
+    // The check-all's checkbox value will contain the name of the checkboxes which are meant to be checked
+    $(`.checkbox[name="${this.value}"]`).prop('checked', this.checked);
```
Throne3d, line 18: I wonder if this (when selected) impacts the data we send to the server. Typically input `name`s are the key for the data we send in a form, so does this add an extra `marked_ids[]=true` to the submitted form? IMO this would be safer as a `data` key

app/views/posts/_list.haml
```diff
@@ -28,7 +28,9 @@
         %th.sub Unread
       %th.sub Last Updated
       - if check_box_name
-        %th.sub
+        %th.sub.check-all-box
+          - unless posts.empty?
+            = check_box_tag :check_all, check_box_name, false, class: 'check-all', id: 'check-all'
```
Throne3d, line 33: If there's multiple post lists on a page that have checkboxes, this might cause invalid HTML (by duplicating the ID). I think most post lists don't have checkboxes, so maybe this isn't an issue - but can you think of anywhere that might be impacted?

(Is the ID necessary? It seems like you did some work to avoid needing an ID by specifying the corresponding `name` for each checkbox, so we might just be able to remove `id` from each of these definitions)

app/assets/javascripts/global.js
Throne3d: Oh. This is the _value_, huh. Does this checkbox have a name in the final HTML?

app/assets/javascripts/global.js
redmvc: `this.name` works fine if it's empty and that does not do anything when there isn't a matching `.check-all`

Throne3d: (discussed aloud, seems to be true but would be nice to avoid having to think this through to be sure of the correctness)

Throne3d (approved):

Throne3d (approved):

PR #2321, opened 2024-12-16
Turn off yjit to see memory impact

Marri:
Per the Rails 7.2 settings file:
```
# Enables YJIT as of Ruby 3.3, to bring sizeable performance improvements. If you are
# deploying to a memory constrained environment you may want to set this to `false`.
#++
# Rails.application.config.yjit = true
```

And Heroku is in fact memory constrained, and we did in fact see a memory spike after deploy:

image

So turning off to see if this helps.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -80,5 +80,9 @@ module Glowfic
     # redis-rails does not support cache versioning
     config.active_record.cache_versioning = false
     config.active_record.collection_cache_versioning = false
+
+    # Setting enables YJIT as of Ruby 3.3, to bring sizeable performance improvements. We are
+    # deploying to a memory constrained environment so we set this to `false`.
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -80,5 +80,9 @@ class Application < Rails::Application
     # redis-rails does not support cache versioning
     config.active_record.cache_versioning = false
     config.active_record.collection_cache_versioning = false
+
+    # Setting enables YJIT as of Ruby 3.3, to bring sizeable performance improvements. We are
+    # deploying to a memory constrained environment so we set this to `false`.
+    Rails.application.config.yjit = false
```
Throne3d, line 86:
```suggestion
    config.yjit = false
```
based on the context, probably this?

Throne3d (approved):

PR #2327, opened 2024-12-20
Add user bookmarks

redmvc:
image
image
image
image
image
image
image
image


app/assets/javascripts/bookmarks/rename.js
```diff
@@ -0,0 +1,76 @@
+let originalValues = {};
```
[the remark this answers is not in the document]
redmvc: ...I don't understand why it says this? I totally edit originalValues just below these lines.

app/views/bookmarks/search.haml
```diff
@@ -0,0 +1,113 @@
...
+      %th.search-results-header
+        - if @search_results
+          = render 'posts/paginator', paginated: @search_results, no_per: true
+  %tbody
+    %tr
+      %td.search-box
+        = form_tag search_bookmarks_path, method: :get do
+          %table.search-form
+            %tr
+              %td{ style: 'width: 80px;' }= label_tag :user_id, 'User'
```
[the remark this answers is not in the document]
redmvc: This one is already used in other search pages so I'ma keep it the same.

app/assets/javascripts/posts/search.js
```diff
@@ -37,6 +37,24 @@ $(document).ready(function() {
     allowClear: true,
   });
 
+  createSelect2('#user_id', {
```
Throne3d, line 40: This file is used in all our searches, right?

Would be nice to rename it to just `javascripts/search.js` at some point!

app/controllers/api/v1/replies_controller.rb
```diff
@@ -23,4 +23,30 @@ def index
...
+      render json: { errors: [error] }, status: :forbidden
+      return
+    end
+
+    bookmark = reply.bookmarks.find_by(user_id: user.id, type: "reply_bookmark")
+
+    if bookmark.present?
+      render json: bookmark.as_json
+    else
+      render json: {}
```
Throne3d, line 49: I wonder if this would be better reported as an error or with a different status code. 404? If an end client needs to distinguish between the reply/user not existing and the bookmark not existing, they can parse the errors string.

Alternatively, if a reply can have _multiple_ bookmarks for a specific user (e.g. ones of different types), we could just return an empty array here and convert this into a pluralized endpoint?

app/controllers/api/v1/bookmarks_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+      return
+    end
+
+    unless @bookmark.destroy
+      error = { message: 'Bookmark could not be removed.' }
+      render json: { errors: [error] }, status: :unprocessable_entity
+      return
+    end
+
+    render json: {}
```
Throne3d, line 75:
```suggestion
    render status: :no_content
```
Might be better (assuming that syntax works) - returning 204 is common for successful deletes

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204

app/controllers/api/v1/users_controller.rb
```diff
@@ -29,11 +31,36 @@ def index
...
+    if !@user.public_bookmarks && @user.id != current_user.try(:id)
+      error = { message: "This user's bookmarks are private." }
+      render json: { errors: [error] }, status: :forbidden
+      return
+    end
+
+    bookmarks = @user.bookmarks.visible_to(current_user)
+    bookmarks = bookmarks.where(post_id: params[:post_id]) if params[:post_id].present?
+
+    bookmarks = paginate bookmarks, per_page: 25
```
Throne3d, line 57: Would be a good idea to order these, since we're paginating them! I don't think we have a default order here, do we?

```suggestion
    bookmarks = paginate bookmarks.order(:id), per_page: 25
```

app/controllers/bookmarks_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+
+    @reply = Reply.find_by_id(params[:at_id])
+    unless @reply
+      flash[:error] = "Reply not found."
+      return redirect_to posts_path
+    end
+
+    bookmark = Bookmark.where(reply_id: @reply.id, user_id: current_user.id, post_id: @reply.post.id,
+      type: 'reply_bookmark',).first_or_initialize
```
Throne3d, line 51:
```suggestion
      type: 'reply_bookmark').first_or_initialize
```

app/models/bookmark.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+class Bookmark < ApplicationRecord
+  belongs_to :user, inverse_of: :bookmarks, optional: false
+  belongs_to :reply, inverse_of: :bookmarks, optional: false
+  belongs_to :post, inverse_of: :bookmarks, optional: false
+
+  validates :type, uniqueness: { scope: [:user, :reply] }
+  validates :type, inclusion: { in: ['reply_bookmark'] }, allow_nil: false
```
Throne3d, line 8: Can `type` be an enum? https://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html

(Can still use string columns if desirable!)

app/views/users/show.haml
```diff
@@ -46,6 +46,11 @@
         = link_to search_posts_path(author_id: [@user.id], commit: 'Search', completed: true), rel: 'nofollow' do
           = image_tag "icons/book.png", class: 'vmid', alt: ''
           Completed Posts
+    %tr
+      %td.centered{class: cycle('even', 'odd')}
+        = link_to search_bookmarks_path(user_id: @user.id, commit: 'Search'), rel: 'nofollow' do
```
Throne3d, line 51: Should we only show this if the user's bookmarks are public?

db/migrate/20241220001400_rename_user_bookmarks.rb
```diff
@@ -0,0 +1,5 @@
+class RenameUserBookmarks < ActiveRecord::Migration[7.2]
+  def change
+    rename_table :user_bookmarks, :bookmarks
```
Throne3d, line 3: (Can probably squish this with the above migration)

app/assets/javascripts/posts/search.js
redmvc: Yeah, I can make another small PR for that later

app/controllers/bookmarks_controller.rb
redmvc: Rubocop won't let me

app/models/bookmark.rb
redmvc: Uhhh I guess it could but other enum columns like editor_mode in reply are not and use the inclusion validation

app/controllers/api/v1/replies_controller.rb
redmvc: We can't have multiple bookmarks _yet_ and I think when we do I'll want to add a parameter to specify the type rather than have a list maybe? idk... that's fair though...

app/views/users/show.haml
redmvc: Maybe? I thought about that but decided against it because it felt unaesthetic to me for the user infoboxes themselves to change that way when nothing else on them does.

db/migrate/20241220001400_rename_user_bookmarks.rb
redmvc: Just delete the first one and update this one, or?

app/views/users/show.haml
redmvc: Ppl seem to prefer that private bookmarks be indistinguishable from public-but-zero bookmarks

app/models/bookmark.rb
redmvc: (Discussed aloud: keeping it as-is to maintain coherence with other tables, though we might want to look into it for all columns that are enumerable in the future.)

Throne3d (approved):

PR #2331, opened 2024-12-22
Add bookmark-specific privacy

redmvc:
Add a setting to make individual bookmarks public even when the user has their bookmarks private by default

image

### DIFF
--- a/app/assets/javascripts/bookmarks/rename.js
+++ b/app/assets/javascripts/bookmarks/rename.js
@@ -1,17 +1,22 @@
 const originalValues = {};
+const originalPublicCheckboxes = {};
 
 $(document).ready(function() {
   const textFields = $(`.bookmark-name-text-field`);
   for (const textField of textFields) {
     originalValues[textField.dataset.bookmarkId] = $(textField).val();
   }
+  const checkBoxes = $(`.bookmark-public-checkbox`);
+  for (const checkBox of checkBoxes) {
+    originalPublicCheckboxes[checkBox.dataset.bookmarkId] = $(checkBox).prop("checked");
+  }
 
-  if (!$(".rename-bookmark").length) return;
+  if (!$(".edit-bookmark").length) return;
 
-  $(".rename-bookmark").click(function() {
+  $(".edit-bookmark").click(function() {
     /* Button to rename a bookmark */
     const bookmarkId = this.dataset.bookmarkId;
-    const editors = $(`.bookmark-name-editor`);
+    const editors = $(`.bookmark-editor`);
     for (const editor of editors) {
       // Hide all editors other than the one clicked
       const editorId = editor.dataset.bookmarkId;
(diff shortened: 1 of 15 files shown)


app/assets/javascripts/bookmarks/rename.js
```diff
@@ -1,17 +1,22 @@
-const originalValues = {};
+const originalNameTextboxes = {};
```
Throne3d, line 1: I'm sure there's a better way to do these multiple JavaScript forms (maybe rails-ujs and data-remote?) but it seems like a straightforward extension of the current approach so I don't think it's a blocker!

app/controllers/api/v1/bookmarks_controller.rb
```diff
@@ -1,29 +1,27 @@
...
-      return
-    end
+    access_denied and return unless reply.post.visible_to?(current_user)
 
-    bookmark = Bookmark.where(user: current_user, reply: reply, post: reply.post, type: "reply_bookmark").first_or_initialize
-    bookmark.assign_attributes(name: params[:name])
-    unless bookmark.save
+    bookmark = Bookmark.where(user: current_user, reply: reply, type: "reply_bookmark").first_or_initialize
+    params[:post_id] = reply.post_id
+    unless bookmark.update(params.permit(:name, :public, :post_id))
```
Throne3d, line 24:
```suggestion
    unless bookmark.update(params.permit(:name, :public).merge(post_id: reply.post_id))
```

app/controllers/api/v1/bookmarks_controller.rb
```diff
@@ -77,8 +66,9 @@ def destroy
 
   private
 
-  def find_bookmark
+  def bookmark_ownership_required
     return unless (@bookmark = find_object(Bookmark))
-    access_denied unless @bookmark.visible_to?(current_user)
+    access_denied and return unless @bookmark.visible_to?(current_user)
+    access_denied unless @bookmark.user.id == current_user.try(:id)
```
Throne3d, line 72:
```suggestion
    access_denied unless @bookmark.user.id == current_user.try(:id)
```
This should be enough, right?

app/controllers/bookmarks_controller.rb
```diff
@@ -50,14 +50,10 @@ def create
...
-      type: 'reply_bookmark',).first_or_initialize
+    bookmark = Bookmark.where(reply_id: @reply.id, user_id: current_user.id, type: 'reply_bookmark').first_or_initialize
     if bookmark.new_record?
-      if params[:bookmark_name].present?
-        bookmark.update!(name: params[:bookmark_name])
-      else
-        bookmark.save!
-      end
+      params[:post_id] = @reply.post_id
+      bookmark.update!(params.permit(:name, :public, :post_id))
```
Throne3d, line 56: (same as before, or even just `bookmark.post_id = @reply.post_id` before the update call: it combines unsaved changes with the ones you pass)

app/controllers/api/v1/bookmarks_controller.rb
redmvc: No bc a post that has a replied bookmark may have become private since the bookmark was added and so the user can't access the bookmark even if they're the owner

Throne3d: discussed aloud, kinda awkward but seems legit

PR #2336, opened 2024-12-23
Turn the labels of the profile form into actual labels

redmvc:


Throne3d:
Fixes the text alignment, which is otherwise a bit weird

image

PR #2344, opened 2024-12-25
Add settings to hide reply buttons

redmvc:
If at least one of the settings is enabled, the "Show Hidden Reply Buttons" appears on the thread menu. Clicking it will reveal the hidden buttons, then clicking "Hide Reply Buttons" will hide the ones set by the user options again.

image
image
image
image
image


app/views/users/edit.haml
```diff
@@ -2,8 +2,10 @@
   %table.form-table
     %thead
       %tr
-        %th.editor-title{colspan: 2} Account Settings
+        %th.editor-title{colspan: 4} Settings
```
Throne3d, line 5: (colspan change seems to be a mistake introduced from another PR)

PR #2366, opened 2024-12-31
Fix sign-up password confirm field validations

Throne3d:


teceler (approved):
:green_heart: :turkey:

PR #2375, opened 2025-01-03
Make editor show the correct user's characters on preview

redmvc:
When using the "preview" feature, the characters and NPCs that would show up to be selected would be the top poster's, not your own.

image

### DIFF
--- a/app/controllers/writable_controller.rb
+++ b/app/controllers/writable_controller.rb
@@ -5,7 +5,7 @@ class WritableController < ApplicationController
   def build_template_groups(user=nil)
     return unless logged_in?
     user ||= current_user
-    user = current_user if user.id != current_user.try(:id) && !current_user.mod? && !current_user.admin?
+    user = current_user if user.id != current_user.id && !current_user.mod? && !current_user.admin?
 
     faked = Struct.new(:name, :id, :plucked_characters)
     faked_npcs = Struct.new(:name, :id, :plucked_npcs)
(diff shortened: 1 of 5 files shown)


app/controllers/writable_controller.rb
```diff
@@ -5,6 +5,7 @@ class WritableController < ApplicationController
   def build_template_groups(user=nil)
     return unless logged_in?
     user ||= current_user
+    user = current_user if user.id != current_user.try(:id) && !current_user.mod? && !current_user.admin?
```
teceler, line 8: You can't get here if you're not logged in, so you don't need the try

redmvc: Oh tru

app/controllers/writable_controller.rb
```diff
@@ -5,6 +5,7 @@ class WritableController < ApplicationController
   def build_template_groups(user=nil)
     return unless logged_in?
     user ||= current_user
+    user = current_user if user.id != current_user.id && !current_user.mod? && !current_user.admin?
```
Throne3d, line 8:
```suggestion
    user = current_user unless user.id == current_user.id || current_user.mod? || current_user.admin?
```

Throne3d:
```suggestion
    user = current_user unless user.id == current_user.id || current_user.has_permission?(:edit_posts)
```

PR #2378, opened 2025-01-03
Add multi-reply feature

redmvc:
This feature allows the user to post multiple replies all at once, or to edit an existing reply and turn it into multiple replies in-place (to, effectively, "split" it).

image
image
image
image
image
image
image
image
image
image


app/controllers/replies_controller.rb
```diff
@@ -297,19 +300,147 @@ def require_edit_permission
...
+    begin
+      Reply.transaction do
+        if @multi_replies.present?
+          # Add new replies after the current one and before the next
+
+          # Reorder the replies after this one
+          original_reply_order = @reply.order
+          num_new_replies = @multi_replies.length
+          following_replies = @reply.post.replies.where("reply_order > ?", original_reply_order)
+          following_replies.update_all("reply_order = reply_order + #{num_new_replies}") # rubocop:disable Rails/SkipsModelValidations
```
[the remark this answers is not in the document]
redmvc: Is there a better way to do it?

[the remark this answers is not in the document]
Throne3d:
```suggestion
          following_replies.update_all(Arel.sql("reply_order = reply_order + ?", num_new_replies)) # rubocop:disable Rails/SkipsModelValidations
```

https://api.rubyonrails.org/classes/Arel.html

(nb. using `?` to avoid injection)

app/controllers/replies_controller.rb
```diff
@@ -116,7 +117,20 @@ def create
...
+    elsif params[:button_discard_multi_reply]
+      flash[:success] = "Replies discarded."
+      if @multi_replies_json.present? && (editing_reply_id = @multi_replies_json.first["id"]).present?
+        # Editing multi reply, going to redirect back to the reply I'm editing
+        redirect_to reply_path(editing_reply_id, anchor: "reply-#{editing_reply_id}")
+      else
+        # Posting a new multi reply, go back to unread
+        redirect_to post_path(params[:reply][:post_id], page: :unread, anchor: :unread)
+      end
+      return
```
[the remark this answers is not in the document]
redmvc: I tried reducing them as much as I could but I don't think I can do better than what I currently have.

redmvc:
I have a couple of lines in `replies_controller.rb` which aren't covered by the tests but I'm not sure how to cover them or what to replace them with.

image
image
image

These are all checks of the "this shouldn't ever happen but _if it does_ I wanna catch it" variety, and given that they shouldn't ever happen I'm really not sure how to test them without just straight-up forcing an error into the code itself.

app/controllers/replies_controller.rb
```diff
@@ -297,19 +300,147 @@ def require_edit_permission
...
     @written.user = current_user unless @written.user
     @audits = @written.id.present? ? { @written.id => @written.audits.count } : {}
+    @reply = @written # So that editor_setup shows the correct characters
+    preview_replies
+  end
+
+  def add_to_multi_reply(reply, reply_params)
+    # Adding a new reply to a multi reply
+    post_id = params[:reply][:post_id]
+    ReplyDraft.draft_for(post_id, current_user.id).try(:destroy)
```
Throne3d, line 324:
```suggestion
    ReplyDraft.draft_for(post_id, current_user.id)&.destroy
```

app/controllers/replies_controller.rb
```diff
@@ -297,19 +300,147 @@ def require_edit_permission
...
     editor_setup
     render :preview
   end
 
+  def post_replies(new_reply: nil)
+    @multi_replies << new_reply if new_reply.present?
+
+    if @multi_replies_json.present? && (@reply = Reply.find_by_id(@multi_replies_json.first["id"]))
+      # The first reply of the multi replies has an ID, that means I'm editing rather than posting a new one
+      original_reply_params = permitted_params(ActionController::Parameters.new({ reply: @reply.attributes }))
```
Throne3d, line 367: Use `@reply.dup`? Then you can pass a `Reply` object around instead of doing a `permitted_params(... @reply.attributes ...)` mutation

app/controllers/replies_controller.rb
```diff
@@ -297,19 +300,147 @@ def require_edit_permission
...
+      render :edit
+    else
+      flash[:success] = "Reply updated."
+      redirect_to reply_path(@reply, anchor: "reply-#{@reply.id}")
+    end
+  end
+
+  def edit_multi_replies(original_reply_params)
+    # Add new replies after the current one and before the next
+    if original_reply_params.blank?
```
Throne3d, line 418: Can we avoid checking if it's blank and just pass in `@reply` reliably in the code paths that need it?

app/controllers/replies_controller.rb
```diff
@@ -297,19 +300,147 @@ def require_edit_permission
     redirect_to post_path(@reply.post)
   end
 
-  def preview(written)
-    @written = written
-    @post = @written.post
+  def get_multi_replies
+    @multi_replies_json = JSON.parse(params.fetch(:multi_replies_json, "[]"))
```
Throne3d, line 304: Maybe we can sanitize `@multi_replies_json` here (put it through `permitted_params`, separate before doing the `Reply.new`) and then assume in subsequent sections that it's safe

app/controllers/replies_controller.rb
```diff
@@ -116,7 +117,24 @@ def create
...
+    elsif params[:button_submit_previewed_multi_reply]
+      if editing_multi_reply?
+        edit_reply(true)
+      else
+        post_replies
+      end
+      return
+    elsif params[:button_discard_multi_reply]
+      flash[:success] = "Replies discarded."
+      if editing_multi_reply?
```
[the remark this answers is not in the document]
redmvc: My brother in Christ if you have a better suggestion I am all ears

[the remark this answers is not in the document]
teceler: Make reply#create a service? frankly the entire problem here is is this is deeply nested and *way too much in a single method*

Out of scope tho

app/controllers/replies_controller.rb
```diff
@@ -116,7 +117,24 @@ def create
...
+        # Editing multi reply, going to redirect back to the reply I'm editing
+        redirect_to reply_path(@reply, anchor: "reply-#{@reply.id}")
+      else
+        # Posting a new multi reply, go back to unread
+        redirect_to post_path(params[:reply][:post_id], page: :unread, anchor: :unread)
```
teceler, line 135: Wait why do you have this on the discard path

app/controllers/replies_controller.rb
```diff
@@ -139,10 +157,10 @@ def create
           flash.now[:error] = "This looks like a duplicate. Did you attempt to post this twice? Please resubmit if this was intentional."
           @allow_dupe = true
           if most_recent_unseen_reply.nil? || (most_recent_unseen_reply.id == last_by_user.id && @unseen_replies.count == 1)
-            preview(reply)
+            preview_reply(reply)
```
teceler, line 160: Why the rename?

app/controllers/replies_controller.rb
```diff
@@ -187,26 +202,20 @@ def edit
...
-      render_errors(@reply, action: 'updated', now: true, err: e)
-
-      @audits = { @reply.id => @post.audits.count }
-      editor_setup
-      render :edit
-    else
-      flash[:success] = "Reply updated."
-      redirect_to reply_path(@reply, anchor: "reply-#{@reply.id}")
-    end
+    edit_reply(false)
```
teceler, line 218: ~~like i said this needs a service~~

app/controllers/replies_controller.rb
```diff
@@ -297,19 +306,153 @@ def require_edit_permission
...
-    @post = @written.post
+  def get_multi_replies
+    # Get the multi replies stored in the page JSON
+    @multi_replies_params = JSON.parse(params.fetch(:multi_replies_json, "[]")).map do |reply_json|
+      permitted_params(ActionController::Parameters.new({ reply: reply_json }), [:id])
+    end
+    @multi_replies = @multi_replies_params.map do |reply_params|
+      Reply.new(reply_params).tap { |r| r.user = current_user }
+    end
+  end
```
teceler, line 317: ...it feels like there should be a more elegant way to do this, is there not?

app/controllers/replies_controller.rb
```diff
@@ -297,19 +306,153 @@ def require_edit_permission
...
+    if reply.character&.new_record?
+      if reply.character.save
+        flash[:success] = "Your new NPC has been persisted!"
+        params[:reply][:character_id] = reply.character.id
+        reply_params[:character_id] = reply.character.id
+        reply.character_id = reply.character.id
+      else
+        flash[:error] = "There was a problem persisting your new NPC."
+      end
+    end
```
teceler, line 344: Terrifying!!! too many things to handle.

app/controllers/replies_controller.rb
```diff
@@ -297,19 +306,153 @@ def require_edit_permission
...
+
+    # Set up editor
+    @post = reply.post
+    empty_reply_hash = permitted_params.permit(:character_id, :icon_id, :character_alias_id)
+    @empty_written = @post.build_new_reply_for(current_user, empty_reply_hash)
+    @empty_written.editor_mode = reply.editor_mode
+    @reply = @empty_written # So that editor_setup shows the correct characters
```
teceler, line 351: Huh, what's the deal with this?

app/controllers/replies_controller.rb
```diff
@@ -297,19 +306,153 @@ def require_edit_permission
...
+    @audits = {}
+
+    # Add reply to list of multi replies
+    reply.user = current_user unless reply.user
```
teceler, line 355: Possibly you want to set this when you're setting up the reply in the last chunk?

app/controllers/replies_controller.rb
```diff
@@ -297,19 +306,153 @@ def require_edit_permission
...
+
+      redirect_to posts_path and return unless errored_reply.post
+      redirect_to post_path(errored_reply.post) and return
+    end
+
+    if @multi_replies.length == 1
+      flash[:success] = "Reply posted."
+    else
+      flash[:success] = "Replies posted."
+    end
```
teceler, line 388:
```suggestion
    flash[:success] = "#{'Reply'.pluralize(@multi_replies.length)} posted."
```

app/controllers/replies_controller.rb
```diff
@@ -297,19 +306,153 @@ def require_edit_permission
...
+    redirect_to reply_path(first_reply, anchor: "reply-#{first_reply.id}")
+  end
+
+  def editing_multi_reply?
+    # If the list of params is present and the first item on the list has the ID stored, I am editing it
+    @multi_replies_params.present? && (@reply = Reply.find_by_id(@multi_replies_params.first["id"]))
```
teceler, line 394:
```suggestion
    @multi_replies_params.present? && (@reply = Reply.find_by(id: @multi_replies_params.first["id"]))
```

app/controllers/replies_controller.rb
```diff
@@ -297,19 +306,153 @@ def require_edit_permission
...
+
+        @reply.save!
+      end
+    rescue ActiveRecord::RecordInvalid => e
+      if @multi_replies.blank?
+        render_errors(@reply, action: 'updated', now: true, err: e)
+      else
+        errored_reply = @multi_replies.detect { |r| r.errors.present? } || @reply
+        render_errors(errored_reply, action: 'updated', now: true, err: e)
+      end
```
teceler, line 410: hmm, might be worth looking at what gallery#add does for erroring icons, here

app/controllers/replies_controller.rb
```diff
@@ -297,19 +306,153 @@ def require_edit_permission
...
     @written.user = current_user unless @written.user
     @audits = @written.id.present? ? { @written.id => @written.audits.count } : {}
+    @reply = @written # So that editor_setup shows the correct characters
+    preview_replies
+  end
 
+  def add_to_multi_reply(reply, reply_params)
+    # Adding a new reply to a multi reply
+    post_id = params[:reply][:post_id]
+    ReplyDraft.draft_for(post_id, current_user.id)&.destroy!
```
teceler, line 332: Why are we destroying the draft?

app/controllers/replies_controller.rb
```diff
@@ -297,19 +306,153 @@ def require_edit_permission
...
+    if new_multi_reply.present?
+      # Including the reply in the text editor, not just the ones in the JSON
+      @multi_replies << new_multi_reply
+      @multi_replies_params << permitted_params
+    end
+    original_reply = @reply.dup
+
+    # Modify the original reply with the parameters of the first reply in the JSON
+    @reply.assign_attributes(@multi_replies_params.shift)
+    @multi_replies.shift
```
teceler, line 433: Hang on what is this doing

app/controllers/replies_controller.rb
```diff
@@ -297,19 +306,153 @@ def require_edit_permission
...
+
+    # Check whether there are any further replies beyond the very first one
+    num_new_replies = @multi_replies_params.length
+    return if num_new_replies == 0
+
+    # Reorder the replies after this one
+    original_reply_order = @reply.order
+    following_replies = @reply.post.replies.where("reply_order > ?", original_reply_order)
+    following_replies.update_all(["reply_order = reply_order + ?", num_new_replies]) # rubocop:disable Rails/SkipsModelValidations
```
teceler, line 442: Surely you don't want to do that until you actually post it? This should probably go inside the transaction for that

app/controllers/replies_controller.rb
```diff
@@ -297,19 +306,153 @@ def require_edit_permission
...
+    # Create the new replies
+    @multi_replies_params.each_with_index do |reply_params, idx|
+      # Create a fake temporary reply with the contents of the original one to be in history
+      @multi_replies[idx] = new_reply = original_reply.dup
+      new_reply.order = original_reply_order + idx + 1
+      new_reply.save!
+
+      # Update the new reply added with the actual params that should be there
+      new_reply.update!(reply_params)
+    end
```
teceler, line 453: Again, this should be in a transaction

app/controllers/writable_controller.rb
```diff
@@ -209,7 +209,7 @@ def process_npc(writable, permitted_character_params)
     )
   end
 
-  def permitted_params(param_hash=nil)
+  def permitted_params(param_hash=nil, extra_params=[])
```
teceler, line 212: What's extra_params doing?

app/views/replies/_write.haml
```diff
@@ -28,12 +30,19 @@
...
+        - if reply.new_record?
+          = submit_tag "Save Draft", class: 'button', id: 'draft_button', name: 'button_draft', data: { disable_with: 'Drafting...' }
+        - if @draft.present?
+          = submit_tag "Delete Draft", class: 'button', id: 'delete_draft_button', name: 'button_delete_draft', data: { disable_with: 'Deleting...', confirm: "Delete draft?" }
+      - else
+        = submit_tag (multi_reply_new ? 'Post All' : 'Save All'), class: 'button', id: 'submit_button', data: { disable_with: 'Saving...' }
+        = submit_tag "Preview Current", class: 'button', id: 'preview_button', name: 'button_preview', data: { disable_with: 'Previewing...' }
+      = submit_tag "Add More Replies", class: 'button', id: 'add_more_button', name: 'button_add_more', data: { disable_with: 'Adding...' }
+      - if multi_replies_params.present?
+        = submit_tag "Discard Replies", class: 'button', id: 'discard_multi_reply_button', name: 'button_discard_multi_reply', data: { disable_with: 'Discarding...', confirm: "Discard all replies added?" }
```
teceler, line 45: Oh god you made replies/write worse (here meaning 'more complicated')

app/views/writable/_editor.haml
```diff
@@ -1,50 +1,50 @@
--# locals: ( item: )
+-# locals: ( writable: )
```
teceler, line 1: I support this move and variable renaming tbh

teceler: Possibly it should be pulled out into another PR tho?

app/controllers/replies_controller.rb
redmvc: If you discard the multi replies from writing new ones you go to unread, but if you discard them from _editing_ one into multiple you go to the one you were editing.

app/controllers/replies_controller.rb
redmvc: 'Cause I split the original "preview" function into a part that is shared between "literally previewing" and "editing multi replies" and a part that is specific to each of those, so now `preview_reply` does the "literally previewing"-specific logic then calls `preview`, while `add_to_multi_reply` does the "editing multi replies"-specific logic then calls `preview`

app/controllers/replies_controller.rb
redmvc: Not one I know, @Throne3d is the one who suggested that one iirc

app/controllers/replies_controller.rb
redmvc: 😔

app/controllers/replies_controller.rb
redmvc: This is so that the editor will always show no reply + the prior multi reply's character, the way the regular editor at the bottom of a thread's page does

app/controllers/replies_controller.rb
redmvc: Which part?

app/controllers/replies_controller.rb
redmvc: I'm pretty sure we're using find_by_id most elsewhere in the code?

app/controllers/replies_controller.rb
redmvc: So that the draft goes into the multi reply and then the editor is empty and waiting for the next reply to be added to the list

app/controllers/replies_controller.rb
teceler: where's the `@empty_written` being used distinct from the `@reply`?

app/controllers/replies_controller.rb
redmvc: Actually editing the reply it's meant to edit

app/controllers/replies_controller.rb
redmvc: That is inside that transaction, that function is only called from within the transaction

app/controllers/replies_controller.rb
teceler: ...wait, that's `reply` not `@reply`, never...mind? probably?

app/controllers/replies_controller.rb
redmvc: It is!

app/controllers/replies_controller.rb
teceler: Oh. sneaky. Okay that makes... sense?

app/controllers/replies_controller.rb
teceler: I thought we were moving away from find_by_id to find_by but -- this is probably also a subject for discord

app/controllers/replies_controller.rb
teceler: Ah, that makes sense.

app/controllers/writable_controller.rb
redmvc: When I'm fetching the data of the multi replies JSON from the hidden field on the page, the very first reply might have a reply ID in case I'm editing a reply (rather than posting new ones), and so I pass :id as an extra permitted param there

app/controllers/replies_controller.rb
teceler: Why is it using shift for that

app/views/writable/_editor.haml
redmvc: I could do that if you think this PR will still need a lot of work before merging?

app/views/replies/_write.haml
redmvc: Sorry 😅

app/controllers/writable_controller.rb
teceler: Aha, that makes sense.

app/controllers/replies_controller.rb
redmvc: `@reply` is used by the `editor_setup` function, `@empty_written` is used by the `_preview` partial

redmvc: ...not partial. just page.

app/views/writable/_editor.haml
teceler: Dunno! It's kind of terrifying but it does seem like it has solid tests.

app/controllers/replies_controller.rb
redmvc: Cause the reply it's meant to edit is the first one of the list, so I'm grabbing its data and removing it from there

app/controllers/replies_controller.rb
teceler: Oh but i think you could
```suggestion
    # Save NPC
    if reply.character&.new_record?
      if reply.character.save
        flash[:success] = "Your new NPC has been persisted!"
        params[:reply][:character_id], reply_params[:character_id], reply.character_id = reply.character.id
      else
        flash[:error] = "There was a problem persisting your new NPC."
      end
    end
```
I'm not sure you *should* but I'm pretty sure that works

teceler: Do you actually have to manually set `reply.character_id` here, rails doesn't handle that?

app/controllers/replies_controller.rb
teceler: Aha. The comment confused me, but that makes sense.

app/controllers/replies_controller.rb
redmvc: It doesn't! I found it out the hard way!

I also think I don't want to do the thing you suggested if you don't have a super strong preference because it is less readable to me personally but I'm not married to mine

teceler: If there were only two of them I might encourage that, but with three it starts being a Lot imo, feel free to go with your thing.

app/controllers/replies_controller.rb
teceler: Having looked at that earlier I'm not sure why you didn't just use `@reply` when adding a variable to it but ok

app/controllers/replies_controller.rb
redmvc: It seems like we just add to a list of strings saying which icons have errors and then paste them to the error flash:

```ruby
Icon.transaction do
  icons.each_with_index do |icon, index|
    next if icon.save
    errors += icon.errors.present? ? icon.get_errors(index) : ["Icon #{index + 1} could not be saved."]
  end
  raise ActiveRecord::Rollback if errors.present?
  @gallery.icons += icons if @gallery
end

if errors.present?
  flash.now[:error] = {
    message: "Icons could not be saved because of the following problems:",
    array: errors,
  }
  render :add
else
  flash[:success] = "Icons saved."
  redirect_to @gallery || user_gallery_path(id: 0, user_id: current_user.id)
end
```

app/controllers/replies_controller.rb
redmvc: 'cause I do different things depending on whether `@written` is present but I guess I could just pass a boolean maybe?

PR #2388, opened 2025-01-06
Reduce error emails

Marri:
Should be released concurrently with adding a Heroku alert for slow response time, to replace the functionality of the `Rack::Timeout::RequestTimeoutException` emails.

The `ActiveRecord::QueryCanceled` emails are also timeouts, and can be mostly replaced with the Heroku alert as well, but do represent an actual problem we should address with code/infra beyond just with General Speed Things Up specifically for search pages. Do not personally think it's worth keeping error emails about it, but open to input, since we don't have fine tuned alerting for things like "slow searches specifically" atm.

`ActionDispatch::Http::MimeNegotiation::InvalidType` has historically only been thrown by nonsense and I do not expect any of them represent real legitimate user-facing errors, so they can go away too.

We also get various emails of the type
```
  invalid value for Integer(): "unread"
  app/controllers/application_controller.rb:95:in `posts_from_relation'
```
coming from controllers that do not support "last" or "unread" pages, but which were passed it as a page anyway. This should protect us, but plausibly we should be designing this better, so counter-designs welcome.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -92,7 +92,8 @@ class ApplicationController < ActionController::Base
     posts = posts_relation_filter(relation, no_tests: no_tests, show_blocked: show_blocked)
     posts_count = posts.except(:select, :order, :group).count('DISTINCT posts.id')
     posts = posts_list_relation(posts, select: select, max: max)
-    posts = posts.paginate(page: page, total_entries: posts_count) if with_pagination
+    safe_page = page.to_i == 0 ? 1 : page.to_i # protects against 'last' and 'unread' on controllers that don't support it
+    posts = posts.paginate(page: safe_page, total_entries: posts_count) if with_pagination
     calculate_view_status(posts, with_unread: with_unread) if logged_in?
     posts
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -43,11 +43,11 @@ class ApplicationController < ActionController::Base
   end
 
   VALID_PAGES = ['last', 'unread']
-  def page
+  def page(allow_special: false)
     return @page if @page
     return (@page = 1) unless params[:page]
     @page = params[:page]
-    return @page if VALID_PAGES.include?(@page)
+    return @page if allow_special && VALID_PAGES.include?(@page)
     @page = @page.to_i
     return @page if @page > 0
     flash.now[:error] = "Page not recognized, defaulting to page 1."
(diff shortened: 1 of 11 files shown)


app/controllers/application_controller.rb
```diff
@@ -92,7 +92,8 @@ def posts_from_relation(relation, no_tests: true, with_pagination: true, select:
     posts = posts_relation_filter(relation, no_tests: no_tests, show_blocked: show_blocked)
     posts_count = posts.except(:select, :order, :group).count('DISTINCT posts.id')
     posts = posts_list_relation(posts, select: select, max: max)
-    posts = posts.paginate(page: page, total_entries: posts_count) if with_pagination
+    safe_page = page.to_i == 0 ? 1 : page.to_i # protects against 'last' and 'unread' on controllers that don't support it
+    posts = posts.paginate(page: safe_page, total_entries: posts_count) if with_pagination
```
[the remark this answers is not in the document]
Marri: No, if it's gotten to here it'll crash if passed 'last' or 'unread', the controllers should have handled it by calculating an actual number. That said, _do_ we support it anywhere other than writable#show? Which isn't using a post list anyway? https://github.com/glowfic-constellation/glowfic/blob/main/app/controllers/writable_controller.rb#L60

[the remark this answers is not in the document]
teceler: Okay, cool, that makes sense. I wouldn't be *shocked* if we supported last somewhere else but since paginator does a different thing for that probably not. (Where do we pass last, anyway?)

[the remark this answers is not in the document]
Marri: Apparently we don't! We just support it :shrug: but I tweaked it anyway.

app/controllers/application_controller.rb
```diff
@@ -43,11 +43,11 @@ def use_javascript(js)
   end
 
   VALID_PAGES = ['last', 'unread']
-  def page
+  def page(allow_special: false)
     return @page if @page
     return (@page = 1) unless params[:page]
     @page = params[:page]
-    return @page if VALID_PAGES.include?(@page)
+    return @page if allow_special && VALID_PAGES.include?(@page)
```
teceler, line 50: Oh that's clever

teceler: Might be more succinct as a positional rather than argument parameter, looking at the calls later?

config/environments/production.rb
```diff
@@ -100,6 +100,11 @@
 
   # use ExceptionNotification to email Marri stack traces
   Rails.application.config.middleware.use ExceptionNotification::Rack,
+    ignore_exceptions: [
+      'Rack::Timeout::RequestTimeoutException',
+      'ActiveRecord::QueryCanceled',
+      'ActionDispatch::Http::MimeNegotiation::InvalidType',
+    ] + ExceptionNotifier.ignored_exceptions,
```
teceler, line 107: What are `ExceptionNotifier.ignored_exceptions`? Is that just inheriting?

spec/system/characters/index_spec.rb
```diff
@@ -187,6 +187,11 @@ def expect_character_icons(characters)
 
     include_examples "characters#index", false
 
+    scenario "Handles bad pages" do
+      visit user_characters_path(user_id: user.id, view: 'list', character_split: 'template', page: "nvOpzp; AND 1=1")
+      expect(page).to have_text("Sample user's Characters")
+    end
```
teceler, line 193: huh why is this one a system spec?

app/controllers/application_controller.rb
Marri: Per Discord chat: seems clearer to read from the call location if we keep it as an argument parameter, gonna leave it be for now.

spec/system/characters/index_spec.rb
Marri: From Discord: because Marri didn't know enough about the differences between system and request specs to know she should use a request spec, and she will move it.

PR #2391, opened 2025-01-07
Make reply#search a service

teceler:


app/services/reply/searcher.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  def construct_query(params, page)
+    @search_results = @search_results
+      .select('replies.*, characters.name, characters.screenname, users.username, users.deleted as user_deleted')
+      .visible_to(@current_user)
+      .joins(:user)
+      .left_outer_joins(:character)
+      .includes(:post)
+
+    @search_results = @search_results.where.not(post_id: @current_user.hidden_posts) unless @current_user.nil? || params[:show_blocked]
```
Throne3d, line 93: Oh huh, this was being applied _after_ pagination before. Nice catch fixing this!

teceler: Yeah i noticed that and was like 'well that can't be right'

app/services/generic/searcher.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+class Generic::Searcher < Object
```
Throne3d, line 2: I'm a bit confused by this class - I don't think we use `name` or `errors` anywhere, nor ActiveModel::Translation or ActiveModel::Validations, right? Is this going to be used for something later?

teceler: Huh, did I wind up refactoring to not use errors, I guess I did. The Translation and Validations are iirc just so errors functions properly.

Throne3d: Ah ok! I don't think `name` is used here either, right?

Makes sense that we'll want this later if/when we ever see errors in searches that we want to pass up to the controller, but I guess none of these fields are necessary right now

Throne3d (approved):
(Beyond the small unused variables, lgtm!)

app/services/generic/searcher.rb
teceler: `:name` is also for errors iirc, I think at some point i worked out the minimal viable setup for using activemodel errors.

teceler: I basically just pulled this class from my old service PRs tho xD

PR #2392, opened 2025-01-09
put the "previous post" button back on the left

CalendulaAsteraceae:
Put the "previous post" button on the left of the page, instead of floating to the right just before the "next post" button.


Marri (approved):
:turtle: sorry Code Climate is being dumb about the diff coverage; we don't track coverage of haml files right now, so if it won't let you merge I can do it for you.

CalendulaAsteraceae:
> 🐢 sorry Code Climate is being dumb about the diff coverage; we don't track coverage of haml files right now, so if it won't let you merge I can do it for you.

Please do!

PR #2406, opened 2025-01-20
Fix various small issues with the multi-reply editor

redmvc:
- Make the quick selector behave in intuitive ways, including adding new characters in the middle of the multi replies.
- Make the alias shown when selecting a character correctly fetch the alias used in the multi replies.
- Use the correct icon on the blank reply textbox (i.e. a character's default) rather than the last icon that was used.
- Make the "there are unseen replies" and "this reply might be a duplicate" warnings not show up when splitting an existing reply.
- Make the "this reply might be a duplicate" warning on adding a new set of multi replies match the first reply being added not the last.
- Remove the reply buttons (permalink, edit, etc) from the top reply of a list of multi replies.

### DIFF
--- a/app/assets/javascripts/posts/editor.js
+++ b/app/assets/javascripts/posts/editor.js
@@ -168,7 +168,7 @@ function setupWritableEditor() {
   $(".char-access-icon").click(function() {
     const id = $(this).data('character-id');
     $("#reply_character_id").val(id);
-    getAndSetCharacterData({ id: id }, { updateCharDropdowns: true,  aliasOverride: findMultiReplyAlias(id)});
+    getAndSetCharacterData({ id: id }, { updateCharDropdowns: true, aliasOverride: findMultiReplyAlias(id)});
   });
 
   $("#character_alias").change(function() {
--- a/app/assets/javascripts/posts/editor.js
+++ b/app/assets/javascripts/posts/editor.js
@@ -80,6 +80,25 @@ function setupMetadataEditor() {
   });
 }
 
+function findMultiReplyAlias(characterId) {
+  // Try to find a character's last used alias on a multi-reply page
+  const multiRepliesJsonElement = document.querySelector("#multi_replies_json");
+  if (multiRepliesJsonElement === null) {
+    return null;
+  }
+
+  // I'm editing a multi-reply, so the relevant character might have aliases here that are not set in the thread
+  let aliasOverride = null;
+  const jsonArray = JSON.parse(multiRepliesJsonElement.value);
+  const latestElement = jsonArray.slice().reverse().find(item => item.character_id === String(characterId));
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/posts/editor.js
```diff
@@ -80,6 +80,25 @@ function setupMetadataEditor() {
...
+  // Try to find a character's last used alias on a multi-reply page
+  const multiRepliesJsonElement = document.querySelector("#multi_replies_json");
+  if (multiRepliesJsonElement === null) {
+    return null;
+  }
+
+  // I'm editing a multi-reply, so the relevant character might have aliases here that are not set in the thread
+  let aliasOverride = null;
+  const jsonArray = JSON.parse(multiRepliesJsonElement.value);
+  const latestElement = jsonArray.slice().reverse().find(item => item.character_id === String(characterId));
```
Throne3d, line 93:
```suggestion
  const latestElement = jsonArray.reverse().find(item => item.character_id === String(characterId));
```

app/controllers/replies_controller.rb
```diff
@@ -294,8 +259,13 @@ def add_to_multi_reply(reply, reply_params)
...
     # Set up editor
     @post = reply.post
-    empty_reply_hash = permitted_params.permit(:character_id, :icon_id, :character_alias_id)
+    empty_reply_hash = permitted_params.permit(:character_id, :character_alias_id)
     @reply = @post.build_new_reply_for(current_user, empty_reply_hash)
+    if @reply.character_id.nil?
+      @reply.icon_id = current_user.avatar_id
+    else
+      @reply.icon_id = @reply.character.default_icon.try(:id)
+    end
```
Throne3d, line 268:
```suggestion
    @reply.assign_default_icon
```
and extract this here in and `build_new_reply_for`?

app/models/post.rb
```diff
@@ -178,15 +178,24 @@ def last_seen_reply_for(user)
...
-      .select('DISTINCT character_id, MAX(id)')
-      .order(Arel.sql('MAX(id) desc'))
-      .pluck(:character_id)
+  def recent_characters_for(user, count, multi_replies_params: nil)
+    # fetch the (count) most recent non-nil character_ids for user in post, including those being added by multi-replies
+    recent_ids = []
+    if multi_replies_params
+      recent_ids = multi_replies_params.reverse.pluck(:character_id).compact_blank.uniq.take(count).map(&:to_i)
+      count -= recent_ids.length
+    end
```
teceler, line 187: You've otherwise been using an instance variable for `multi_replies_params`?

Throne3d (approved):

app/models/post.rb
Throne3d: This is on the post model though right? Previously it's on the controller

teceler: Ah, so it is, and the method does pass it, so

PR #2409, opened 2025-01-21
Load bookmarks only once per loaded post page

redmvc:


app/models/reply.rb
```diff
@@ -52,7 +52,7 @@ def order=(val)
 
   def bookmark_by(user)
```
Throne3d, line 53: Is there anything still using this after the change?

Throne3d (approved):
LGTM assuming it passes tests

PR #2411, opened 2025-01-22
Add DraftsController

teceler:
Move draft creation and destruction to its own controller


teceler:
A bunch of our reply integration tests save drafts, so I'm tentatively convinced this works fine?

config/routes.rb
```diff
@@ -106,6 +106,9 @@
     end
     collection { get :search }
   end
+  resources :drafts, only: [:create, :destroy] do
+    collection { post :preview }
```
Throne3d, line 110: Do we have a `drafts#preview`? I don't think this is used based on the controller!

teceler: Oh yeah that's leftover from when I was also trying to get a preview path working, i should probably remove it. (That will have another PR later probably but didn't need to be in this one xD)

Throne3d (approved):

config/routes.rb
Throne3d: Ooh ok! Yeah, I think we can remove it from this PR but otherwise LGTM!

PR #2419, opened 2025-01-29
Add ExceptionNotification on Resque failure

Throne3d:
As per https://github.com/smartinez87/exception_notification/blob/master/lib/generators/exception_notification/templates/exception_notification.rb.erb#L7

Not 100% sure how to avoid this potentially entering an infinite loop if something goes wrong with ExceptionNotifier itself (does it send async?), but would be good to have something like this to make sure we get notifications when emails are broken.

### DIFF
--- a/config/initializers/resque.rb
+++ b/config/initializers/resque.rb
@@ -28,3 +28,11 @@ end
 # require 'resque/failure/redis'
 # Resque::Failure::MultipleWithRetrySuppression.classes = [Resque::Failure::Redis]
 # Resque::Failure.backend = Resque::Failure::MultipleWithRetrySuppression
+
+# logs Resque failures and sends them to ExceptionNotification
+require 'resque/failure/multiple'
+require 'resque/failure/redis'
+require 'exception_notification/resque'
+
+Resque::Failure::Multiple.classes = [Resque::Failure::Redis, ExceptionNotification::Resque]
+Resque::Failure.backend = Resque::Failure::Multiple


Throne3d:
Could fire the equivalent of https://github.com/smartinez87/exception_notification/blob/master/lib/exception_notification/resque.rb#L11 manually upon failures in a specific mailer instead! Might be safer to avoid infinite loops

config/initializers/resque.rb
```diff
@@ -28,3 +28,11 @@
 # require 'resque/failure/redis'
 # Resque::Failure::MultipleWithRetrySuppression.classes = [Resque::Failure::Redis]
 # Resque::Failure.backend = Resque::Failure::MultipleWithRetrySuppression
+
+# logs Resque failures and sends them to ExceptionNotification
+require 'resque/failure/multiple'
+require 'resque/failure/redis'
+require 'exception_notification/resque'
+
+Resque::Failure::Multiple.classes = [Resque::Failure::Redis, ExceptionNotification::Resque]
```
Throne3d, line 37: ExceptionNotification's EmailNotifier tries to use deliver_now if it's available (so shouldn't trigger a background job, so shouldn't cause infinite loop issues here)

https://github.com/smartinez87/exception_notification/blob/60e158888bbff96f7a9ae00ae97305cdd33726e9/lib/exception_notifier/email_notifier.rb#L199

I think we're good!

PR #2436, opened 2025-02-09
Locking various functionality with ENV variables

Marri:


teceler:
We probably also wanna do something with the visible_to scope

PR #2444, opened 2025-02-20
Improve messaging with upgrade locks

Marri:


app/controllers/users_controller.rb
```diff
@@ -36,7 +38,12 @@ def create
       render :new and return
     end
 
-    @user.role_id = Permissible::READONLY if params[:secret] != ENV["ACCOUNT_SECRET"]
+    if params[:secret] != ENV["ACCOUNT_SECRET"]
+      @user.role_id = Permissible::READONLY
+    elsif ENV.fetch("UPGRADES_LOCKED", nil).present?
+      flash.now[:error] = "We're sorry, full accounts are currently unavailable."
+      @user.role_id = Permissible::READONLY
+    end
```
teceler, line 46: I was going to say I think these are in the wrong order and they might still be but only because of the flash, the first one is also setting reader

app/views/users/new.haml
```diff
@@ -29,11 +29,12 @@
...
-      .sub.vtop= f.label :secret, "Secret (optional)"
-      .even
-        = text_field_tag :secret, params[:secret], placeholder: "Secret"
-        %span.details (If you don't know what this is, please leave it blank!)
+    - unless ENV.fetch("UPGRADES_LOCKED", nil).present?
+      #signup-secret
+        .sub.vtop= f.label :secret, "Secret (optional)"
+        .even
+          = text_field_tag :secret, params[:secret], placeholder: "Secret"
+          %span.details (If you don't know what this is, please leave it blank!)
```
teceler, line 37: Oh this is smart

PR #2463, opened 2025-03-12
Handle ignored notifications in unread_notif_count

teceler:


app/views/layouts/application.haml
```diff
@@ -87,7 +87,9 @@
                   %li
                     = link_to notifications_path do
                       Notifications
-                      - unread_notif_count = current_user.notifications.unread.visible_to(current_user).count
+                      - unread_notifs = current_user.notifications.unread.visible_to(current_user)
+                      - unread_notifs = unread_notifs.not_ignored_by(current_user) if current_user&.hide_from_all
```
Marri, line 91: I don't think we need the `&.` here since it's within a `logged_in?` block

teceler: Oh probably not I just copied it over from notifications#index

app/views/layouts/application.haml
```diff
@@ -87,7 +87,9 @@
                   %li
                     = link_to notifications_path do
                       Notifications
-                      - unread_notif_count = current_user.notifications.unread.visible_to(current_user).count
+                      - unread_notifs = current_user.notifications.unread.visible_to(current_user)
+                      - unread_notifs = unread_notifs.not_ignored_by(current_user) if current_user.hide_from_all
+                      - unread_notif_count = unread_notifs.count
```
Throne3d, line 92: Maybe we can extract this logic (`current_user.notifications.visible_to(current_user)` + `.not_ignored_by(current_user) if current_user.hide_from_all`) somewhere so we can share it between the NotificationsController and here (so it doesn't get out of sync)

but lgtm!

Throne3d (approved):

PR #2506, opened 2025-05-12
Avoid using negation matchers before verifying the page loaded

teceler:
Also some general tests cleanup


spec/system/replies/new_spec.rb
```diff
@@ -391,19 +399,24 @@
         fill_in id: "reply_content", with: reply.content
         click_button "Add More Replies"
       end
-      expect(page).to have_no_selector('.error')
+
+      # TODO: add positive selector check
```
Throne3d, line 403: (are these TODOs for a follow-up PR?)

teceler: ...no, I missed those when i was going back through, whoops

Throne3d (approved):
LGTM, nice!

PR #2547, opened 2025-06-20
Add includes to reduce N+1 queries

setupminimal:
I have no particular insight on whether this will _help_ with the current performance problems, but I don't see how it can hurt.

I used Bullet to look for places where the app was making N+1 queries — making a subsequent database request for every item in a collection. Ten of those places turned out to be too complicated for me to figure out how to fix without making deep changes to the code. This PR has everything _else_, i.e., everything that was easy to fix.

This will have slightly higher memory usage because more will be pulled from the DB, but not by very much. And since it will be pulled into a signal array that can get garbage collected at once, instead of pulled as many separate database transactions, it should be less work for the site overall.

### DIFF
--- a/app/controllers/icons_controller.rb
+++ b/app/controllers/icons_controller.rb
@@ -56,7 +56,7 @@ class IconsController < UploadingController
       @times_used = (posts_using.count + replies_using.count)
       @posts_used = (posts_using.pluck(:id) + replies_using.select(:post_id).distinct.pluck(:post_id)).uniq.count
     end
-    @galleries = @icon.galleries.ordered_by_name
+    @galleries = @icon.galleries.includes([:default_icon]).ordered_by_name
     @meta_og = og_data
     response.headers['X-Robots-Tag'] = 'noindex' if params[:view]
   end
--- a/app/controllers/replies_controller.rb
+++ b/app/controllers/replies_controller.rb
@@ -13,7 +13,7 @@ class RepliesController < WritableController
     @page_title = 'Search Replies'
     use_javascript('search')
 
-    @post = Post.find_by(id: params[:post_id]) if params[:post_id].present?
+    @post = Post.find_by(id: params[:post_id]).includes([:audits]) if params[:post_id].present?
     @icon = Icon.find_by(id: params[:icon_id]) if params[:icon_id].present?
 
     if @post&.visible_to?(current_user)
--- a/app/views/writable/_editor.haml
+++ b/app/views/writable/_editor.haml
@@ -1,6 +1,6 @@
 -# locals: ( writable:, multi_replies_params: nil )
 
-- character_galleries = writable.character.galleries.ordered if writable.character && current_user.icon_picker_grouping? && writable.character.galleries.count > 1
+- character_galleries = writable.character.galleries.includes([:icons]).ordered if writable.character && current_user.icon_picker_grouping? && writable.character.galleries.count > 1
 #post-editor.padding-10
   .post-info-box
     #current-icon-holder.post-icon
--- a/app/controllers/bookmarks_controller.rb
+++ b/app/controllers/bookmarks_controller.rb
@@ -27,6 +27,7 @@ class BookmarksController < ApplicationController
 
     @search_results = @search_results
       .joins(:post)
+      .includes([:icon])
       .order('posts.subject, replies.created_at, posts.id')
       .joins(:user)
       .left_outer_joins(:character)
--- a/app/controllers/characters_controller.rb
+++ b/app/controllers/characters_controller.rb
@@ -167,9 +167,10 @@ class CharactersController < ApplicationController
   def replace
     @page_title = 'Replace Character: ' + @character.name
     if @character.template
-      @alts = @character.template.characters
+      @alts = @character.template.characters.includes([:aliases, :default_icon])
     else
-      @alts = @character.user.characters.where(template_id: nil)
+      @alts = @character.user.characters.includes([:aliases, :default_icon])
+                .where(template_id: nil)
     end
     @alts -= [@character] unless @alts.size <= 1 || @character.aliases.exists?
     use_javascript('icons')
--- a/app/controllers/templates_controller.rb
+++ b/app/controllers/templates_controller.rb
@@ -78,7 +78,8 @@ class TemplatesController < ApplicationController
 
   def editor_setup
     @selectable_characters = @template.try(:characters) || []
-    @selectable_characters += current_user.characters.where(template_id: nil).ordered
+    @selectable_characters += current_user.characters.includes([:default_icon])
+                                .where(template_id: nil).ordered
     @selectable_characters.uniq!
     @character_ids = permitted_params[:character_ids] if permitted_params.key?(:character_ids)
     @character_ids ||= @template.try(:character_ids) || []


app/controllers/icons_controller.rb
```diff
@@ -56,7 +56,7 @@ def show
       @times_used = (posts_using.count + replies_using.count)
       @posts_used = (posts_using.pluck(:id) + replies_using.select(:post_id).distinct.pluck(:post_id)).uniq.count
     end
-    @galleries = @icon.galleries.ordered_by_name
+    @galleries = @icon.galleries.includes([:default_icon]).ordered_by_name
```
teceler, line 59: ...i think default_icon is a *character* relation? what's going on here

app/controllers/replies_controller.rb
```diff
@@ -13,7 +13,7 @@ def search
     @page_title = 'Search Replies'
     use_javascript('search')
 
-    @post = Post.find_by(id: params[:post_id]) if params[:post_id].present?
+    @post = Post.find_by(id: params[:post_id]).includes([:audits]) if params[:post_id].present?
```
Marri, line 16: Hm, I'm concerned about this because I think we're not supposed to *use* audits on the reply search page (we set it to [] on purpose so we don't do the query). This should more likely be fixed by finding if/where it's loading them and telling it to knock it off; we don't care in the context of search results.

app/views/writable/_editor.haml
```diff
@@ -1,6 +1,6 @@
 -# locals: ( writable:, multi_replies_params: nil )
 
-- character_galleries = writable.character.galleries.ordered if writable.character && current_user.icon_picker_grouping? && writable.character.galleries.count > 1
+- character_galleries = writable.character.galleries.includes([:icons]).ordered if writable.character && current_user.icon_picker_grouping? && writable.character.galleries.count > 1
```
Marri, line 3: This one looks good from what I can tell :turtle:

app/controllers/replies_controller.rb
setupminimal: Ah, okay. That seems easy enough.

app/controllers/messages_controller.rb
```diff
@@ -19,8 +19,12 @@ def index
       from_table = from_table.where.not(sender_id: blocked_ids).left_outer_joins(:sender).where('users.deleted IS NULL OR users.deleted = false')
       includes << :sender
     end
+    thread_counts = Message.from(from_table, "messages")
+      .select('thread_id, COUNT(*) as thread_count')
+      .group('thread_id').to_sql
     @messages = Message.from(from_table, "messages").joins(:first_thread)
-      .select('*', 'first_threads_messages.subject as thread_subject')
+      .joins("LEFT JOIN (#{thread_counts}) AS thread_counts ON thread_counts.thread_id = messages.thread_id")
+      .select('*', 'first_threads_messages.subject as thread_subject', 'thread_counts.thread_count')
```
teceler, line 27: What are you doing here? Or -- what is the purpose of this, I suppose.

app/models/character.rb
```diff
@@ -19,6 +19,7 @@ class Character < ApplicationRecord
 
   has_many :character_tags, inverse_of: :character, dependent: :destroy
   has_many :settings, -> { ordered_by_char_tag }, through: :character_tags, source: :setting, dependent: :destroy
+  has_many :unordered_settings, through: :character_tags, source: :setting, dependent: :destroy
```
teceler, line 22: I wouldn't expect that to remove any queries, given it's going through char_tags anyway? unless rails magic will join those without the intervening model.

app/controllers/messages_controller.rb
teceler: Okay i see the *purpose* further down is for num_in_thread but i'm still moderately confused as to how this even works.

app/views/characters/_group.haml
```diff
@@ -8,8 +8,8 @@
       = group ? "Group: #{group.name}" : 'Ungrouped Characters'
 - if characters.try(:exists?)
   - template_ids = characters.group(:template_id).pluck(:template_id).uniq.compact
-  - ::Template.where(id: template_ids).ordered.each do |template|
-    - next if skip_grouped_templates && template.characters.where.not(character_group_id: nil).exists?
+  - ::Template.select('*', 'exists (select * from characters where characters.template_id = templates.id and characters.character_group_id is not NULL) as has_non_null_character').where(id: template_ids).ordered.each do |template|
+    - next if skip_grouped_templates && template.has_non_null_character
```
teceler, line 12: Hm, this feels like there should be a less... obscure... way to check this. Also we hardly ever load character group view but that's not an excuse.

app/controllers/application_controller.rb
```diff
@@ -16,6 +16,17 @@ class ApplicationController < ActionController::Base
...
+    around_action :n_plus_one_detection
+
+    def n_plus_one_detection
+      Prosopite.scan
+      yield
+    ensure
+      Prosopite.finish
+    end
+  end
+
```
teceler, line 29: Hmm, I don't think this is how I'd do this, even if I wanted it on in both test and dev -- which i'm also not sure about -- what i would do is probably

```
  before_action :check_forced_logout
  around_action :set_timezone
  around_action :n_plus_one_detection
  after_action :store_location
```

and then down after the other methods
```
  def n_plus_one_detection
    return if Rails.env.production?
    Prosopite.scan
    yield
  ensure
    Prosopite.finish
  end
```
which is i think more in keeping with project style

teceler (commented):
I'll also look at the ones you marked as difficult and see if they're magic I know

app/controllers/messages_controller.rb
setupminimal: This makes the database engine do the counting, producing a new table, and then joins that onto the existing query.

It is, I admit, a bit esoteric.

app/models/character.rb
setupminimal: So the problem here was that because of the `-> { ordered_by_char_tag }`, I couldn't use the existing :settings relation in an .includes. So adding a second relation like this does enable removing queries.

app/controllers/messages_controller.rb
teceler: OH it's a SUBQUERY

app/views/characters/_group.haml
setupminimal: My Rails SQL skills are not expert; I would be completely unsurprised to learn that there's a more elegant way to do this.

I've definitely seen other `selects` in the code though. Do you have a suggestion of what would be better?

app/controllers/application_controller.rb
setupminimal: Alright; that's an easy fix.

app/views/characters/_group.haml
teceler: I was just looking into this! it looks like possibly rails expects you to do something like
```suggestion
  - templates = ::Template.where(id: template_ids)
  - templates.where.not(id: users.characters.where('character_group_id IS NOT NULL').select(:template_id).uniq) if skip_grouped_templates
  - templates.ordered.each do |template|
```
or similar. since they did stuff to their exists method. 

`.where.not(character_group_id: nil` *ought* to work but iirc instead does something weird.

Anyway i think that's approximately the same (selects templates that have characters with a group id and excludes them) and postgres is smart enough to do it with a join if that's more efficent.

app/controllers/messages_controller.rb
teceler: Okay, I... don't think I'd do that that way but i'm not sure how *else* you would do it in one query either, because you don't want it grouped...

...it's also annoying that you can't to_sql rails `count` apparently, that would simply your subquery...

PR #2559, opened 2025-07-12
Fix: Preserve gallery page after deleting multiple icons (#2368)

### LINKED ISSUE #2368: Icons#delete_multiple should redirect back to current gallery page
https://discord.com/channels/521407923629457428/521407923629457431/1324055463188496384

Currently going to a specific **paginated** gallery (galleries#show, e.g. /galleries/2276), selecting icons, clicking "Remove selected icons from gallery", resets you to the first page of the gallery. Preserving the page would be good, requested by Anya.

j-stri:
## Problem
When deleting multiple icons from a paginated gallery page, users are redirected to the first page instead of staying on their current page. This creates a poor user experience, especially when working with icons on later pages.

## Solution/Changes
In app/views/galleries/_list_section.haml and app/views/galleries/_icon_view.haml, added page and per_page parameters (the latter accessible only through directly writing it in the URL). 
Altered logic in icons_controller to redirect you to the page you were previously on, taking into account the per_page.
Changed galleries_controller and application_controller to enable per_page parameter.

## Testing
Wrote tests to create paginated gallery with 10 icons per page. 
Tested deletion:
- page 1 -> stay on page 1, 
- page 2 and icons left -> stay on page 2
- page 2 and no more icons left -> redirected to page 1 because page 2 is now empty.

Fixes #2368


app/controllers/application_controller.rb
```diff
@@ -64,6 +64,13 @@ def per_page
   end
   helper_method :per_page
 
+  def icons_per_page
+    default = 100
+    per = params[:per_page].present? ? params[:per_page].to_i : default
+    per = default if per <= 0 || per > default
+    @per_page = per || @per_page || default
+  end
+
```
teceler, line 73: I think possibly this should be merged with general per_page?

app/controllers/icons_controller.rb
```diff
@@ -164,13 +164,24 @@ def require_permission
...
   end
 
+  def calculate_page_for_gallery(gallery)
+    per = icons_per_page
+    max_page = [(gallery.icons.count / per.to_f).ceil, 1].max
+    requested_page = [params[:page].to_i, 1].max
+    page_to_use = [requested_page, max_page].min
+    page_to_use = nil if page_to_use <= 1
+    page_to_use
+  end
```
teceler, line 174: huh, I'm not finding an idiomatic way to do this. If it passes rubocop it's fine

app/controllers/icons_controller.rb
```diff
@@ -164,13 +164,24 @@ def require_permission
...
+
   def icon_redirect(gallery)
     if params[:return_to] == 'index'
       redirect_to user_galleries_path(current_user, anchor: "gallery-#{gallery.id}")
     elsif params[:return_tag].present? && (tag = Tag.find_by(id: params[:return_tag]))
       redirect_to tag_path(tag, anchor: "gallery-#{gallery.id}")
     elsif gallery
-      redirect_to gallery_path(id: gallery.id)
+      page_to_use = calculate_page_for_gallery(gallery)
+      per_page_num = icons_per_page == 100 ? nil : icons_per_page
```
teceler, line 183: Why is this being checked against 100?

teceler (commented):
I'm not sure why you're doing icons_per_page as its own thing but I think otherwise it looks fine?

app/controllers/icons_controller.rb
teceler: Ah, to use the default page if it's 1?

PR #2731, opened 2026-02-16
Add auto-switching theme variants (prefers-color-scheme)

l1n:
## Summary
- Add "Auto" variants for each theme (Default, Starry, Starry Light, Monochrome, Milky River, Iconless) that automatically switch between light and dark appearances based on the user's OS/browser `prefers-color-scheme` preference
- Extract CSS rules from 5 selector files into SCSS mixins so they can be included inside `@media` blocks with different variable values — zero CSS output change for existing themes
- Design new **Dark Monochrome** (pure grayscale dark) and **Dark River** (blue-tinted dark) color palettes for themes that lacked dark counterparts
- Handle auto-switching icons server-side by rendering both light/dark variants wrapped in `.auto-light-only`/`.auto-dark-only` CSS utility classes

## Theme Pairings

| Auto Theme | Light Mode | Dark Mode |
|---|---|---|
| Default (Auto) | Default | Dark |
| Starry (Auto) | Starry | Starry Dark |
| Starry Light (Auto) | Starry Light | Starry Dark |
| Monochrome (Auto) | Monochrome | Dark Monochrome (new) |
| Milky River (Auto) | River | Dark River (new) |
| Iconless (Auto) | Default + iconless | Dark + iconless |

## Test plan
- [x] Verify all SCSS compiles without errors (`sass` compilation verified locally)
- [x] Verify existing themes produce identical CSS output after mixin refactor
- [x] Select each auto theme in account settings and toggle `prefers-color-scheme` in browser dev tools
- [x] Verify TinyMCE editor respects auto theme color switching
- [x] Verify privacy icons, swap icons, unread/lastlink icons switch correctly in auto themes
- [x] Test Dark Monochrome and Dark River color palettes for readability and contrast


l1n:
grid_auto_dark_continuity
grid_auto_dark_thread
grid_auto_light_continuity
grid_auto_light_thread
grid_existing_thread

Marri (commented):
I'm inclined to make this the default rather than adding all the (auto) ones; it's a lot less complex and I don't expect "force override my browser" to be a common o compelling use case. Open to thoughts.

l1n:
Fair point — the parallel-themes approach added 6 new theme slots to make the choice opt-in. Two ways I can read your suggestion:

**A. Make all existing themes auto-switching in place.** Each theme file picks light/dark via `prefers-color-scheme`, no user choice. Simpler everything: 6 themes total, no new ones, less SCSS sprawl. Breaking change: users who explicitly picked "Starry" because they wanted the dark version will start getting Starry Light during the day, with no way to override.

**B. Make auto the default for new users, keep fixed-color themes selectable.** New signups land on an auto-switching default; everyone else keeps what they picked. No parallel slots needed if we just have one "Default Auto" and rename the existing one to something explicit.

I think you mean (A)? Happy to refactor that way — it's a real cleanup, and the "force one mode" use case is small enough that "open devtools and pick a fixed theme" is fine. The only thing I'd want to preserve is **Dark Monochrome** and **Dark River** (the new palettes this PR designed) because right now those themes have no dark variant at all, so without auto-switching them users on those themes get only light.

If you confirm (A) I'll restructure: drop the 6 "Auto" entries, fold the dark CSS into each base theme behind `@media (prefers-color-scheme: dark)`, keep the new palettes as the dark branches. Net: -1500 lines, no new theme picker entries, one universal behavior change.

If you want (B) — narrower scope, more conservative — say so and I'll split that out.

PR #2732, opened 2026-02-17
Add sort-by-unread for Unread, Favorites, and Daily Report

l1n:
## Summary
- Adds a `?sort=unread` query parameter to sort posts by most-unread-first in the Unread Posts, Favorites, and Daily Report views
- Implements a `with_unread_count` scope on Post that computes unread reply counts via a correlated SQL subquery, enabling ORDER BY before pagination
- Makes "Unread" and "Last Updated" column headers into clickable sort links that preserve view-specific params (e.g. `started`, pagination) across pages
- Uses the SQL-computed `unread_count` attribute on each post when available, falling back to existing `@unread_counts` display logic for the default sort

## Test plan
- [x] Visit `/posts/unread?sort=unread` — posts should sort by most unread replies first
- [x] Visit `/posts/unread?started=true&sort=unread` — opened posts sort by unread, `started` param preserved across pagination
- [x] Visit `/favorites?sort=unread` — favorite posts sort by unread count
- [x] Visit `/reports/daily?sort=unread` — daily report sorts by unread count
- [x] Click "Last Updated" header to return to default sort
- [x] Verify pagination preserves sort parameter on all views
- [x] Verify default sort (no `?sort` param) behaves identically to before

### DIFF
--- a/app/assets/stylesheets/layout.scss
+++ b/app/assets/stylesheets/layout.scss
@@ -281,6 +281,10 @@ td, .table-list li { font-size: $font_size_small; }
   .link-box { font-size: $font_size_smallish; }
 }
 
+th.sub a.current-sort {
+  font-weight: bold;
+}
+
 .check-all-box {
   /* The .sub or .subber that has the "check all" checkbox */
   width: 30px;
--- a/app/controllers/favorites_controller.rb
+++ b/app/controllers/favorites_controller.rb
@@ -18,7 +18,16 @@ class FavoritesController < ApplicationController
 
     @posts = Post.where(id: author_posts).or(Post.where(id: post_favorites)).or(Post.where(board_id: board_favorites))
     @posts = @posts.not_ignored_by(current_user) if current_user&.hide_from_all
-    @posts = posts_from_relation(@posts.ordered, with_unread: true)
+
+    @sort = params[:sort] if params[:sort] == 'unread'
+    if @sort == 'unread'
+      ordered = @posts.with_unread_count(current_user).order(Arel.sql('unread_count DESC'), tagged_at: :desc)
+    else
+      ordered = @posts.ordered
+    end
+    @posts = posts_from_relation(ordered, with_unread: true)
+    @paginate_params = { sort: @sort }.compact if @sort
+
     @hide_quicklinks = true
   end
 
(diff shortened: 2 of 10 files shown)


l1n:
1_unread_sort_unread
2_opened_sort_unread
3_favorites_sort_unread
4_daily_sort_unread
5_unread_default_sort
6_click_last_updated

Marri (commented):
We don't have a load test environment or anything but my concern about this isn't "does it work", it's "how often will it time out", given unread and search are I believe our two top timeouts atm. Any profiling / query plans / etc we can consider?

l1n:
Fair concern. I can't run EXPLAIN against prod from here, but here's the worst-case analysis from the schema:

The scope is a correlated subquery per post:
```sql
SELECT COALESCE((
  SELECT COUNT(*) FROM replies
  LEFT JOIN post_views AS pv_unread
    ON replies.post_id = pv_unread.post_id AND pv_unread.user_id = $user
  WHERE replies.post_id = posts.id
    AND (pv_unread.read_at IS NULL OR replies.created_at > pv_unread.read_at)
), 0) AS unread_count
```

For each candidate post, PG should:

1. Index-scan `index_replies_on_post_id` to get this post's replies — bounded by reply count for that thread.
2. For each reply, look up `(user_id, post_id)` in `index_post_views_on_user_id_and_post_id` (UNIQUE) — single index hit, microseconds.
3. Filter the joined rows by the OR clause and `COUNT(*)`.

So cost per post is roughly **O(replies on that post)** with an index hit per reply. The Unread page paginates to 25 posts. Even at 1000 replies per post (large RP threads), that's 25k indexed-lookup row evaluations — well under any reasonable threshold.

**What I'd actually be worried about:**

- The page applies `Post.visible_to(user)` *before* this scope, which expands to a giant `WHERE … OR (privacy='access_list' AND id IN (...thousands...))` for users with many circles. That's the existing slowness — it predates this PR — and once it's evaluated, the unread-count subquery only runs over the visible set.
- The `posts_count = posts.count('DISTINCT posts.id')` for pagination uses the same `visible_to` scope twice. PR #2771 (skip DISTINCT when no joins) cuts that for the non-`hide_from_all` case.

If you want me to:
1. Push a draft commit adding `Rails.logger.info` with `ActiveRecord::Base.connection.explain(@posts.to_sql)` behind a `DEBUG_SORT_BY_UNREAD` env var, so you can capture an EXPLAIN on a real session before merge — happy to do that.
2. Or land it behind a feature flag (`sort=unread` is already opt-in via query param, but I could gate the controller path on a per-user setting) so impact is bounded to whoever flips it on.

The view-side fallback also matters: when `sort=unread` isn't set the new scope isn't applied, so the default Unread/Favorites/Daily pages are unchanged.

Let me know which (or both) and I'll push it.

PR #2736, opened 2026-03-14
Add stop word pruning info and exact string match to post search

l1n:
When PostgreSQL's full-text search prunes common words (stop words) from a search query, display a notice explaining which words were ignored. When all words are pruned, automatically fall back to ILIKE matching.

Add an "Exact string match" checkbox that bypasses full-text search entirely and uses ILIKE for the subject field, which handles queries with short/common words that PostgreSQL's English dictionary removes.

https://claude.ai/code/session_01AUdgfaFCydFiBRpbWiHerJ

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -295,8 +295,18 @@ class PostsController < WritableController
       if params[:abbrev].present?
         search = params[:subject].chars.join('% ')
         @search_results = @search_results.where('subject ILIKE ?', "%#{search}%")
+      elsif params[:exact].present?
+        @search_results = @search_results.where('subject ILIKE ?', "%#{params[:subject]}%")
       else
-        @search_results = @search_results.search(params[:subject]).where('subject ILIKE ?', "%#{params[:subject]}%")
+        pruned = detect_pruned_words(params[:subject])
+        if pruned[:all_pruned]
+          @search_results = @search_results.where('subject ILIKE ?', "%#{params[:subject]}%")
+          @pruned_words = pruned[:pruned_words]
+          @all_pruned = true
+        else
+          @search_results = @search_results.search(params[:subject]).where('subject ILIKE ?', "%#{params[:subject]}%")
+          @pruned_words = pruned[:pruned_words] if pruned[:pruned_words].any?
+        end
       end
     end
     @search_results = @search_results.complete if params[:completed].present?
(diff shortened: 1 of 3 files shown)


app/controllers/posts_controller.rb
```diff
@@ -358,6 +367,27 @@ def do_split
...
+    quoted = ActiveRecord::Base.connection.quote(query)
+    tsquery_result = ActiveRecord::Base.connection.select_value("SELECT plainto_tsquery('english', #{quoted})::text")
+
+    # Extract the lexemes from the tsquery result (e.g., "'nor' & 'phrase'" -> ["nor", "phrase"])
+    surviving_lexemes = (tsquery_result || '').scan(/'([^']+)'/).flatten
+
+    # A word is "pruned" if none of its stemmed forms appear in the surviving lexemes
+    pruned_words = words.select do |word|
+      word_quoted = ActiveRecord::Base.connection.quote(word)
+      word_tsvector = ActiveRecord::Base.connection.select_value("SELECT plainto_tsquery('english', #{word_quoted})::text")
```
Throne3d, line 384: Not loving doing a query per word - this doesn't run in parallel, so long search strings will end up doing a lot of queries in succession

Throne3d: Does ActiveRecord::Base.connection support array/map types of any kind? We could try executing it all in one go if so - I admit I have very little familiarity with how to run SQL queries directly in Rails, haha

Marri: Seems like we should be leveraging the official list of Postgres stop words, and comparing; it might miss stemming but it'd be much more efficient.

l1n: The current state of the PR already does this — these reviews predate a refactor that replaced the per-word `plainto_tsquery` calls with an in-memory Set lookup against Postgres's [english.stop](https://github.com/postgres/postgres/blob/master/src/backend/snowball/stopwords/english.stop) list. The current implementation:

```ruby
POSTGRES_ENGLISH_STOP_WORDS = Set.new(%w[
  i me my myself we our ours ourselves you your ...
]).freeze

def detect_pruned_words(query)
  words = query.split(/\s+/).compact_blank
  pruned_words = words.select { |word| POSTGRES_ENGLISH_STOP_WORDS.include?(word.downcase) }
  { pruned_words: pruned_words, all_pruned: pruned_words.length == words.length && words.any? }
end
```

Zero SQL queries, O(words) Set lookups — exactly what Marri suggested. Throne3d's "query per word, no parallelism" concern is also gone.

The trade-off Marri called out — it misses stemming, so `"running"` won't be detected as a pruned word even though `tsearch` would stem to `run` — is still the case. For the stop-word *warning* use case ("these words got dropped from your search") that's fine because tsearch doesn't actually prune stems; only the literal stop-word list members get dropped. Stems are kept and matched.

Resolvable from this end if it lines up with what you wanted.

Throne3d: That sounds OK to me! The stop list isn't too big.

I'll check out the current PR state.

Throne3d (approved):

PR #2737, opened 2026-04-02
Avoid breaking tags across pages when printing

miniBill:
Before:
image
image
image

After:
image
image
image


Talia-12:
Wow I was not expecting that to be a literal single line change, neat

this seems like an improvement to me, but I don't actually ever print so there's no way for it to break my workflow 😅

redmvc (approved):

Throne3d (approved):

PR #2753, opened 2026-05-11
Filter posts by visibility on continuity and section edit pages

l1n:
## Summary
- The edit-continuity page (`BoardsController#edit`) and the section edit page (`BoardSectionsController#edit`) listed posts directly from the board / section without applying the `visible_to` scope, so an editor could see the subjects of posts they didn't have read permission for (e.g. private or access-list posts authored by a coauthor).
- Pipe both collections through `.visible_to(current_user)` and render them via `@unsectioned_posts` / `@posts` in the views.

Also included on this branch: a small `bin/with-bundle` tooling improvement that surfaces the underlying error message when the active Ruby doesn't match the Gemfile pin. Happy to split it into its own PR if you'd prefer.

## Test plan
- [x] `bundle exec rspec spec/controllers/boards_controller_spec.rb`
- [x] `bundle exec rspec spec/controllers/board_sections_controller_spec.rb`
- [ ] Manually: as a continuity coauthor, open the edit-continuity and edit-section pages and confirm posts you don't have read access to are not listed.

### DIFF
--- a/app/controllers/api/v1/posts_controller.rb
+++ b/app/controllers/api/v1/posts_controller.rb
@@ -132,7 +132,7 @@ class Api::V1::PostsController < Api::ApiController
       end
     end
 
-    posts = Post.where(board_id: board.id, section_id: section_id)
+    posts = Post.where(board_id: board.id, section_id: section_id).visible_to(current_user)
     render json: { post_ids: posts.ordered_in_section.pluck(:id) }
   end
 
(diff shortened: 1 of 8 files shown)


teceler:
Does this handle reordering posts when there's one you can't see cleanly?

(Also wondering about the bundler changes)

l1n:
Point, will adjust. Bundler changes are because Claude Code on web has a hard time doing the rspec tests without a nudge since it has the wrong Ruby

Throne3d (approved):

PR #2754, opened 2026-05-16
Cache reply word counts in a column

l1n:
## Why

Rendering `/posts/:id/stats` is one of the most memory-heavy actions in the app. From per-request memory profiler data, single requests to `posts#stats` were spiking RSS by **+25 MB**, **+9 MB**, **+5.6 MB** on long posts — by far the largest per-request allocations of any action.

Looking at the implementation, `Post#total_word_count` does:

```ruby
contents = replies.pluck(:content)
full_sanitizer = Rails::Html::FullSanitizer.new
word_count + contents.inject(0) { |r, e| r + full_sanitizer.sanitize(e).split.size }.to_i
```

This loads every reply body into a giant `Array<String>` at once, then allocates two more string copies per reply (sanitized + split). On a long collab post that can be tens of MB. And `author_word_counts` then re-runs `word_count_for(author)` for **every** joined author, repeating the whole pluck + sanitize once per author.

## What

- New `replies.word_count` integer column (nullable; non-rewriting `ADD COLUMN` migration).
- `Reply` has a `before_save :cache_word_count, if: :content_changed?` callback that sets the column.
- `Writable#word_count` now prefers the stored column when present, falling back to the existing compute-on-the-fly behavior. `Post` has no column, so `post.word_count` is unchanged.
- `Post#total_word_count` and `Post#word_count_for` are rewritten to use `replies.sum(:word_count)` — pure SQL aggregates, zero Ruby allocations for reply bodies.
- `word_count` is added to the `audited ..., except:` list, so the cache update doesn't pollute the audit log.
- `lib/tasks/word_count.rake` adds `rake replies:backfill_word_count` which populates existing rows in batches of 1000 using `update_all` (no callbacks, no audit, no timestamps).

## Deploy steps

1. Merge + deploy. New replies start populating the column immediately.
2. `heroku run rake replies:backfill_word_count -a vast-journey-9935` to backfill historical replies. The task is idempotent (`where(word_count: nil)`) and can be re-run safely.
3. Until backfill completes, `total_word_count` will undercount on posts with un-backfilled replies (`SUM(NULL)` is treated as 0 by Postgres). After backfill, counts will match the previous implementation exactly.

## Notes

- Couldn't run the test suite locally — relying on CI.
- Added focused specs for the new caching behavior in `spec/models/reply_spec.rb` under the existing Writable describe block. The existing word_count specs in `spec/models/post_spec.rb` should keep passing unchanged.
- Considered also caching `Post#word_count` (the OP), but kept scope to replies since that's where the allocation bloat actually came from.


Throne3d (approved):

Throne3d:
Resolved safely, didn't need to do anything else

(Validated locally and this is correct, fortunately this migration is dated after the other one)

PR #2757, opened 2026-05-16
Preload the Rails app in Puma so workers share memory via CoW

l1n:
## Why

Without \`preload_app!\`, every Puma worker independently boots its own copy of the Rails app — no copy-on-write sharing between master and children. With \`WEB_CONCURRENCY=2\` each worker is ~270 MB resident, putting the dyno at ~540 MB before any request runs. Standard-1X's ceiling is 512 MB, which is exactly why R14 events are constant in the logs.

With \`preload_app!\`, the master boots Rails once and forks workers from it. The Linux kernel keeps unmodified pages shared between parent and child, so the bulk of the loaded framework — gem code, parsed class trees, the audited/pg_search/aws-sdk-s3 class graphs — is paid for once per dyno instead of once per worker. Expected drop: roughly 100–150 MB of resident RSS per web dyno at WEB_CONCURRENCY=2, which is enough to clear R14 on Standard-1X.

## What

- \`preload_app!\` in \`config/puma.rb\`.
- \`before_fork\` closes the master's \`\$redis\` connection (from \`config/initializers/redis.rb\`, which opens it eagerly) and clears the AR pool. Sharing a live TCP socket between forked processes corrupts both ends; closing before fork forces each worker to reconnect lazily on first use.
- \`Barnes.start\` moves from \`before_fork\` to \`on_worker_boot\`. Barnes runs a background thread that reports GC/process stats via statsd, and threads don't survive \`fork(2)\` — calling it in the master with preloading on means only the master gets monitored. Per-worker boot is where it belongs.

NewRelic and Scout auto-detect Puma preload and restart their own threads on worker boot, so no explicit hook is needed for either.

## Caveats / what to watch after deploy

- Verify with \`heroku ps -a vast-journey-9935\` that web dyno total RSS drops by ~100–150 MB and the R14 rate falls.
- Check that both APM dashboards (Scout + NewRelic) keep reporting per-worker after the deploy — if one goes quiet, that gem version probably doesn't auto-handle preload and needs an explicit \`on_worker_boot\` hook.
- The Resque worker dyno is unaffected (Procfile runs \`rake resque:work\` directly, not Puma). Its baseline stays the same.
- Couldn't validate locally; this is a real prod-behavior change, so worth watching dyno memory metrics for the first hour after merge.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -48,13 +48,35 @@ workers ENV.fetch("WEB_CONCURRENCY", 2)
 # into R14 at WEB_CONCURRENCY=2.
 preload_app!
 
-# Drop any connections the master may have opened (Redis namespace,
-# ActiveRecord pool) before workers fork. Sharing a live socket between
-# forked processes corrupts both ends; closing here forces each worker to
-# reconnect lazily on first use.
+# Drop any connections the master may have opened before workers fork.
+# Sharing a live socket between forked processes corrupts both ends;
+# closing here forces each worker to reconnect lazily on first use.
+#
+# Three Redis clients reach into prod:
+#   1. $redis (config/initializers/redis.rb) wraps REDIS_URL via
+#      Redis::Namespace and is handed to Resque.redis=. Eagerly built at
+#      app boot, so always live in the master.
+#   2. Rails.cache (config/environments/production.rb) is a
+#      RedisCacheStore pointed at REDIS_CACHE_URL.
+#   3. Rack::Attack.cache.store (config/initializers/rack_attack.rb) is
+#      another RedisCacheStore pointed at HEROKU_REDIS_TEAL_URL.
+#
+# (2) and (3) hold ConnectionPools whose connections only open on first
+# read/write. With preload_app! the master itself never serves a request
+# so those pools should be empty — but defensively drain them in case
+# any initializer pokes at the cache during boot.
 before_fork do
   ActiveRecord::Base.connection_handler.clear_active_connections! if defined?(ActiveRecord::Base)
   $redis&.close
+  [Rails.cache, (defined?(Rack::Attack) ? Rack::Attack.cache.store : nil)].compact.each do |store|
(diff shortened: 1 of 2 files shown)


config/puma.rb
```diff
@@ -39,9 +39,30 @@
...
+# into R14 at WEB_CONCURRENCY=2.
+preload_app!
+
+# Drop any connections the master may have opened (Redis namespace,
+# ActiveRecord pool) before workers fork. Sharing a live socket between
+# forked processes corrupts both ends; closing here forces each worker to
+# reconnect lazily on first use.
 before_fork do
+  ActiveRecord::Base.connection_handler.clear_active_connections! if defined?(ActiveRecord::Base)
+  $redis&.close
```
Marri, line 70: We've got three Redis floating around, do we need to close the other two here as well?

l1n: Good catch — three clients indeed. Pushed [1f4294db4](https://github.com/l1n/glowfic/commit/1f4294db4) to drain all of them. Tradeoffs:

1. **`$redis`** (REDIS_URL via Redis::Namespace) — eagerly built in the initializer, so it's definitely live in the master. Closed unconditionally.

2. **`Rails.cache`** (REDIS_CACHE_URL via RedisCacheStore) — ConnectionPool-backed; connections open lazily on first read/write. With `preload_app!` the master never serves a request, so this pool *should* be empty at fork time, but I'm draining it anyway since the cost is zero and the failure mode of inheriting a live socket across fork is bad.

3. **`Rack::Attack.cache.store`** (HEROKU_REDIS_TEAL_URL via RedisCacheStore) — same shape, same reasoning, same defensive drain.

The cache-store branch handles both shapes the underlying `.redis` call can return (`ConnectionPool::Wrapper` with `.shutdown`, or a plain Redis client with `.close`) since the API depends on the redis-rb version and pool config.

Couldn't actually test the fork-safety locally (needs the prod Redis topology + concurrent dynos), so this is "looks right, watch on deploy."

PR #2764, opened 2026-05-18
QoS: prioritize logged-in users when the dyno is saturated

l1n:
## Why

Tonight's outage had a clean signal in NewRelic: the rotated-UA scraper was **100% anonymous** (zero login attempts, zero non-GET methods, zero session attributes), and the cascade was sustained entirely by anonymous requests queueing past \`RACK_TIMEOUT_WAIT_TIMEOUT\`. Logged-in users were a small minority of the volume and an even smaller minority of the resulting 5xx pain. Splitting the rate limits + adding a load-shed layer means logged-in users get protected on both the steady-state ceiling AND on capacity during a queue blowup.

### Steady-state: differentiated throttles

The existing \`req/ip\` throttle counted both anonymous and logged-in traffic against the same per-IP bucket. Split into:

- **\`anon/ip\`** — anonymous IPs, keyed on \`req.ip\`, limit \`RACK_ATTACK_IP_LIMIT\` (existing env var, default 25). With #2760's shared cache store, this is the actual fleet-wide ceiling for anonymous scrape volume.
- **\`user\`** — logged-in users, keyed on user_id rather than IP so multiple users on a shared NAT (corporate, mobile carrier, household) don't fight for the same quota. Limit \`RACK_ATTACK_USER_LIMIT\` (new env var, default 1000) — generous enough that interactive use can never hit it.

### Saturation: load-shed anonymous traffic

\`app/middleware/anon_load_shed.rb\` reads \`env['rack.timeout.info'].wait\` (the seconds rack-timeout recorded the request as having spent in Puma's queue before reaching a worker). If wait exceeds 5 seconds AND the request is anonymous, return 503 with \`Retry-After: 30\` immediately instead of doing any controller work.

Effect under saturation:
- Anonymous requests get a fast 503 instead of sitting in queue to be served late or rack-timeout-aborted with a 500.
- Workers freed up serve the queued logged-in requests.
- The cascade can't snowball because each timed-out anonymous request that would have generated a retry now gets a clean signal to back off.

Under normal load this middleware is a no-op: queue wait is sub-100ms and the threshold isn't crossed. It only fires during genuine saturation, well above normal latency and well below \`RACK_TIMEOUT_WAIT_TIMEOUT=20\`.

## Combined effect on tonight's cascade (counterfactual)

The May 17 attacker was 100% anonymous. With these changes in place:

- The \`anon/ip\` throttle (once #2760 lands shared Redis storage) would have rate-limited the per-IP volume fleet-wide instead of per-worker. The scraper's per-IP rate would have been cut at the limit instead of letting 8 workers each allow 15 requests independently.
- AnonLoadShed would have started 503ing the bulk of the anonymous queue once queue depth crossed 5s, leaving the remaining worker capacity to serve real users. The visible outcome for logged-in users would have been "site feels slower" instead of "site is down."

## Notes

- Both layers depend on **PR #2760** (Rack::Attack on shared Redis store) to enforce cluster-wide. Without it, the throttles are per-worker as before. AnonLoadShed itself doesn't depend on shared storage — it reads rack-timeout state inline.
- **New env var:** \`RACK_ATTACK_USER_LIMIT\` (default 1000).
- Tests in \`spec/middleware/anon_load_shed_spec.rb\`.
- Couldn't run the suite locally; relying on CI.

## Open question for review

Whether to also extend this pattern with a behavioral IP killer (UA-rotation detection per IP + 5xx-tolerance detection) — that's a stacked follow-up I can put up as #2764 if the team wants it. The signal data from NR makes it look high-value: the attacker had exactly 8 distinct UAs rotating across many IPs, which is a fingerprint humans don't produce.


Throne3d:
I like this proposed mechanism and implementation. Checking with the team in case, but it seems good to me.

Throne3d:
One thought raised: this might shed genuine login attempts after 5s, which could (maybe) have passed after a long delay previously. Maybe we could also special-case those into the long queue? It opens us up to having a ton of load (and not shedding correctly) if someone spams our login path, but since our current "threat model" is about scraping, that doesn't seem like a problem.

Can we add a special-case to allow logins to bypass the shedding?

PR #2772, opened 2026-05-18
Enhance Indexes (perms, discoverability, pagination, flat view)

l1n:
Five small enhancements to the Indexes feature, surfaced from a quick audit. One commit per change so each can be reviewed (or reverted) independently.

## Commits

1. **Register `:edit_indexes` in `MOD_PERMS`** — `Index#editable_by?` calls `has_permission?(:edit_indexes)` for users editing someone else's `authors_locked: true` index, but the permission wasn't in `MOD_PERMS`. Mods couldn't moderate locked indexes (admins still passed via the early return). One-line fix.

2. **Show "+ New Index" to all full users** — the controller permits any non-read-only user to POST `/indexes`, but the index list view only rendered the create link for admins. Match the view to the controller. If the policy is meant to be admin-only the controller should enforce it; the inconsistency was just hiding a working feature.

3. **Drop redundant `dependent: :destroy` on `Index#posts`** — `has_many :posts, through: :index_posts, dependent: :destroy` is redundant with the direct `has_many :index_posts, dependent: :destroy` on the same model. Drop the through-target one; it invites the misreading that destroying an Index also destroys its Posts.

4. **Paginate sectionless posts on `indexes#show`** — `show` was passing `with_pagination: false` to `posts_from_relation`. For long-running indexes with hundreds of sectionless entries the page is a wall. Flip it on (default behavior) and render the standard paginator when `total_pages > 1`. Sections themselves are still rendered one-at-a-time without inter-section pagination since they're already curator-grouped.

5. **Add flat view for indexes** — `/indexes/:id?view=flat` renders a layout-stripped table of contents: index name + description, each section's name + description, each post (linked, with author shown). Mirrors the existing `posts/show` flat view; useful for saving/printing a curated reading list. Doesn't expand post bodies inline — linking to each post keeps the page tractable for large indexes.

## Audit results not in this PR

Two items I expected to find turned out to already be done:
- IndexSection already has a `description` column, form, and show rendering.
- Sectionless and sectioned posts are already privacy-filtered via `posts_from_relation` → `posts.visible_to(current_user)`.

Bigger features I scoped out:
- Bulk import (add all posts from a board/section/tag to an index) — feature ask, separate PR.
- Within-section pagination — felt wrong to paginate something the curator deliberately ordered.

## Test plan
- [ ] CI rspec suite passes
- [ ] Manual: as a mod (non-admin), edit a `authors_locked: true` index owned by another user — should work
- [ ] Manual: as a full non-admin user on `/indexes`, see "+ New Index" link
- [ ] Manual: create an index with >25 sectionless posts, see paginator on `show`
- [ ] Manual: `/indexes/:id?view=flat` renders a clean table-of-contents page

### DIFF
--- a/app/controllers/indexes_controller.rb
+++ b/app/controllers/indexes_controller.rb
@@ -40,9 +40,16 @@ class IndexesController < ApplicationController
     end
 
     @page_title = @index.name.to_s
-    @sectionless = @index.posts.where(index_posts: { index_section_id: nil })
-    @sectionless = @sectionless.ordered_by_index
     dbselect = ', index_posts.description as index_description, index_posts.id as index_post_id'
+    @sectionless = @index.posts.where(index_posts: { index_section_id: nil }).ordered_by_index
+
+    if params[:view] == 'flat'
+      response.headers['X-Robots-Tag'] = 'noindex'
+      @sectionless = posts_from_relation(@sectionless, with_pagination: false, select: dbselect)
+      render :flat, layout: false
+      return
+    end
+
     @sectionless = posts_from_relation(@sectionless, select: dbselect)
   end
 
--- a/app/controllers/indexes_controller.rb
+++ b/app/controllers/indexes_controller.rb
@@ -43,7 +43,7 @@ class IndexesController < ApplicationController
     @sectionless = @index.posts.where(index_posts: { index_section_id: nil })
     @sectionless = @sectionless.ordered_by_index
     dbselect = ', index_posts.description as index_description, index_posts.id as index_post_id'
-    @sectionless = posts_from_relation(@sectionless, with_pagination: false, select: dbselect)
+    @sectionless = posts_from_relation(@sectionless, select: dbselect)
   end
 
   def edit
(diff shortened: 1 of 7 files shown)


app/models/index.rb
```diff
@@ -3,7 +3,7 @@ class Index < ApplicationRecord
   include Concealable
 
   has_many :index_posts, inverse_of: :index, dependent: :destroy
-  has_many :posts, through: :index_posts, dependent: :destroy
+  has_many :posts, through: :index_posts
```
teceler, line 6: The dependent destroy is deliberate, we have that on a lot of through associations, iirc because it makes rails do callbacks and update relations properly

app/controllers/indexes_controller.rb
```diff
@@ -40,10 +40,17 @@ def show
...
+    @sectionless = @index.posts.where(index_posts: { index_section_id: nil }).ordered_by_index
+
+    if params[:view] == 'flat'
+      response.headers['X-Robots-Tag'] = 'noindex'
+      @sectionless = posts_from_relation(@sectionless, with_pagination: false, select: dbselect)
+      render :flat, layout: false
+      return
+    end
+
+    @sectionless = posts_from_relation(@sectionless, select: dbselect)
```
teceler, line 53: Does this paginate properly if there are both sectionless posts and sections?

PR #2782, opened 2026-06-15
Give search autocomplete API GETs a separate rate-limit bucket

l1n:
The flat per-IP throttle counted every request, including the /api/v1 autocomplete calls the search forms fire via select2. With no minimum input length, those dropdowns request on open and on each keystroke, so a single search fans out into dozens of requests and blows the 25/5min limit that is otherwise fine for normal browsing (~1 request per page).

Exclude read-only /api/v1 GETs from the general req/ip throttle and count them under a separate, more generous api/ip bucket (default 150/5min, configurable via RACK_ATTACK_API_LIMIT).

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -16,10 +16,27 @@ ENV.fetch("RACK_ATTACK_BAD_IP", "").split(",").compact_blank.each { |ip| Rack::A
 url = ENV.fetch("HEROKU_REDIS_TEAL_URL", nil)
 Rack::Attack.cache.store = ActiveSupport::Cache::RedisCacheStore.new(url: url, ssl_params: { verify_mode: OpenSSL::SSL::VERIFY_NONE }) if url
 
-# Throttle all requests by IP (60rpm)
+# Read-only API GETs power the search forms' autocomplete dropdowns (select2).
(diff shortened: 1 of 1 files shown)


config/initializers/rack_attack.rb
```diff
@@ -16,10 +16,27 @@
...
 Rack::Attack.cache.store = ActiveSupport::Cache::RedisCacheStore.new(url: url, ssl_params: { verify_mode: OpenSSL::SSL::VERIFY_NONE }) if url
 
-# Throttle all requests by IP (60rpm)
+# Read-only API GETs power the search forms' autocomplete dropdowns (select2).
+# A single search interaction fans out into many of these requests - one per
+# dropdown opened and one per keystroke - which is fundamentally different from
+# normal page browsing (~1 request per navigation). They get their own, more
+# generous bucket below so a search doesn't exhaust the general per-IP limit.
+def autocomplete_api_request?(req)
+  req.get? && req.path.start_with?('/api/v1/')
```
Throne3d, line 25: I guess this isn't exactly the category for autocomplete drop-downs, but it's close enough and I don't think we've noticed bots hitting these API endpoints too much - so seems fair!

Throne3d (approved):

PR #2792, opened 2026-06-21
Add DB connection resilience for maintenance/failover events

l1n:
## Why

After a Postgres maintenance/failover, the long-lived connections held in each Puma worker's pool die server-side but aren't proactively recycled. With no reaping/timeout config, dead sockets linger in the pool and get handed to requests (erroring until the pool cycles), so a **manual app restart** has been needed to pick up fresh connections — especially when a failover changes the endpoint IP behind the hostname (a restart forces fresh DNS resolution).

## What

- **`config/database.yml`** (production):
  - `reaping_frequency` — reaper thread prunes dead pooled connections instead of waiting until they're next checked out
  - `connect_timeout` / `checkout_timeout` — fail fast against a dead/changed endpoint rather than hanging
  - All three env-tunable (`DB_REAPING_FREQUENCY`, `DB_CONNECT_TIMEOUT`, `DB_CHECKOUT_TIMEOUT`), defaulting to 10s / 5s / 5s
- **`config/puma.rb`**: `on_worker_boot` re-establishes Active Record connections per forked worker, mirroring the existing Resque fork hooks in `config/initializers/resque.rb`

## Notes

Config-only; no behavior change in normal operation. Intent is to let the app recover from DB maintenance without a manual restart.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -20,5 +20,12 @@ production:
   <<: *default
   database: db/production.sqlite3
   pool: <%= ENV["DB_POOL"] || ENV['RAILS_MAX_THREADS'] || 5 %>
+  # Prune dead connections from the pool so a DB maintenance/failover event
+  # doesn't leave stale sockets to be handed out until they're next used.
+  reaping_frequency: <%= ENV["DB_REAPING_FREQUENCY"] || 10 %>
+  # Fail fast against a dead/changed endpoint instead of hanging on connect
+  # or waiting the full default for a pool checkout.
+  connect_timeout: <%= ENV["DB_CONNECT_TIMEOUT"] || 5 %>
+  checkout_timeout: <%= ENV["DB_CHECKOUT_TIMEOUT"] || 5 %>
   variables:
     statement_timeout: <%= ENV["STATEMENT_TIMEOUT"] %>
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -45,6 +45,14 @@ before_fork do
   Barnes.start
 end
 
+# Each clustered worker is forked from the master, so re-establish its own
+# Active Record connections on boot rather than inheriting the master's.
+# Belt-and-suspenders against stale connections after a DB maintenance event
+# (mirrors what config/initializers/resque.rb does on fork).
+on_worker_boot do
+  ActiveRecord::Base.establish_connection if defined?(ActiveRecord::Base)
+end
+
 # Allow puma to be restarted by `bin/rails restart` command.
 plugin :tmp_restart
 


config/database.yml
```diff
@@ -20,5 +20,12 @@ production:
   <<: *default
   database: db/production.sqlite3
   pool: <%= ENV["DB_POOL"] || ENV['RAILS_MAX_THREADS'] || 5 %>
+  # Prune dead connections from the pool so a DB maintenance/failover event
+  # doesn't leave stale sockets to be handed out until they're next used.
+  reaping_frequency: <%= ENV["DB_REAPING_FREQUENCY"] || 10 %>
```
Throne3d, line 25: Seems this [defaults to 60](https://github.com/rails/rails/blob/v8.0.4/activerecord/lib/active_record/database_configurations/hash_config.rb#L98) in our version (8.0), though it's interesting that it's been removed from the docs for Rails 8.1 (and had its default reduced). https://github.com/rails/rails/commit/3e3f97e6c4be1ab5f28c73a380bb6caf0125d4ee

We can adjust it to be lower, but I'm pretty sure this just makes recovery happen _faster_ - it doesn't actually fix an underlying issue.

config/database.yml
```diff
@@ -20,5 +20,12 @@ production:
...
+  # Fail fast against a dead/changed endpoint instead of hanging on connect
+  # or waiting the full default for a pool checkout.
+  connect_timeout: <%= ENV["DB_CONNECT_TIMEOUT"] || 5 %>
+  checkout_timeout: <%= ENV["DB_CHECKOUT_TIMEOUT"] || 5 %>
```
Throne3d, line 29: [Defaults to 5 already](https://github.com/rails/rails/blob/v8.0.4/activerecord/lib/active_record/database_configurations/hash_config.rb#L93) - I'm OK making it configurable, but won't do anything as-is.

config/database.yml
```diff
@@ -20,5 +20,12 @@ production:
   <<: *default
   database: db/production.sqlite3
   pool: <%= ENV["DB_POOL"] || ENV['RAILS_MAX_THREADS'] || 5 %>
+  # Prune dead connections from the pool so a DB maintenance/failover event
+  # doesn't leave stale sockets to be handed out until they're next used.
+  reaping_frequency: <%= ENV["DB_REAPING_FREQUENCY"] || 10 %>
+  # Fail fast against a dead/changed endpoint instead of hanging on connect
+  # or waiting the full default for a pool checkout.
+  connect_timeout: <%= ENV["DB_CONNECT_TIMEOUT"] || 5 %>
```
Throne3d, line 28: Fair - seems this defaults to infinite with libpq(?).

config/puma.rb
```diff
@@ -45,6 +45,14 @@
   Barnes.start
 end
 
+# Each clustered worker is forked from the master, so re-establish its own
+# Active Record connections on boot rather than inheriting the master's.
+# Belt-and-suspenders against stale connections after a DB maintenance event
+# (mirrors what config/initializers/resque.rb does on fork).
+on_worker_boot do
```
Throne3d, line 52: ✅

Throne3d:
I think the main impacts here are in connect_timeout and the Puma config change. I don't think this will solve the underlying issue of needing an app restart - the main goal of reaping is apparently already enabled by default in Rails, so something else must have gone wrong here.

PR #2793, opened 2026-06-21
Custom CSS skins

l1n:
## Summary

User-authored **custom CSS skins**: themes a user can write, apply site-wide to their own browsing, recommend on a post, or share publicly in a gallery — gated by a CSS sanitizer and a moderator-approval workflow.

> **Split per review (@teceler's suggestion):** this PR originally also bundled the reply-layout features (alternating icons + Pesterchum). Those are now their own PR, #2804 — the two features are independent. This PR is just the skins feature.

## Custom CSS skins

- **`Skin` model**: user-authored CSS with a name, optional description, and a public/private flag. Audited and versioned.
- **`Glowfic::CssSanitizer`**: a security boundary (skins restyle the whole page and can be shared), not a cosmetic nicety. It allowlists safe properties, strips `!important` so the app's own safety overrides always win, blocks external resources (`@import`, `@font-face`, non-`data:` `url(...)`) to close CSS's finer attribute-selector / `@media` exfiltration, blocks clickjacking/script vectors (`position: fixed/sticky`, `expression()`, `-moz-binding`, `behavior`), and discards anything it cannot confidently parse. Anything it drops for a security reason flags the skin for review.
- **Two-tier serving** (`Skin#css_for`): the owner always sees their own raw CSS; everyone else gets the sanitized version unless a moderator has approved that exact CSS. Approval lapses automatically when the CSS is edited (tracked by a content digest).
- **Served as a standalone stylesheet** (`SkinsController#css`, linked via `<link>`): a skin is delivered as a `text/css` file, scoped under `:root:root` so it out-ranks the app's own theming without `!important`, rather than being inlined into every page's `<style>`. There's no markup context to break out of, and the browser caches it across page loads (ETag + private cache).
- **Moderator review** (`approve_skins` permission): a skin that uses "dangerous" (stripped) CSS and is shared publicly or recommended on a post is routed to a review queue where a mod can **Approve** (serve raw CSS to readers) or **Reject** (unlist).
- **Where skins apply**:
  - *Your active skin* — applied site-wide to your own browsing (`User#active_skin`).
  - *Post recommended skin* — a post author can recommend one of their skins; readers see it (opt-out) unless they've turned it off globally (`hide_skins`) or for that view (`?skin=off` / `?skin=on`).
- **Safety overrides** (`Skin::SAFETY_OVERRIDES`): a trailing `!important` block keeps critical chrome (content warnings, flashes, the ToS gate) visible and un-overlaid no matter what a skin tries — pinning display/visibility/opacity, position/transform (and the independent `scale`/`rotate`/`translate`), filter, clip, and pointer-events to their no-ops.
- **Gallery & built-ins**: public skins are browsable and forkable ("Copy to mine") from a gallery. The bundled layouts are seeded as pre-approved, forkable public skins via the `skins:seed_builtin` rake task (`Glowfic::BuiltinSkins`).

### Routes / migrations

- `resources :skins` with member `css`/`use`/`fork`/`approve`/`reject` and collection `gallery`/`review`/`clear`.
- Migrations: `create_skins`, `add_skin_references` (`posts.skin_id`, `users.skin_id`, `users.hide_skins`), `add_skin_approval`.

## Tests

New specs: `Glowfic::CssSanitizer`, `Glowfic::BuiltinSkins`, `Skin` model, `SkinsController` (including the `css` stylesheet endpoint), plus additions to the posts/users controller and application-helper specs.

## Screenshots

| Your Skins | Gallery |
| --- | --- |
| Your skins | Skin gallery |

| New skin (editor) | Skin detail |
| --- | --- |
| New skin form | Skin detail |

A post with a recommended reading skin applied (reader view, with the "Turn off" toggle):

Post with recommended skin

<sub>Screenshots are hosted on the `pr2793-screenshots` branch so they don't bloat the PR diff; that branch can be deleted once the PR is merged.</sub>


miniBill:
> **Alternating icons** (Account Settings → _Reply layout_): a compact layout where reply icons alternate left/right so rapid back-and-forth dialogue is easier to follow. Implemented as an `alternating-icons` body class; `replies/_single.haml` alternates an `icon-left`/`icon-right` class per reply via `reply_counter`.

Couldn't this be done with CSS only?

> **Dynamic icon sizing** (`posts/alternating_icons.js`, loaded only when the layout is active): each reply's icon is scaled to its own body height, clamped to 28–96px.

Same question (this is probably harder and possibly not worth it)

app/assets/javascripts/posts/alternating_icons.js
```diff
@@ -0,0 +1,48 @@
...
+  const sizeIcons = function() {
+    const posts = document.querySelectorAll('#content .post-container');
+    Array.from(posts).forEach(function(post) {
+      const icon = post.querySelector('.post-icon img, .post-icon div');
+      const content = post.querySelector('.post-content');
+      if (!icon || !content) return;
+
+      const size = clamp(MIN, content.offsetHeight, CAP);
+      icon.style.width = size + 'px';
+      icon.style.height = size + 'px';
```
miniBill: What happens to non-square icons?

app/helpers/application_helper.rb
```diff
@@ -2,6 +2,49 @@
...
+  SKIN_SAFETY_OVERRIDES = <<~CSS
+    #{SKIN_SCOPE} .flash, #{SKIN_SCOPE} .flash.error, #{SKIN_SCOPE} .flash-margin, #{SKIN_SCOPE} #tos {
+      display: block !important;
+      visibility: visible !important;
+      opacity: 1 !important;
+      position: static !important;
+      height: auto !important;
+      max-height: none !important;
+      overflow: visible !important;
+      transform: none !important;
```
miniBill: Would this override a `scale : 0` too?

app/helpers/application_helper.rb
```diff
@@ -2,6 +2,49 @@
...
+      clip: auto !important;
+      clip-path: none !important;
+      pointer-events: auto !important;
```
miniBill: In general, how was this list generated?

app/helpers/application_helper.rb
```diff
@@ -2,6 +2,49 @@
...
+  # element into markup/script.
+  def skin_style_tag(skin, viewer: current_user)
+    return if skin.nil?
+
+    css = skin.css_for(viewer)
+    return if css.blank?
+
+    scoped = Glowfic::CssSanitizer.scope(css, SKIN_SCOPE)
+    payload = "#{scoped}\n#{SKIN_SAFETY_OVERRIDES}".gsub('</', '<\/')
+    tag.style(payload.html_safe, type: 'text/css')
```
miniBill: Could we serve the skin from a separate controller instead of embedding it in responses? This would avoid the need for the `gsub` which is a denylist approach to escaping the tag, whereas "a separate file" is just... not something you can break out of

db/migrate/20260621020000_add_skin_references.rb
```diff
@@ -0,0 +1,15 @@
+class AddSkinReferences < ActiveRecord::Migration[8.0]
+  def change
+    change_table :users, bulk: true do |t|
+      # A user's active personal skin, applied to their own browsing site-wide.
+      t.integer :skin_id
+      # Reader opt-out of post-recommended skins.
+      t.boolean :hide_skins, default: false
```
miniBill, line 7: Are we sure we want this to be an opt-out? It's convenient to be opt-out, because people can experience custom skins with minimal friction, but is it what we want for users?

lib/glowfic/css_sanitizer.rb
```diff
@@ -0,0 +1,301 @@
+# frozen_string_literal: true
+
+# Sanitizes user-authored "skin" CSS before it is injected into a page.
+#
+# Skins are allowed to restyle the whole page and can be shared publicly, so
+# this is a security boundary, not a cosmetic nicety. The rules enforced here:
```
miniBill, line 6: ~~Gah~~

lib/glowfic/css_sanitizer.rb
```diff
@@ -0,0 +1,301 @@
...
+#
+#   * Only allowlisted properties survive (unknown/dangerous ones are dropped).
+#   * `!important` is stripped from every declaration, so the application can
+#     always win a specificity battle with its own trailing safety overrides
+#     (e.g. keeping content warnings visible).
+#   * External resources are blocked: `@import`, `@font-face`, and any `url(...)`
+#     that is not a `data:` URI are dropped. This neutralises CSS-based tracking
```
miniBill: Good idea in general but... we allow hotlinking images, external-resource-based tracking is kinda... already possible right now? Maybe with CSS you get to do finer stuff (using `@media` and friends), but let's not hide our heads in the sand

lib/glowfic/css_sanitizer.rb
```diff
@@ -0,0 +1,301 @@
...
+    animation animation-name animation-duration animation-timing-function
+    animation-delay animation-iteration-count animation-direction
+    animation-fill-mode animation-play-state
+    filter backdrop-filter
+
+    table-layout caption-side empty-cells
+    object-fit object-position
+    cursor user-select pointer-events
+    columns column-count column-width column-rule column-fill
+    column-span
```
miniBill, line 101: How was this list generated?

lib/glowfic/css_sanitizer.rb
```diff
@@ -0,0 +1,301 @@
...
+    sanitizer.sanitized
+    sanitizer.dangerous?
+  end
+
+  # Prefix every style-rule selector with `scope` (a descendant prefix such as
+  # ":root:root") so an injected skin out-ranks the application's own theming
+  # rules — many of which use :nth-child or #id selectors that a bare ".x" skin
+  # selector cannot beat — without relying on !important, which the sanitizer
+  # strips for non-owners. Keyframe selectors (from/to/percent) are never scoped.
+  # Falls back to returning the input unchanged if it cannot be parsed.
```
miniBill: This should probably fail closed, not open. That said, when asking for the sanitized version it needs to be accepted by the parser, so it's probably not a security issue?

lib/glowfic/css_sanitizer.rb
```diff
@@ -0,0 +1,301 @@
...
+    render_nodes(Crass.parse(@css)).strip
+  rescue StandardError, SystemStackError
+    # Never let malformed or pathological input bubble up as a usable stylesheet.
+    ''
+  end
+
+  # True if sanitizing dropped or defused a security-relevant construct. Only set
+  # after #sanitized has run. Unsupported-but-harmless properties do not count.
+  def dangerous?
+    @dangerous
```
miniBill, line 215: I really dislike that this is not (correct by construction). Could we instead set @dangerous in `initialize`?

lib/glowfic/css_sanitizer.rb
```diff
@@ -0,0 +1,301 @@
...
+  def render_node(node)
+    case node[:node]
+      when :style_rule then render_style_rule(node)
+      when :at_rule    then render_at_rule(node)
+    end
+  end
+
+  def render_style_rule(node)
+    selector = node.dig(:selector, :value).to_s.strip
+    return if selector.empty? || dangerous_value?(selector)
```
miniBill: This doesn't set `@dangerous`

lib/glowfic/css_sanitizer.rb
```diff
@@ -0,0 +1,301 @@
...
+  def render_at_rule(node)
+    name = node[:name].to_s.downcase
+    prelude = Crass::Parser.stringify(node[:prelude]).strip
+
+    # @import / @font-face pull in external resources; flag and drop them.
+    # (@import has no block, so this must run before the block check below.)
+    @dangerous = true if %w[import font-face].include?(name)
+
+    return if node[:block].nil?
+    return if dangerous_value?(prelude)
```
miniBill: This doesn't set `@dangerous`

lib/glowfic/css_sanitizer.rb
```diff
@@ -0,0 +1,301 @@
...
+
+      sanitize_declaration(decl)
+    end
+  end
+
+  def sanitize_declaration(decl)
+    name = decl[:name].to_s.strip.downcase
+    value = decl[:value].to_s.strip
+    return if name.empty? || value.empty?
+    return unless property_allowed?(name) # unsupported, but not dangerous
```
miniBill, line 282: This doesn't set `@dangerous`. I see that it's intentional but I don't like it

lib/glowfic/css_sanitizer.rb
```diff
@@ -0,0 +1,301 @@
...
+      return
+    end
+    # Angle brackets are never needed in a declaration value, and allowing them
+    # (e.g. content: "</style><script>...") would let a skin break out of the
+    # inline <style> element it is injected into.
+    if value.match?(/[<>]/)
+      @dangerous = true
+      return
+    end
+    if name == 'position' && value.match?(/\b(?:fixed|sticky)\b/i)
```
miniBill, line 294: This is a denylist, should probably be an allowlist instead

app/assets/javascripts/posts/alternating_icons.js
l1n: They're not distorted: `replies.scss` sets `object-fit: contain` on these icons, so a non-square image is letterboxed inside the square box — the same way glowfic presents icons everywhere else. Added a comment in the JS noting this.

app/helpers/application_helper.rb
l1n: Good catch — it didn't. The override pinned `transform: none` but the independent transform properties (`scale`/`rotate`/`translate`) bypass that, so `scale: 0` on a flash would've slipped through. Now pinning those three plus `filter` to their no-op as well.

app/helpers/application_helper.rb
l1n: Documented it (and moved the constant to `Skin::SAFETY_OVERRIDES` as part of the controller change below). It's the set of ways CSS can hide / shrink-to-nothing / move off-screen / clip / obscure / de-interact an element, each pinned to its no-op value: display·visibility·opacity, height·max-height·overflow, position·transform·scale·rotate·translate, clip·clip-path, filter, pointer-events.

app/helpers/application_helper.rb
l1n: Done. Skins are now served from `SkinsController#css` as a standalone `text/css` file referenced via `<link>`, instead of being embedded in an inline `<style>`. That removes the `</`-escaping `gsub` entirely (there's no markup context to break out of), and lets the browser cache the stylesheet across page loads (ETag + private cache) rather than re-sending it inline on every HTML response. Access mirrors the previous exposure: a skin you can see, or one recommended on a post (served to that post's readers regardless of the skin's own privacy); anything else 404s.

db/migrate/20260621020000_add_skin_references.rb
l1n: This is the one genuinely product-level call in the review. Opt-out maximizes discovery — readers experience a post's recommended skin with zero setup — but I take the point that defaulting people into rendering others' CSS is a different bar than the rest. Flagging it for @l1n to decide; happy to flip it to opt-in if that's preferred.

lib/glowfic/css_sanitizer.rb
l1n: Fair — updated the comment to concede that we already allow image hotlinking, so this isn't about coarse "did someone load the page" tracking. What it actually buys is blocking the finer-grained exfiltration CSS uniquely enables: an attribute-selector + `background: url(...)` that fires a request keyed on page content, and `@media`/value-conditional requests that leak viewport/device characteristics. Keeping skins self-contained closes those, so I kept the block.

lib/glowfic/css_sanitizer.rb
l1n: Documented the provenance: started from the visual/layout property groups in MDN's CSS properties reference, then removed every property whose *value* can load a resource or run script (those are gated by `dangerous_value?` regardless) and the ones that are escape hatches on their own. Anything not listed is dropped as unsupported-but-harmless, so growing the list is the safe way to add theming power.

lib/glowfic/css_sanitizer.rb
l1n: Done — `scope` now returns `''` (drops the skin) on a parse error instead of emitting it unscoped. You're right that it isn't strictly a security issue (the input here is already sanitized and has to parse), but failing closed is the right default.

lib/glowfic/css_sanitizer.rb
l1n: Fixed the concrete gaps you flagged just below (dangerous selector / prelude now set it). A fully correct-by-construction version — each sanitize step returning a result that carries the danger verdict instead of mutating an ivar — is a larger refactor I left for now, but the behavior is now correct on every drop-for-security path.

lib/glowfic/css_sanitizer.rb
l1n: Fixed — dropping a rule for a dangerous selector now sets `@dangerous`, so such a skin is still routed to mod review.

lib/glowfic/css_sanitizer.rb
l1n: Fixed likewise — a dropped at-rule with a dangerous prelude now sets `@dangerous`.

lib/glowfic/css_sanitizer.rb
l1n: Left this one as-is: an unsupported-but-allowlisted-out property genuinely isn't dangerous, it's just unused, so flagging it would route perfectly benign skins to review. Kept the comment noting it's intentional.

lib/glowfic/css_sanitizer.rb
l1n: The property list above *is* an allowlist; this is the value check, and CSS values are effectively unbounded, so a value allowlist isn't practical to keep complete/correct. The layered defense is: property allowlist + this value denylist for the few known-dangerous tokens + `:root:root` scoping + `!important` strip + the safety overrides. I'd keep the value side as a denylist within that stack rather than try to enumerate all safe values.

db/migrate/20260621020000_add_skin_references.rb
l1n: Decision from @l1n: keep it opt-out, since there's already a per-user setting to make it opt-in. The `hide_skins` user setting ("Don't apply posts' recommended skins", in account settings) lets any reader turn off recommended skins by default; they can still enable a specific post's skin with the "Use it" link (`?skin=on`). And the CSS served to non-owners is sanitized + mod-gated when dangerous, so the default-on path is bounded. So: opt-out by default, opt-in available per user. No change.

teceler:
This is too many things for one pr tbh, can it be split up?

l1n:
Per @teceler's suggestion, I've split this PR: it's now **just the custom CSS skins** feature. The reply-layout work (alternating icons + Pesterchum / Pesterchum Memo) is now its own PR, #2804. The two features are independent, so they can be reviewed and merged in either order. The skins review threads above are unaffected and all addressed.

lib/glowfic/css_sanitizer.rb
miniBill: Fair enough

PR #2794, opened 2026-06-21
Speed up CI: parallel_tests, MinIO cache, faster system specs

l1n:
## Why

The `rspec` job is the entire CI bottleneck — ~8.5 min, single-threaded — while every other job finishes in under a minute. Of that, the ~143 system/Capybara specs (≈5% of examples) dominate the wall-clock.

## What

**Parallelize the suite** across 4 `parallel_tests` workers (`parallel_rspec`), each with its own Postgres database via `TEST_ENV_NUMBER`, split by recorded per-file runtime (cached). The parallel run lands around **~3 min vs ~8.5 min**.

**Merge coverage across workers**: each worker writes to `coverage/parallel/<n>`; a final `rake coverage:report` collates them and enforces the existing `line: 95.2, branch: 88.1` gate on the merged total (the per-process gate is skipped under parallel runs, since a worker only sees a slice).

**Cache the MinIO binary** instead of re-downloading ~50 MB every run.

### Test-isolation fixes (required for parallel)

Parallel regrouping of spec files per worker surfaced latent order-dependencies that the old fixed single-process order quietly hid. These are real isolation bugs, fixed at the root (not papered over with retries):

- **ActiveJob**: pin the `:test` adapter for the whole test env (it was only switched ad hoc by specs that `include ActiveJob::TestHelper`), and reset queued/performed jobs + the `perform_enqueued_jobs` flag before each example.
- **Redis**: `parallel_tests` gives each worker its own DB but they share one Redis server, so keys collided cross-worker and `flushdb` wiped siblings. Each worker now gets its own Redis namespace (`$redis` and the cache store), and the per-example reset clears only the worker's own keys. This also fixes the real leak behind the failures: `GenerateFlatPostJob.enqueue` takes a 30-min Redis lock released only when the job performs, so in tests the lock leaked and silently no-op'd later enqueues (`have_been_enqueued` saw 0).
- **Auditing**: reset `auditing_enabled` to its default before each example — specs that flip it on inline leaked the flag if they raised before flipping it back, corrupting later audit counts.
- **MinIO bucket**: make `create_bucket`-on-boot idempotent, since the N concurrent Rails boots for the parallel DB setup raced on it (also correct for sibling Puma workers).

### Other

- `Capybara.disable_animation = true` (small reliability win). I measured the headless-Chrome perf flags I first tried and they made no per-spec difference, so they were dropped.

## Notes

- Locally validated: parallel split + coverage collate work, and the non-system suite is stable across repeated parallel runs. (Set `CI=true` to match CI's eager-loading when reproducing locally — otherwise a couple of specs reference constants defined in oddly-named files and fail on autoload order, which can't happen with eager load on.)
- The HTML rspec report formatter was dropped (can't be shared across parallel workers); the coverage artifact is still uploaded.


Gemfile
```diff
@@ -84,15 +84,18 @@ group :development, :test do
...
   gem 'rspec-rails'
   gem 'seed_dump', '~> 3.4'
+  gem 'test-prof'
 end
 
 group :test do
   gem 'capybara'
   gem 'factory_bot_rails'
   gem 'rails-controller-testing'
+  gem 'rspec-github', require: false
```
Throne3d, line 98: Oh neat 👀

spec/controllers/admin/posts_controller_spec.rb
```diff
@@ -32,7 +32,7 @@
   end
 
   describe "POST #do_regenerate" do
-    before(:each) { create_list(:post, 10) }
+    let_it_be(:regen_posts) { create_list(:post, 10) }
```
Throne3d, line 35: Neat, haven't used this before but makes sense

(https://github.com/test-prof/test-prof/blob/master/docs/recipes/let_it_be.md)

spec/rails_helper.rb
```diff
@@ -35,6 +36,38 @@
...
+    end
+
+    # 2. Redis — e.g. GenerateFlatPostJob.enqueue takes a 30-minute Redis lock and
+    #    only releases it when the job performs; in tests jobs are enqueued but not
+    #    performed, so the lock would leak and silently no-op a later enqueue.
+    #    Clear only this worker's namespace (NOT flushdb — parallel workers share
+    #    the Redis server) and the cache (rate-limit buckets etc.).
+    if $redis
+      keys = $redis.keys('*')
+      $redis.del(*keys) unless keys.empty?
```
Throne3d, line 61: I guess this isn't a very big list, so it's probably safe (used to having to use `scan`)

lib/tasks/coverage.rake
```diff
@@ -0,0 +1,19 @@
...
+    result_files = Dir['coverage/parallel/*/.resultset.json']
+    if result_files.empty?
+      abort 'No parallel coverage results found under coverage/parallel/*. Did the test run write coverage?'
+    end
+
+    # Collate the per-worker results into one report and apply the same gate the
+    # single-process run uses (see spec/spec_helper.rb). Exits non-zero if below.
+    SimpleCov.collate(result_files, 'rails') do
+      enable_coverage :branch
+      minimum_coverage line: 95.2, branch: 88.1
```
Throne3d, line 17: Should share these constants somewhere, but not sure where the best place is - gonna leave a note to keep these in sync

```suggestion
      # NB: keep in sync with spec/spec_helper.rb minimum_coverage
      minimum_coverage line: 95.2, branch: 88.1
```

spec/spec_helper.rb
```diff
@@ -49,7 +59,7 @@ def coverage_exceeding_source_warn
       end
     end
     enable_coverage :branch
-    minimum_coverage line: 95.2, branch: 88.1
+    minimum_coverage line: 95.2, branch: 88.1 unless parallel_worker
```
Throne3d, line 63:
```suggestion
    # NB: keep in sync with lib/tasks/coverage.rake minimum_coverage
    minimum_coverage line: 95.2, branch: 88.1 unless parallel_worker
```

Throne3d (approved):

PR #2796, opened 2026-06-22
Add manual "keep my place" bookmark mode for the unread marker

l1n:
Adds an opt-in per-account setting (manual_unread) that turns the unread marker into a deliberate bookmark instead of an auto-advancing "best guess" of where you left off:

- The marker no longer auto-advances as you read; it only moves when you explicitly mark a spot (e.g. "Mark Unread Here"). The first view of a post still records a baseline so unread tracking works.
- Threads load anchored at the marker (lands on the marker's page and scrolls it to the top) without needing the #unread hash.
- The marked reply is highlighted with a distinct amber tint + left accent, visually separate from the red "jumping to unread" outline.
- Threads containing a marker get a distinct favicon on the tab.

Includes a migration for the new column, settings UI, and specs covering the no-auto-advance behavior, the favicon flag, and the rendered highlight/anchor.

Claude-Session: https://claude.ai/code/session_01HbT92gmGTizsc5bXWQcdMs


teceler:
Maybe redundant with https://github.com/glowfic-constellation/glowfic/pull/2814

PR #2801, opened 2026-06-28
Disallow SVG uploads (#1655)

### LINKED ISSUE #1655: Disallow svg uploads
Report from ilzo:
> So, I did a test of uploading an SVG icon (is deleted now) and what seemed to happen was that the icon got shown as a PNG but when I went to open the image in a new tab I got weird failures. This does not seem, like, urgent or important, but I thought I'd let people know that you might want to either poke at that or turn off SVG uploads or something.

compwron:
## Problem
Fixes #1655. SVG files can be uploaded as icons today: they pass the uploader's `image/` filetype check, but they render unreliably (the reported case showed an SVG displayed as a PNG that then failed when opened directly) and SVGs can embed scripts.

## Solution
Per the maintainer's suggestion that this "should just be a check in the image upload JavaScript", `invalidFileType()` in `galleries/uploader.js` now rejects `image/svg+xml`, mirroring the existing `.tiff` rejection — the user gets an alert and the upload is aborted before it reaches S3.

## Tests
Added a `:js` system spec in `spec/system/galleries/add_spec.rb` that drives the real uploader in headless Chrome: it attaches an `.svg` fixture and asserts the rejection alert fires (and that nothing gets saved). Passes locally alongside the existing gallery-add scenarios.

## Note / question
This is a client-side guard, consistent with the maintainer's scope. Because uploads go directly to S3 via a presigned policy whose only content-type condition is `starts-with image/`, a determined user could still bypass the JS. If you'd like defense-in-depth, I'm happy to follow up with a server-side restriction on the S3 policy / a model validation — let me know.


Throne3d:
Seems fair! I'm a bit concerned that we're forcing this client-side so it doesn't actually bring us any extra security, but I had a look at the presigned post logic and I'm not sure how to better handle this (we can't list specific permitted content types, only a valid prefix, so I guess we'd need a new endpoint to be called with the given mime type you need, whitelist it on our side, and then return a presigned post that's appropriate?)

Throne3d:
I'll keep #1655 open after we merge this so we can deal with this fully from a security perspective, but this at least helps from a UI perspective (so people don't accidentally upload unsupported types)

PR #2814, opened 2026-07-02
Don't clobber unread position when opening a reply permalink

Throne3d:
## Summary
- Opening a permalink to a reply used to silently auto-advance the reader's unread/read position to whatever page the reply landed on, even skipping over content they hadn't actually read. There was no way to undo this.
- `RepliesController#show` now checks whether the permalinked page advances beyond the reader's real unread boundary (`Post#first_unread_for`). If it does, the read position is left untouched for that view, and a dismissible inline notice is shown at the permalinked reply explaining why, with a link to their actual reading position.
- The notice also offers a "Mark as read up to here" button (reusing the existing "mark unread at reply" endpoint) to explicitly opt in to advancing.

## Screenshots

default_earlier_desktop

(we don't show a warning if you're on the expected page or if you're "in the past" - users won't want the noise while re-reading where they often don't adjust their unread position)

mobile view:
default_earlier_mobile

starrylight view:
starrylight_earlier_mobile

## Test plan
- [x] `bin/with-bundle rspec spec/controllers/replies/show_spec.rb` — new permalink read-position cases (never read / partially read / fully read / matching)
- [x] `bin/with-bundle rspec spec/system/replies/show_spec.rb` — new system spec, the "mark as read" button, and dismiss
- [x] `bin/with-bundle rspec spec/controllers/replies spec/controllers/posts` — no regressions (427 examples, 0 failures)
- [x] `bin/with-bundle rubocop` / `bin/with-bundle haml-lint` on touched files

## Screenshot script

<details><summary>spec/system/replies/_tmp_screenshot_spec.rb</summary>

```
rm -f /tmp/screenshots/*.png
bin/with-bundle rspec spec/system/replies/_tmp_screenshot_spec.rb 2>&1
```

```rb
RSpec.describe "Reply permalink screenshots", :js do
  def make_scenario(layout)
    user = create(:user, layout: (layout == 'default' ? nil : layout))
    post = create(:post, subject: 'Sample post about wandering the archives')
    replies = (1..12).map { |i| Timecop.freeze(post.created_at + i.minutes) { create(:reply, post: post) } }
    login(user)
    [user, post, replies]
  end

  def enable_mobile_viewport!
    page.driver.browser.manage.window.resize_to(390, 844)
    page.driver.browser.execute_cdp(
      'Emulation.setDeviceMetricsOverride',
      width: 390, height: 844, deviceScaleFactor: 3, mobile: true,
    )
  end

  def reset_viewport!
    page.driver.browser.execute_cdp('Emulation.clearDeviceMetricsOverride')
    page.driver.browser.manage.window.resize_to(1366, 768)
  end

  %w[default starrylight].each do |layout|
    context "#{layout} layout" do
      scenario "earlier notice - desktop" do
        _user, post, replies = make_scenario(layout)
        target = replies[10]
        visit reply_path(target, anchor: "reply-#{target.id}", per_page: 5)
        page.save_screenshot("/tmp/screenshots/#{layout}_earlier_desktop.png")
      end

      # (irrelevant in latest code)
      scenario "later notice - desktop" do
        user, post, replies = make_scenario(layout)
        Timecop.freeze(replies.last.created_at + 30.seconds) { post.mark_read(user) }
        target = replies[1]
        visit reply_path(target, anchor: "reply-#{target.id}", per_page: 5)
        page.save_screenshot("/tmp/screenshots/#{layout}_later_desktop.png")
      end

      scenario "earlier notice - mobile" do
        _user, post, replies = make_scenario(layout)
        enable_mobile_viewport!
        target = replies[10]
        visit reply_path(target, anchor: "reply-#{target.id}", per_page: 5)
        page.save_screenshot("/tmp/screenshots/#{layout}_earlier_mobile.png")
        reset_viewport!
      end

      # (irrelevant in latest code)
      scenario "later notice - mobile" do
        user, post, replies = make_scenario(layout)
        Timecop.freeze(replies.last.created_at + 30.seconds) { post.mark_read(user) }
        enable_mobile_viewport!
        target = replies[1]
        visit reply_path(target, anchor: "reply-#{target.id}", per_page: 5)
        page.save_screenshot("/tmp/screenshots/#{layout}_later_mobile.png")
        reset_viewport!
      end
    end
  end
end
```

</details>

### DIFF
--- a/app/assets/stylesheets/replies.scss
+++ b/app/assets/stylesheets/replies.scss
@@ -563,9 +563,19 @@ div.post-subheader { width: 100%; }
   .permalink-read-notice-message img { margin-right: 5px; }
 
   .permalink-read-notice-actions {
+    align-items: center;
+    display: flex;
+    flex-wrap: wrap;
+    gap: 10px;
     margin-top: 8px;
 
-    .link-box { margin-left: 10px; }
+    .link-box {
+      font-size: $font_size_smallish;
+      height: auto;
+      line-height: 1;
+      margin-left: 0;
+      padding: 6px 12px;
+    }
   }
 }
 
--- a/app/assets/stylesheets/replies.scss
+++ b/app/assets/stylesheets/replies.scss
@@ -556,6 +556,13 @@ div.post-subheader { width: 100%; }
 
 .reply-highlighted { border: 1px $border_color_unread_reply solid; }
 
+/* Reply permalink read-position notice */
+.permalink-read-notice {
+  margin-bottom: 10px;
+
+  .link-box { margin-left: 10px; }
+}
+
 /* History page */
 .history-header { width: 170px; }
 
--- a/app/assets/stylesheets/replies.scss
+++ b/app/assets/stylesheets/replies.scss
@@ -560,7 +560,13 @@ div.post-subheader { width: 100%; }
 .permalink-read-notice {
   margin-bottom: 10px;
 
-  .link-box { margin-left: 10px; }
+  .permalink-read-notice-message img { margin-right: 5px; }
+
+  .permalink-read-notice-actions {
+    margin-top: 8px;
+
+    .link-box { margin-left: 10px; }
+  }
 }
 
 /* History page */
--- a/app/assets/stylesheets/replies.scss
+++ b/app/assets/stylesheets/replies.scss
@@ -563,19 +563,17 @@ div.post-subheader { width: 100%; }
   .permalink-read-notice-message img { margin-right: 5px; }
 
   .permalink-read-notice-actions {
-    align-items: center;
-    display: flex;
-    flex-wrap: wrap;
-    gap: 10px;
     margin-top: 8px;
 
+    // font-size matches the .subber .link-box convention in layout.scss for smaller containers
     .link-box {
       font-size: $font_size_smallish;
       height: auto;
       line-height: 1;
-      margin-left: 0;
       padding: 6px 12px;
     }
+
+    > a:first-child .link-box { margin-left: 0; }
   }
 }
 
--- a/app/controllers/replies_controller.rb
+++ b/app/controllers/replies_controller.rb
@@ -206,7 +206,6 @@ class RepliesController < WritableController
     end
     return unless unread_page < cur_page
 
-    @skip_read_marking = true
     @permalink_jumped_ahead = true
   end
 
--- a/app/controllers/replies_controller.rb
+++ b/app/controllers/replies_controller.rb
@@ -204,7 +204,8 @@ class RepliesController < WritableController
     else
       unread.post_page(per_page)
     end
-    return if unread_page == cur_page
+    return if unread_page == cur_page # loaded our latest unread page: nothing skipped
+    return if unread_page == cur_page + 1 # loaded the page we just finished reading: not meaningfully "going back"
 
     @skip_read_marking = true
     @permalink_read_direction = unread_page < cur_page ? :earlier : :later
--- a/app/controllers/replies_controller.rb
+++ b/app/controllers/replies_controller.rb
@@ -84,6 +84,8 @@ class RepliesController < WritableController
     @page_title = @post.subject
     params[:page] ||= @reply.post_page(per_page)
 
+    check_permalink_read_position if logged_in?
+
     show_post(params[:page])
   end
 
@@ -186,6 +188,26 @@ class RepliesController < WritableController
     end
 
     @page_title = @post.subject
+  end
+
+  # Skips auto-advancing the read position when the permalinked page doesn't contain the reader's actual read/unread boundary.
+  def check_permalink_read_position
+    @permalink_reply = @reply
+    cur_page = params[:page].to_i
+    return unless cur_page > 0
+
+    unread = @post.first_unread_for(current_user)
+    unread_page = if unread.nil?
+      @post.replies.paginate(per_page: per_page, page: 1).total_pages
+    elsif unread.is_a?(Post)
+      1
+    else
+      unread.post_page(per_page)
+    end
+    return if unread_page == cur_page
+
+    @skip_read_marking = true
+    @permalink_read_direction = unread_page < cur_page ? :earlier : :later
   end
 
   def require_create_permission
--- a/app/controllers/replies_controller.rb
+++ b/app/controllers/replies_controller.rb
@@ -190,7 +190,7 @@ class RepliesController < WritableController
     @page_title = @post.subject
   end
 
-  # Skips auto-advancing the read position when the permalinked page doesn't contain the reader's actual read/unread boundary.
+  # Only skips auto-advancing when the permalink jumps ahead of the unread boundary; mark_read's own regression guard already covers the reverse case.
   def check_permalink_read_position
     @permalink_reply = @reply
     cur_page = params[:page].to_i
@@ -204,11 +204,10 @@ class RepliesController < WritableController
     else
       unread.post_page(per_page)
     end
-    return if unread_page == cur_page # loaded our latest unread page: nothing skipped
-    return if unread_page == cur_page + 1 # loaded the page we just finished reading: not meaningfully "going back"
+    return unless unread_page < cur_page
 
     @skip_read_marking = true
-    @permalink_read_direction = unread_page < cur_page ? :earlier : :later
+    @permalink_jumped_ahead = true
   end
 
   def require_create_permission
--- a/app/controllers/writable_controller.rb
+++ b/app/controllers/writable_controller.rb
@@ -132,7 +132,7 @@ class WritableController < ApplicationController
         @draft = ReplyDraft.draft_for(@post.id, current_user.id)
       end
 
-      @post.mark_read(current_user, at_time: @post.read_time_for(@replies)) unless @skip_read_marking
+      @post.mark_read(current_user, at_time: @post.read_time_for(@replies)) unless @permalink_jumped_ahead
     end
 
     if display_warnings?
--- a/app/controllers/writable_controller.rb
+++ b/app/controllers/writable_controller.rb
@@ -132,7 +132,7 @@ class WritableController < ApplicationController
         @draft = ReplyDraft.draft_for(@post.id, current_user.id)
       end
 
-      @post.mark_read(current_user, at_time: @post.read_time_for(@replies))
+      @post.mark_read(current_user, at_time: @post.read_time_for(@replies)) unless @skip_read_marking
     end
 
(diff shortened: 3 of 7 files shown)


app/assets/stylesheets/replies.scss
```diff
@@ -556,6 +556,29 @@ div.post-subheader { width: 100%; }
...
+    .link-box {
+      font-size: $font_size_smallish;
+      height: auto;
+      line-height: 1;
+      margin-left: 0;
+      padding: 6px 12px;
+    }
+  }
+}
+
```
teceler, line 579: Hm, kinda feel like this should use existing css? I think most of it is duplication from that

app/controllers/replies_controller.rb
```diff
@@ -188,6 +190,27 @@ def find_model
     @page_title = @post.subject
   end
 
+  # Skips auto-advancing the read position when the permalinked page doesn't contain the reader's actual read/unread boundary.
+  def check_permalink_read_position
+    @permalink_reply = @reply
```
teceler, line 195: Is this needed?

app/views/replies/_permalink_read_notice.haml
```diff
@@ -0,0 +1,17 @@
...
+
+.permalink-read-notice.flash.success
+  = link_to '#' do
+    = image_tag 'icons/cross.png'.freeze, class: 'float-right flash-dismiss'.freeze, alt: 'Dismiss'
+  .permalink-read-notice-message
+    = image_tag 'icons/eye.png'.freeze, class: 'vmid'.freeze, alt: ''
+    - if direction == :earlier
+      Your reading position on this thread is earlier than this page, so it has not been updated.
+    - else
+      You have already read further in this thread than this page.
```
teceler: ...this seems silly? people reread things? can we condition having this on it being relevant?

teceler: oh we do, this is just dead code, okay

Throne3d: it only does this if you're on a reply permalink _and_ back earlier in the thread - if you're re-reading a full thread I expect you'd reset your unread state so you don't accidentally lose your position (right?)

Throne3d: (maybe I lose tabs more than most people though - I usually reset my unread position on threads I reread)

spec/system/replies/show_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    target = replies[1]
+
+    visit reply_path(target, anchor: "reply-#{target.id}", per_page: 5)
+
+    within(".post-container:has(#reply-#{target.id})") do
+      expect(page).to have_selector('.permalink-read-notice', text: 'You have already read further in this thread than this page')
+      expect(page).to have_link('Go to previous position')
+      expect(page).to have_no_link('Mark read here')
+    end
+    expect(post.reload.last_read(user)).to be_the_same_time_as(replies.last.created_at + 30.seconds)
```
teceler: Shouldn't do anything

app/assets/stylesheets/replies.scss
Throne3d: There's probably an existing style I can wrap in here - the `.flash` styles didn't bring this in automatically though

app/views/replies/_permalink_read_notice.haml
teceler: Yeah i super do not reset my position when rereading

app/assets/stylesheets/replies.scss
teceler: huh, weird

spec/system/replies/show_spec.rb
Throne3d: That's what this is asserting, right? `Timecop.freeze(replies.last.created_at + 30.seconds) { post.mark_read(user) }` was what we set the unread position at earlier

app/views/replies/_permalink_read_notice.haml
Throne3d: would it bug you a lot if when rereading _and_ you click a permalink (I dunno if you do that a lot) you get this notice?

app/views/replies/_single.haml
```diff
@@ -9,6 +9,8 @@
 .post-container{class: [write, (reply == @unread && current_user.try(:visible_unread?)) ? 'reply-highlighted' : '']}
   - if reply.is_a?(Reply) && reply.id.present?
     %a.noheight{id: "reply-#{reply.id}"}= " "
+  - if reply == @permalink_reply && @permalink_read_direction.present?
+    = render 'replies/permalink_read_notice', reply: reply, post: @post, direction: @permalink_read_direction
```
teceler: Oh i see you're setting read direction earlier even if it's behind, okay, i think that's wrong but

app/assets/stylesheets/replies.scss
Throne3d: Yeah, both the image padding and the button alignment were messed up - there are definitely other places we have buttons in flashes though, I'll see if there's another style I can steal

Throne3d: Looks like `.subber` has custom styling for link-boxes to make them play nicer, but nothing to make sure they have consistent gaps between each other. Hmmmm.

app/views/replies/_permalink_read_notice.haml
Throne3d: polling people in discord, could totally remove this other version

app/controllers/replies_controller.rb
Throne3d: `@reply` is reassigned later on when we create the draft reply (for authors) in `show_post`, it's not bound long enough to work in the partial it seems

app/assets/stylesheets/replies.scss
Throne3d: I shrank the CSS a bit with 3335498f8215db5a97e4e95b76cb8ee78093775b, but there's a limit to what I could do without making it ugly like this:

image

There don't seem to be many other examples where we reformat how the buttons work together - they just "naturally" work OK if the size is right, more or less. (I guess we could improve that more broadly in our CSS so it works in a wider variety of locations)

Throne3d: Is the updated CSS ok?

app/controllers/replies_controller.rb
Throne3d: (I tried removing it but it ends up having the wrong assignment if you're an author of the thread)

app/views/replies/_permalink_read_notice.haml
Throne3d: avoided! no longer warns if you're linked a permalink to some older reply

app/views/replies/_single.haml
Throne3d: (fixed)

app/controllers/replies_controller.rb
teceler: That makes sense!

app/controllers/writable_controller.rb
```diff
@@ -132,7 +132,7 @@ def show_post(cur_page=nil)
         @draft = ReplyDraft.draft_for(@post.id, current_user.id)
       end
 
-      @post.mark_read(current_user, at_time: @post.read_time_for(@replies))
+      @post.mark_read(current_user, at_time: @post.read_time_for(@replies)) unless @skip_read_marking
```
teceler: Is this using a different variable from permalink-skipped-ahead for forward-compat?

app/views/replies/_permalink_read_notice.haml
```diff
@@ -0,0 +1,13 @@
...
+  = link_to '#' do
+    = image_tag 'icons/cross.png'.freeze, class: 'float-right flash-dismiss'.freeze, alt: 'Dismiss'
+  .permalink-read-notice-message
+    = image_tag 'icons/eye.png'.freeze, class: 'vmid'.freeze, alt: ''
+    Your reading position on this thread is earlier than this page, so it has not been updated.
+  .permalink-read-notice-actions
+    = link_to unread_path(post) do
+      .link-box.action-edit Go to previous position
+    = link_to post_path(post, unread: true, at_id: reply.id), method: :put do
+      .link-box.action-new Mark read here
```
teceler: Not sure if this should have a confirm against misclicks?

app/views/replies/_single.haml
```diff
@@ -9,6 +9,8 @@
 .post-container{class: [write, (reply == @unread && current_user.try(:visible_unread?)) ? 'reply-highlighted' : '']}
   - if reply.is_a?(Reply) && reply.id.present?
     %a.noheight{id: "reply-#{reply.id}"}= " "
+  - if reply == @permalink_reply && @permalink_jumped_ahead
+    = render 'replies/permalink_read_notice', reply: reply, post: @post
```
teceler: ...huh does equals and and combine like that?

app/assets/stylesheets/replies.scss
teceler: I'm probably gonna wanna generalize this and subber linkboxes at some point but you don't want this to be a subber so yeah this is probably the best you can do without scope creep

Throne3d: subber seemed strange to me when it's more like a header(?), and it comes with other formatting that I think is tough to handle here

Makes sense to generalize though, it'd be good if buttons worked properly in a variety of environments with our current CSS rather than needing something custom each time

app/controllers/writable_controller.rb
Throne3d: Hmm, I don't mind much - could use the same variable. I think it was for forward compat yeah but it's not a big deal

app/views/replies/_permalink_read_notice.haml
Throne3d: Good point!

app/views/replies/_single.haml
Throne3d: Yeah! I can invert the condition if you want? `@permalink_jumped_ahead && reply == @permalink_reply`? (It's the same either way, just changes which side it evaluates first and might be easier to read, but the bare `@permalink_jumped_ahead` just checks it evaluates to true and the == is a separate clause on the other side)

(e.g. `current_user.id != @character.user_id && params.dig(:character, :audit_comment).blank?` in `app/controllers/characters_controller.rb:82` is the same thing, it's just that the right-hand side is longer in that case)

teceler: Yeah that would make it less confusing to read, it currently reads like you're checking both of them match reply

teceler (approved):
Looks good!

PR #2821, opened 2026-07-03
Add minor edit option for posts and replies

l1n:
## Summary

Adds a **"Minor edit"** checkbox to the post and reply edit forms. When checked, the edit does not bump the post's `edited_at`/`tagged_at` — so it won't jump to the top of recently-updated listings or be re-marked unread for readers. Edit history is still recorded as usual.

## How it works

Rides the existing model flags rather than new machinery:
- `PostsController#update`: `@post.skip_edited = true` when `params[:minor]` is set.
- `RepliesController#update`: `@reply.skip_post_update = true` likewise.
- The checkbox only appears when editing an existing post/reply (never on create), with a tooltip explaining the effect.

Typical use: fixing a typo in an old tag without pinging every reader's unread list.

## Tests

Controller specs for both paths — minor edits leave `tagged_at` untouched and readers' unread state alone; normal edits still bump as before.


Throne3d:
This only really affects the behavior for the last reply in a thread, right? I don't think editing replies earlier in the post updates the tagged_at time.

Can we adjust this to only show up on the last reply's edit UI? Might confuse users otherwise (thinking there's extra behavior change when they're editing an older reply)

PR #2830, opened 2026-07-04
Upgrade simplecov to 1.2.0, use its native parallel_tests support

Throne3d:
- Upgrades `simplecov` to `1.2.0` (originally `1.0.0.rc5`), which drops the coverage-exceeds-source-lines check entirely (so the small-HAML-line-count-diff warning is gone without needing a patch)
- Built-in `parallel_tests` support: the first worker automatically waits for its siblings, merges their coverage, formats the report, and enforces `minimum_coverage` in-process at exit. Replaces our manual wiring.
- Drops the redundant "Merge coverage and enforce threshold" CI step

Ran the full suite locally via `bundle exec parallel_rspec spec` (4 workers).


Throne3d:
Probably won't merge until the official release comes out, but looks like quite a nice improvement!

PR #2843, opened 2026-07-04
Upgrade Ruby to 4.0.6

Throne3d:
Follow-up to https://github.com/glowfic-constellation/glowfic/pull/2617

Comments:

> Rails isn't 100% compatible with Ruby 4 yet (from what I understand), and this was only released super recently anyway. Just pushing some changes to leave an idea of the next steps here, but we should leave it for a bit.

Probably safer now but should check our dependency list I guess.


Throne3d:
Again, probably safer now, but stiiiill not confident.

PR #2855, opened 2026-07-07
Secondary continuities

redmvc:
## Summary

Posts can now belong to any number of continuities: one **main** continuity (as before) plus optional **secondary** memberships. A post can then be browsed, navigated, and shared *through* any continuity it belongs to, while keeping a single canonical identity.

Three pieces, mirroring the commits:

1. **Many-to-many model** — `posts.board_id/section_id/section_order` move to a new `post_boards` join table (`is_main` flag; per-membership section and ordering).
2. **Editor support** — an "Other continuities" multi-select on the post form, with a per-continuity section picker.
3. **Continuity-scoped viewing** — nested show routes `/boards/:continuity_id/posts/:id` (and `/replies/:id`), used by board pages when a post is there secondarily (not mandatory; if `boards/:continuity_id` is absent links will work as before pointing to the main continuity).

## Database

- New `post_boards` table: unique `(post_id, board_id)`; a partial unique index enforces exactly one `is_main` membership per post; `section_id`/`section_order` live per membership, so each continuity orders the post independently.
- The migration backfills a main membership for every existing post, then drops the old `posts` columns. Reversible.
- `Post#board_id/#section_id/#section_order` (readers and writers) delegate to the main membership, so the vast majority of existing code and queries are untouched.

## Behavior when viewing through a secondary continuity

- **Links**: a continuity's page links its posts through itself — nested URL iff the membership is secondary; a post's own main continuity always uses the plain flat URL.
- **Breadcrumbs, prev/next, pagination, reply permalinks** all follow the viewed continuity; prev/next chains include posts that are in the continuity secondarily.
- **Edit round-trips**: editing/deleting a reply, or editing the post itself, returns you to the continuity you came from.
- **Failure mode**: a continuity the post isn't in fails exactly like an unknown post id.
- **SEO**: `rel=canonical` and `og:url` always point at the flat main-continuity URL (no duplicate content); the OG title reflects the viewed continuity.
- **Unread tracking is unchanged**: read state is per-post (`Post::View`); the never-opened fallback still reads the main continuity's `BoardView`.
- **Reordering**: the reorder API takes an explicit `board_id`, so a continuity's edit pages can reorder posts that belong to it secondarily (each membership's ordering is independent).

## Testing

- New specs: membership resolution (`Post#post_board_for`), adjacency within secondary continuities (including mixed main/secondary chains), nested show (context assigns, canonical, OG, not-found), redirect round-trips for reply update/destroy and post update, reorder with explicit `board_id`, and the path helpers.

## Screenshots

*(1) The post editor's new "Other continuities" field:*
4-editor-other-continuities

*(2) The same post listed on its main continuity's page and on a secondary continuity's page:*
5-silmaril-board-listing
3-crossover-board-listing

*(3) The post viewed flat — its main continuity's breadcrumb:*
1-flat-main-continuity

*(4) …and viewed through the secondary continuity (`/boards/10/posts/38`) — that continuity's breadcrumb, with prev/next navigating its section:*
2-viewed-through-secondary

### DIFF
--- a/app/assets/javascripts/posts/editor.js
+++ b/app/assets/javascripts/posts/editor.js
@@ -66,7 +66,7 @@ function setupMetadataEditor() {
   if ($("#post_section_id").val() === '') setSections();
   $("#post_board_id").change(function() { setSections(); });
 
-  setupSecondaryBoards();
+  setupSecondaryContinuities();
 
   $("#post_privacy").change(function() {
     if ($(this).val() === 'access_list') {
@@ -511,7 +511,7 @@ function setSections() {
   }, 'json');
 }
 
-function setupSecondaryBoards() {
+function setupSecondaryContinuities() {
   const select = $('#post_secondary_board_ids');
   if (select.length === 0) return;
 
--- a/app/assets/javascripts/posts/editor.js
+++ b/app/assets/javascripts/posts/editor.js
@@ -66,6 +66,8 @@ function setupMetadataEditor() {
   if ($("#post_section_id").val() === '') setSections();
   $("#post_board_id").change(function() { setSections(); });
 
+  setupSecondaryBoards();
+
   $("#post_privacy").change(function() {
     if ($(this).val() === 'access_list') {
       $("#access_list").show();
@@ -506,6 +508,60 @@ function setSections() {
       $("#post_section_id").val("").trigger("change.select2");
       $("#section").hide();
     }
+  }, 'json');
+}
+
+function setupSecondaryBoards() {
+  const select = $('#post_secondary_board_ids');
+  if (select.length === 0) return;
+
+  createSelect2('#post_secondary_board_ids', {
+    width: '300px',
+    minimumResultsForSearch: 20,
+    placeholder: 'Choose other continuities for this post'
+  });
+
+  select.change(function() {
+    const selected = ($(this).val() || []).map(Number);
+    const container = $('#secondary-sections');
+
+    // Remove rows for boards no longer selected
+    container.find('.secondary-membership-row').each(function() {
+      const boardId = parseInt($(this).data('board-id'));
+      if (!selected.includes(boardId)) $(this).remove();
+    });
+
+    // Add rows for newly selected boards
+    selected.forEach(function(boardId) {
+      if (container.find('.secondary-membership-row[data-board-id="' + boardId + '"]').length > 0) return;
+      addSecondaryMembershipRow(boardId, container);
+    });
+  });
+}
+
+function addSecondaryMembershipRow(boardId, container) {
+  $.authenticatedGet('/api/v1/boards/' + boardId, {}, function(resp) {
+    const row = $('<div class="secondary-membership-row">').attr('data-board-id', boardId);
+    row.append($('<b>').text(resp.name));
+    row.append($('<input type="hidden">').attr({
+      name: 'post[secondary_memberships][][board_id]',
+      value: boardId
+    }));
+    if (resp.board_sections && resp.board_sections.length > 0) {
+      row.append(' section: ');
+      const sectionSelect = $('<select>').attr('name', 'post[secondary_memberships][][section_id]');
+      sectionSelect.append('<option value="">— Choose Section —</option>');
+      resp.board_sections.forEach(function(section) {
+        sectionSelect.append($('<option>').attr('value', section.id).text(section.name));
+      });
+      row.append(sectionSelect);
+    } else {
+      row.append($('<input type="hidden">').attr({
+        name: 'post[secondary_memberships][][section_id]',
+        value: ''
+      }));
+    }
+    container.append(row);
   }, 'json');
 }
 
--- a/app/concerns/owable.rb
+++ b/app/concerns/owable.rb
@@ -50,7 +50,7 @@ module Owable
     end
 
     def update_board_cameos
-      boards.uniq.each { |b| cameo_authors_into(b) }
+      boards.each { |b| cameo_authors_into(b) }
     end
 
     def cameo_authors_into(board)
--- a/app/concerns/owable.rb
+++ b/app/concerns/owable.rb
@@ -50,6 +50,10 @@ module Owable
     end
 
     def update_board_cameos
+      boards.uniq.each { |b| cameo_authors_into(b) }
+    end
+
+    def cameo_authors_into(board)
       return unless board.authors_locked?
 
       # adjust for the fact that the associations are managed separately
--- a/app/controllers/api/v1/boards_controller.rb
+++ b/app/controllers/api/v1/boards_controller.rb
@@ -42,7 +42,7 @@ class Api::V1::BoardsController < Api::ApiController
       .ordered_by_id
       .with_reply_count
       .select('posts.*')
-      .includes(:board, :joined_authors, :section)
+      .includes(:joined_authors, main_post_board: [:board, :section])
     posts = paginate(queryset, per_page: 25)
     render json: { results: posts }
   end
--- a/app/controllers/api/v1/posts_controller.rb
+++ b/app/controllers/api/v1/posts_controller.rb
@@ -65,7 +65,13 @@ class Api::V1::PostsController < Api::ApiController
     section_id = params[:section_id]&.to_i
     post_ids = params[:ordered_post_ids].map(&:to_i).uniq
 
-    post_boards = PostBoard.main.where(post_id: post_ids)
+    # Board pages pass the continuity whose ordering is being edited, since posts may be in
+    # it secondarily; without it, fall back to the posts' main continuities.
+    post_boards = if params[:board_id].present?
+      PostBoard.where(board_id: params[:board_id], post_id: post_ids)
+    else
+      PostBoard.main.where(post_id: post_ids)
+    end
     unless post_boards.count == post_ids.count
       missing_posts = post_ids - post_boards.pluck(:post_id)
       error = { message: "Some posts could not be found: #{missing_posts * ', '}" }
@@ -89,7 +95,7 @@ class Api::V1::PostsController < Api::ApiController
     end
 
     PostBoard.transaction do
-      section_scope = PostBoard.main.where(board_id: board.id, section_id: section_id)
+      section_scope = PostBoard.where(board_id: board.id, section_id: section_id)
       all_section_pbs = section_scope.ordered_in_section.to_a
       visible_post_ids = Post.where(id: all_section_pbs.map(&:post_id)).visible_to(current_user).pluck(:id).to_set
       post_id_set = post_ids.to_set
@@ -131,7 +137,7 @@ class Api::V1::PostsController < Api::ApiController
       end
     end
 
-    ordered = PostBoard.main.where(board_id: board.id, section_id: section_id).ordered_in_section.to_a
+    ordered = PostBoard.where(board_id: board.id, section_id: section_id).ordered_in_section.to_a
     visible_ids = Post.where(id: ordered.map(&:post_id)).visible_to(current_user).pluck(:id).to_set
     render json: { post_ids: ordered.map(&:post_id).select { |id| visible_ids.include?(id) } }
   end
--- a/app/controllers/api/v1/posts_controller.rb
+++ b/app/controllers/api/v1/posts_controller.rb
@@ -54,7 +54,7 @@ class Api::V1::PostsController < Api::ApiController
     end
   end
 
-  api :POST, '/posts/reorder', 'Update the order of posts. This is an unstable feature, and may be moved or renamed; it should not be trusted.'
+  api :POST, '/posts/reorder', 'Update the order of posts within their main continuity. This is an unstable feature, and may be moved or renamed; it should not be trusted.'
   error 401, "You must be logged in"
   error 403, "Continuity is not editable by the user"
   error 404, "Post IDs could not be found"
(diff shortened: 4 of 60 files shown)


app/services/reply/searcher.rb
```diff
@@ -65,12 +65,16 @@ def sort_results(sort, content_search)
...
     if @post
       @search_results = @search_results.where(post_id: @post.id)
-    else
+    elsif board_id.present? || exclude_board_ids.present?
       # Subqueries instead of pluck+IN so PG can use a hash/index join on
       # board_id rather than serializing a thousands-long id list and
       # falling off the index. Has caused statement_timeout in production.
-      @search_results = @search_results.where(post_id: Post.where(board_id: board_id).select(:id)) if board_id.present?
-      @search_results = @search_results.where.not(post_id: Post.where(board_id: exclude_board_ids).select(:id)) if exclude_board_ids.present?
+      if board_id.present?
```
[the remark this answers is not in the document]
redmvc: Bro you were literally the one who asked me to break this into multiple lines...

app/assets/javascripts/posts/editor.js
```diff
@@ -66,6 +66,8 @@ function setupMetadataEditor() {
   if ($("#post_section_id").val() === '') setSections();
   $("#post_board_id").change(function() { setSections(); });
 
+  setupSecondaryBoards();
```
teceler: Okay i'm gonna be extremely fussy here: Continuities. so we don't have to change it later xD

app/assets/javascripts/posts/editor.js
```diff
@@ -509,6 +511,60 @@ function setSections() {
   }, 'json');
 }
 
+function setupSecondaryBoards() {
```
teceler: please have someone who can read javascript look at this

app/concerns/owable.rb
```diff
@@ -50,6 +50,10 @@ def add_creator_to_authors
     end
 
     def update_board_cameos
+      boards.uniq.each { |b| cameo_authors_into(b) }
```
teceler: Why does this have a uniq?

app/controllers/api/v1/boards_controller.rb
```diff
@@ -42,7 +42,7 @@ def posts
       .ordered_by_id
       .with_reply_count
       .select('posts.*')
-      .includes(:board, :joined_authors, :section)
+      .includes(:joined_authors, main_post_board: [:board, :section])
```
teceler, line 45: is main_post_board an alias?

app/controllers/api/v1/posts_controller.rb
```diff
@@ -64,76 +64,82 @@ def update
   def reorder
     section_id = params[:section_id]&.to_i
     post_ids = params[:ordered_post_ids].map(&:to_i).uniq
-    posts = Post.where(id: post_ids)
-    posts_count = posts.count
-    unless posts_count == post_ids.count
-      missing_posts = post_ids - posts.pluck(:id)
+
+    # Board pages pass the continuity whose ordering is being edited, since posts may be in
+    # it secondarily; without it, fall back to the posts' main continuities.
```
teceler, line 69: Having trouble understanding what the first clause is even saying

app/controllers/api/v1/posts_controller.rb
```diff
@@ -64,76 +64,82 @@ def update
...
+    post_boards = if params[:board_id].present?
+      PostBoard.where(board_id: params[:board_id], post_id: post_ids)
```
teceler, line 71: do posts have post_boards for their main board?

app/controllers/api/v1/posts_controller.rb
```diff
@@ -64,76 +64,82 @@ def update
...
     end
 
-    board = boards.first
+    board = Board.find(board_ids.first)
     access_denied and return unless board.editable_by?(current_user)
 
-    post_section_ids = posts.select(:section_id).distinct.pluck(:section_id)
-    unless post_section_ids == [section_id] &&
+    pb_section_ids = post_boards.distinct.pluck(:section_id)
+    unless pb_section_ids == [section_id] &&
```
teceler: slightly confusing shortening there

app/controllers/posts_controller.rb
```diff
@@ -136,14 +136,21 @@ def create
...
     begin
-      @post.save!
+      Post.transaction do
+        @post.assign_continuities(
+          main_board_id: permitted_params[:board_id],
+          main_section_id: permitted_params[:section_id],
+          secondaries: permitted_params[:secondary_memberships],
+        )
+        @post.save!
+      end
```
teceler, line 153: Hmm. Not sure how I feel about how  much  more complicated it makes this.

app/controllers/posts_controller.rb
```diff
@@ -250,7 +263,7 @@ def update
       render :edit
     else
       flash[:success] = "Post updated."
-      redirect_to @post
+      redirect_to post_path_with_continuity(@post)
```
teceler, line 266: It also seems sad we have to complicated this

app/controllers/replies_controller.rb
```diff
@@ -59,10 +59,10 @@ def create
       flash[:success] = "Replies discarded."
       if editing_multi_reply?
         # Editing multi reply, going to redirect back to the reply I'm editing
-        redirect_to reply_path(@reply, anchor: "reply-#{@reply.id}")
+        redirect_to reply_path_with_continuity(@reply, anchor: "reply-#{@reply.id}")
```
teceler, line 62: Hm, actually, why *are* we renaming these, you're  not even passing the board?

app/helpers/post_helper.rb
```diff
@@ -28,8 +28,32 @@ def allowed_boards(obj, user)
...
+  # continuity or nil.
+  def post_in_board_path(post, board=nil, **opts)
+    return post_path(post, **opts) if board.nil? || board.id == post.board_id
+    continuity_post_path(board, post, **opts)
+  end
+
+  # As above; callers only pass secondary continuities (or nil), so no main check needed.
+  def reply_in_board_path(reply, board=nil, **opts)
+    return reply_path(reply, **opts) if board.nil?
+    continuity_reply_path(board, reply, **opts)
```
teceler: Again, you probably want to name these methods with continuity

app/models/post.rb
```diff
@@ -291,12 +297,130 @@ def user_joined(user)
...
-    adjacent_posts_for(user) { |relation| relation.find_by('section_order > ?', self.section_order) }
+  def next_post(user, post_board=main_post_board)
+    adjacent_posts_for(user, post_board) do |relation|
+      relation.where('post_boards.section_order > ?', post_board.section_order).first
+    end
+  end
+
+  # Reconcile post_boards (main + secondaries) against the form's submitted state.
+  # Caller must wrap in a transaction with post.save!.
+  def assign_continuities(main_board_id:, main_section_id: nil, secondaries: nil)
```
teceler, line 384: i need to look at this when less braindead but it's scary?

app/models/post.rb
```diff
@@ -311,23 +435,49 @@ def reply_word_count_sum(scope)
...
-    return unless new_record? || board_id_changed?
-    return if board.open_to?(user)
-    errors.add(:board, "is invalid – you must be able to write in it")
+  def validate_post_boards
+    # Force-load post_boards so validation sees DB state for existing posts where
+    # the has_many target wasn't otherwise populated (e.g. main was loaded only via
+    # the has_one :main_post_board cache).
+    post_boards.load if persisted? && !post_boards.loaded?
+
+    candidates = post_boards.target.dup
```
teceler, line 448: what...

app/views/board_sections/edit.haml
```diff
@@ -12,7 +12,7 @@
     .editor-title Edit #{@board_section.name}
     = render 'editor', f: f
 
-- if @posts.present?
+- if @section_posts.present?
```
teceler: any particular reason this got renamed?

spec/controllers/api/v1/posts_controller_spec.rb
```diff
@@ -341,6 +341,30 @@
...
+
+        api_login_as(coauthor)
+        post :reorder, params: { ordered_post_ids: [visible2.id, visible1.id] }
+        expect(response).to have_http_status(200)
+
+        # both hidden posts followed visible1 → they stay, in original order, right after it
+        expect(visible2.reload.section_order).to eq(0)
+        expect(visible1.reload.section_order).to eq(1)
+        expect(hidden1.reload.section_order).to eq(2)
+        expect(hidden2.reload.section_order).to eq(3)
```
teceler:
```suggestion
        user = create(:user)
        coauthor = create(:user)
        board = create(:board, creator: user, writers: [coauthor])
        visible1 = create(:post, board: board, user: user)
        hidden = create_list(:post, 2, board: board, user: user, privacy: :private)
        visible2 = create(:post, board: board, user: user)
        posts = [visible1, *hidden, visible2]

        posts.each(&:reload)
        expect(posts.map(&:section_order).to eq([0, 1, 2, 3]

        api_login_as(coauthor)
        post :reorder, params: { ordered_post_ids: [visible2.id, visible1.id] }
        expect(response).to have_http_status(200)

        # both hidden posts followed visible1 → they stay, in original order, right after it
        posts.each(&:reload)
        expect(posts.map(&:section_order).to eq([1, 2, 3, 0]
```

teceler (commented):
Okay I did not read all the tests, but overall I'm not sure why we're doing this as full many-many when there's an established main post and infra around that, I realize not doing that would also require some workarounds but probably less scary ones?

This is a lot of overhead and i'm not sure it's worth it if secondary continuities are relatively rare. The version that special-cases main ones might be better that way

app/assets/javascripts/posts/editor.js
redmvc: Don't we use "Boards" everywhere?

app/concerns/owable.rb
redmvc: Good... question... I don't remember, I'll look at it in a bit.

app/controllers/api/v1/posts_controller.rb
redmvc: Yeah, it's fully many-to-many.

app/controllers/api/v1/posts_controller.rb
redmvc: The continuity whose ordering is being edited is passed as an argument when editing ordering. If no continuity is passed as argument, assume I'm editing the ordering of the main board.

app/controllers/posts_controller.rb
redmvc: I think even if we _weren't_ making the relation fully many-to-many we'd still need something like this? To assign all the secondary continuities.

app/controllers/posts_controller.rb
redmvc: I'm open to alternatives but couldn't think of one.

app/controllers/replies_controller.rb
redmvc: I rewrote these so many times it might be an artefact of when I did. Or maybe I forgot to pass when I should've. I'll need to go check, the point was that I wanted to go back to the previous continuity every time.

app/helpers/post_helper.rb
redmvc: Yeah fair I'll do that.

app/models/post.rb
redmvc: This is a big change, it is expected to be kinda scary :P

app/assets/javascripts/posts/editor.js
redmvc: Nvm I'll just change it to continuities everywhere then.

spec/controllers/api/v1/posts_controller_spec.rb
```diff
@@ -341,6 +341,27 @@
...
+        user = create(:user)
+        coauthor = create(:user)
+        board = create(:board, creator: user, writers: [coauthor])
+        visible1 = create(:post, board: board, user: user)
+        hidden = create_list(:post, 2, board: board, user: user, privacy: :private)
+        visible2 = create(:post, board: board, user: user)
+        posts = [visible1, *hidden, visible2]
+
+        posts.each(&:reload)
+        expect(posts.map(&:section_order).to eq([0, 1, 2, 3]
```
teceler:
```suggestion
        expect(posts.map(&:section_order)).to eq([0, 1, 2, 3])
```

spec/controllers/api/v1/posts_controller_spec.rb
```diff
@@ -341,6 +341,27 @@
...
+
+        api_login_as(coauthor)
+        post :reorder, params: { ordered_post_ids: [visible2.id, visible1.id] }
+        expect(response).to have_http_status(200)
+
+        # both hidden posts followed visible1 → they stay, in original order, right after it
+        posts.each(&:reload)
+        expect(posts.map(&:section_order).to eq([1, 2, 3, 0]
```
teceler:
```suggestion
        expect(posts.map(&:section_order)).to eq([1, 2, 3, 0])
```
oops!!

redmvc:
> Okay I did not read all the tests, but overall I'm not sure why we're doing this as full many-many when there's an established main post and infra around that, I realize not doing that would also require some workarounds but probably less scary ones?
> 
> This is a lot of overhead and i'm not sure it's worth it if secondary continuities are relatively rare. The version that special-cases main ones might be better that way

The reason I wanted to do it as full many-many is because not doing that would mean having to have exceptions everywhere for what's a main board and what isn't in the same way we have with the top post/reply distinction. Also it's inelegant.

Plus I think "it's scary" is not really an objection? Some features are big and they'll be scary, that's, like, fine, unless we never want to implement big features we'll have to deal with them sometimes.

app/views/board_sections/edit.haml
redmvc: Newp

app/controllers/api/v1/boards_controller.rb
redmvc: Yeah from the post model
```ruby
  has_one :main_post_board, -> { where(is_main: true) }, class_name: 'PostBoard',
    inverse_of: :post, autosave: true, validate: false, dependent: false # Destroyed via has_many :post_boards
```

app/concerns/owable.rb
redmvc: I think I just had a brain fart abt what "Owables" are, it's unnecessary

app/models/post.rb
redmvc: Claude invented this bit and I forgot to remove it when I was validating, gonna be gone in a sec.

app/controllers/replies_controller.rb
redmvc: Oh right no this one grabs the board from the URL itself.
```ruby
  def reply_path_with_continuity(reply, **opts)
    helpers.reply_in_continuity_path(reply, secondary_board_for(reply.post), **opts)
  end

  def secondary_board_for(post)
    return if params[:continuity_id].blank?
    pb = post.post_board_for(params[:continuity_id])
    pb.board unless pb.nil? || pb.is_main?
  end

  def reply_in_continuity_path(reply, board=nil, **opts)
    return reply_path(reply, **opts) if board.nil?
    continuity_reply_path(board, reply, **opts)
  end
```

PR #2859, opened 2026-07-09
Fix unread markers post splitting post

redmvc:
As per the polls I ran on Discord, the consensus for how unread markers should work after splitting a post is:
- if your unread marker was before the split, nothing changes;
- if your unread marker was just before the split, new post is fully unread, old post gets a "0 unread replies" notification as if the last reply had been edited;
- if your unread marker was after the split but before the very end, old post gets "0 unread replies" and new post gets unread up to your actual existing marker;
- otherwise, both posts are marked as having "0 unread replies".

Depends on #2858.


redmvc:
Another fix to add to this PR: probably a split should enqueue generate flat post jobs.

PR #2871, opened 2026-07-18
Fix AnonLoadShed reading the wrong rack-timeout env key

l1n:
## Summary

`AnonLoadShed` (added in #2764) has been inert since it shipped: it reads `env['rack.timeout.info']`, but rack-timeout stores its `RequestDetails` under `"rack-timeout.info"` — hyphen, not dot (`Rack::Timeout::ENV_INFO_KEY`, [core.rb:46](https://github.com/zombocom/rack-timeout/blob/v0.7.0/lib/rack/timeout/core.rb#L46)). The lookup always returned nil, so `waited.nil?` was always true and every request passed through unshed.

This showed up in production on 2026-07-18 at ~21:17 UTC: a distributed scraper (rotating desktop UAs) spiked traffic ~20× for two minutes, Puma queue waits hit the 20s rack-timeout limit, and ~2,400 requests — logged-in users included — got 500s. New Relic shows **zero** 503s from the shedder during the event; the timeout backtraces all pass through `anon_load_shed.rb`'s pass-through branch. Had the shedder worked, anonymous crawler traffic would have been 503'd at 5s wait and workers freed for logged-in users.

## Fix

Use `Rack::Timeout::ENV_INFO_KEY` — the gem's constant, not a string literal — in both the middleware and the spec, so the key can't silently drift from the gem again. The spec previously hand-built its fake env with the same wrong string, which is why all its examples passed against broken code; it now exercises the real key.

## Testing

`spec/middleware/anon_load_shed_spec.rb` — 7 examples, 0 failures, run against the real rack-timeout 0.7.0 gem.

### DIFF
--- a/app/middleware/anon_load_shed.rb
+++ b/app/middleware/anon_load_shed.rb
@@ -53,10 +53,12 @@ class AnonLoadShed
     env['PATH_INFO'] == '/login'
   end
 
-  # rack-timeout populates env['rack.timeout.info'].wait with the seconds
-  # the request spent in the dyno's queue before reaching a worker.
+  # rack-timeout stores its RequestDetails (including .wait, the seconds the
+  # request spent in the dyno's queue before reaching a worker) under
+  # Rack::Timeout::ENV_INFO_KEY. Resolved at call time so this file can be
+  # required before Bundler loads the gem.
   def wait_seconds(env)
-    info = env['rack.timeout.info']
+    info = env[Rack::Timeout::ENV_INFO_KEY]
     info&.wait
   end
 end
--- a/app/middleware/anon_load_shed.rb
+++ b/app/middleware/anon_load_shed.rb
@@ -55,10 +55,11 @@ class AnonLoadShed
 
   # rack-timeout stores its RequestDetails (including .wait, the seconds the
   # request spent in the dyno's queue before reaching a worker) under
-  # Rack::Timeout::ENV_INFO_KEY. Resolved at call time so this file can be
-  # required before Bundler loads the gem.
+  # Rack::Timeout::ENV_INFO_KEY. The gem is production-only, so resolve the
+  # constant defensively: where it isn't loaded there is no queue-wait info
+  # and we never shed.
   def wait_seconds(env)
-    info = env[Rack::Timeout::ENV_INFO_KEY]
-    info&.wait
+    return nil unless defined?(Rack::Timeout::ENV_INFO_KEY)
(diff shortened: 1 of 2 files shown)


app/middleware/anon_load_shed.rb
```diff
@@ -53,10 +53,13 @@ def login_request?(env)
...
+  # rack-timeout stores its RequestDetails (including .wait, the seconds the
+  # request spent in the dyno's queue before reaching a worker) under
+  # Rack::Timeout::ENV_INFO_KEY. The gem is production-only, so resolve the
+  # constant defensively: where it isn't loaded there is no queue-wait info
+  # and we never shed.
   def wait_seconds(env)
-    info = env['rack.timeout.info']
-    info&.wait
+    return nil unless defined?(Rack::Timeout::ENV_INFO_KEY)
+    env[Rack::Timeout::ENV_INFO_KEY]&.wait
```
teceler, line 63: So in dev or test (~~do we test this behavior?~~ I see later we do but stub it there) this returns nil, and on staging if we don't load it there but I expect we do.

Is there a reason you're checking the key exists and then still safe-navigating? I'd expect running wait on nil to error anyway.

l1n: The two guards stop two different nil conditions.

The `defined?` guard is for the constant. In dev and test, the gem is not in the bundle. There, a reference to the constant causes a NameError, not a nil value. If staging uses the production gem group, the constant is available there.

The `&.` guard is for the env entry. The gem can be loaded while the entry is not set. Today, rack-timeout inserts itself at the top of the middleware stack. AnonLoadShed is at the bottom. So in production, the entry is always set before this code runs.

The `&.` is protection against a future change in the stack order. This middleware must fail open. If the entry is not set, the middleware must not shed. It must pass the request through. Without `&.`, each anonymous request gets a 500 error from `nil.wait`. That is the failure this middleware exists to prevent.

I can remove the `&.` if you prefer a loud failure.

On the test question: the spec stubs the constant only when the gem is absent. When the gem is present, the spec uses the real constant. This finds drift between our key and the key in the gem.

teceler: Aha, that makes sense. You might want Marri's opinion on loud vs quiet failure tho I'm sure we'd hear from the community if anon browsing stopped working as soon as someone's browser logged them out.

Marri (approved):
Good with quiet fails for this one

PR #2874, opened 2026-07-19
Scope reply audit lookups by auditable_type

l1n:
## Summary

New Relic shows `ActiveRecord::QueryCanceled` (Postgres statement timeout) 500ing `POST /replies` four times this past week. The slow-query log pins it to the audit count in the multi-reply duplication check:

```
SELECT COUNT(*) ... FROM audits WHERE auditable_id = $1 GROUP BY auditable_id
```

`audits` only has `(auditable_type, auditable_id, version)` and `(associated_type, associated_id)` indexes — a bare `auditable_id` filter can't use either, so it sequential-scans the entire audit history and times out under load, killing the user's reply submission.

Three lookups in `RepliesController` had this shape (the dupe-check group-count, and both audit lookups in `restore`). Adding `auditable_type: 'Reply'`:

- makes all three use the existing `auditable_index`, and
- fixes a latent correctness issue — bare `auditable_id` also matches **Post** audits sharing the same numeric id, so `restore` keyed off whatever same-id `destroy` audit sorted last regardless of type, and the dupe-check counts could include another model's audits.

## Testing

Covered by the existing restore/create controller specs; no behavior change for correctly-typed audits.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -172,7 +172,11 @@ class PostsController < WritableController
     audit_ids = @post.associated_audits.where(action: 'destroy').where(auditable_type: 'Reply') # all destroyed replies
     audit_ids = audit_ids.joins('LEFT JOIN replies ON replies.id = audits.auditable_id').where(replies: { id: nil }) # not restored
     audit_ids = audit_ids.group(:auditable_id).pluck(Arel.sql('MAX(audits.id)')) # only most recent per reply
-    @deleted_audits = Audited::Audit.where(id: audit_ids).paginate(per_page: 1, page: page)
+    # `audit_ids` is already restricted to Reply audits above, so this scope
+    # cannot change the result today. It is stated anyway: an audit id is only
+    # meaningful together with its type, and leaving that to be inferred three
+    # lines up is how the bug this fixes was written in the first place.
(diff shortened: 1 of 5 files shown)


teceler (commented):
Ooof, yeah, good catch this is not a great bug to have. Can you also get the instance in PostController#delete_history and WritableControler#show_post? (Search also turned up a few in specs but that's less concerning if probably still bad practice)

l1n:
I pushed commit c6bdf9c7d. It adds the type scope to `show_post` and to the three spec queries.

I did not change `delete_history`. That query chain gets the type scope on line 172 (`.where(auditable_type: 'Reply')`). The `group(:auditable_id)` on line 174 operates inside that scope.

teceler:
What i was noticing on delete_history is the second query, not the first
```@deleted_audits = Audited::Audit.where(id: audit_ids).paginate(per_page: 1, page: page)```

Otherwise looks good through!

l1n:
I pushed commit c9c1a9e0d. It adds `auditable_type: 'Reply'` to the second query in `delete_history`.

The scope does not change the result today. `audit_ids` on lines 172 to 174 is already restricted to Reply audits, and line 175 selects by those primary keys. The query therefore cannot return an audit of another type.

I made the change anyway. An audit id is meaningful only together with its type. To leave that fact three lines up is how this bug was written the first time.

PR #2885, opened 2026-09-09
Shed the scrape before the readers queued behind it

l1n:
> [!IMPORTANT]
> This branch is stacked on #2880. It contains that commit as well, until #2880 merges. Review only the second commit, `Shed the scrape before the readers queued behind it`. Merge #2880 first.

## Problem

`AnonLoadShed` gave every anonymous request the same 5 second queue-wait budget. When the queue became deep, the shedder divided the damage between the scrape and the people who read the site. It did not aim the damage.

These are the HTML navigations in the seven days to 2026-09-09:

| Class | Requests | Dyno-hours | Average |
|---|---|---|---|
| Forged Chrome | 9,745,482 | 468.6 | 0.173 s |
| Real browsers | 2,744,787 | 124.9 | 0.162 s |

The scrape is 79 percent of the compute that serves pages. In the same period the shedder turned away 65,595 genuine page loads, which is 2.4 percent of real readers.

The per-IP throttle does not reach the scrape. It caught 9,696 of 9,745,482 requests, which is 0.1 percent. The scrape comes from more than 300 source IP addresses.

## Change

Requests that have the shape of the scrape are shed at `SCRAPER_WAIT_THRESHOLD_SECONDS`, which is an order of magnitude below the reader threshold. When the queue is deep, the scrape loses its thread slot first.

This is still a threshold and not a block. Below it, nothing changes for anyone. It acts only when the queue is deep enough that the shedder must turn away somebody. It decides who that is.

## Classification

A request is scraper-shaped when all three conditions are true:

1. The User-Agent claims Chrome.
2. The Accept header starts with `text/html`.
3. The Accept header does not contain `signed-exchange`.

Chrome announces `application/signed-exchange;v=b3;q=0.7` on HTML navigations. Every genuine Chrome release from 120 to 141 measures at 95 to 100 percent. The forged User-Agent strings of the scrape measure at 0.0 to 0.3 percent.

Each condition is necessary. Condition 1 keeps the test safe: Firefox and Safari also do not send the token, so a test on Accept alone classifies all of their users as scrapers. Condition 3 makes the test useful, because the User-Agent strings are forged and real readers share them. Condition 2 limits the test to navigations, so the subresources of a page are not judged apart from the page.

This is a header-level heuristic. It is the most forgeable class of signal. It therefore informs a threshold and not a block.

## Test

- 12 new examples in `spec/middleware/anon_load_shed_spec.rb`. They cover the two thresholds, real Chrome, Firefox, subresources, an empty environment, and the login, session and remembered-cookie exemptions at the new threshold.
- Validated against production traffic. 200 real Accept and User-Agent pairs, which cover 4,852,753 requests, were replayed through the middleware. The result was 0 of 1,112,924 real browser requests classified as scraper-shaped.

https://claude.ai/code/session_01HkFVi2WDjaNzjzqjwu2pNY


app/middleware/anon_load_shed.rb
```diff
@@ -21,19 +21,41 @@
...
+  # In the seven days to 2026-09-09 the scrape was 9.7M of the 12.5M HTML
+  # navigations and 469 of the 594 dyno-hours spent serving them, while 65,595
+  # genuine page loads (2.4%) were shed as collateral. Both classes were
+  # competing on the same 5s threshold, so the shedder was splitting the
+  # damage between them instead of aiming it.
+  #
+  # This is still a threshold and not a block: below it, scraper-shaped
+  # traffic is served exactly as before. It only bites once the queue is deep
+  # enough that somebody is going to be shed regardless, and it decides who.
+  SCRAPER_WAIT_THRESHOLD_SECONDS = 0.5
```
Throne3d, line 38: This is the time spent in the router before we _start_ serving the request, so it being super low (~500ms) seems OK to me. If people (not scrapers) complain after we merge, we can always adjust. Cool.

PR #2910, opened 2026-09-12
Bump load_defaults to 8.1

Throne3d:
## load_defaults(8.1)

- `action_controller.escape_json_responses = false` - stops escaping `<>&` and U+2028/2029 in `render json:` bodies. Not an issue for us, idk why they would need to be escaped if you use a proper JSON parser.
- `action_controller.action_on_path_relative_redirect = :raise` - raises on `redirect_to` targets that look like a relative path without a leading slash. `bookmarks_controller.rb` has `redirect_to` with request-controlled input (Referer) but shouldn't trip in practice.
- `active_record.raise_on_missing_required_finder_order_columns`, `active_support.escape_js_separators_in_json`, `action_view.render_tracker`, `action_view.remove_hidden_field_autocomplete` 👀 

We set `config.yjit = false`, overriding load_defaults(8.1) too.

## 8.1's other new deprecations (nothing to do here)

Went through everything else 8.1 deprecates (removal targeted for 8.2/9.0) to see if this app needed to change anything ahead of time. All clear, no code changes needed [validated with Claude]:

- `insert_all`/`upsert_all` with unpersisted records in associations - not used anywhere.
- `update_all` combined with `WITH`/`WITH RECURSIVE`/`DISTINCT` - checked every `update_all` call site (`character/relocator.rb`, `replies_controller.rb`, `boards_controller.rb`, `tag.rb`, etc.); none of the relations they're called on carry a `.distinct` or a CTE through to the `update_all` itself (the `.distinct` calls nearby are on separate lookup queries feeding into a fresh `where(id: ...)`).
- `String#mb_chars` / `ActiveSupport::Multibyte::Chars` - not used.
- `ActiveSupport::Configurable` - not used directly by app code.
- `ActiveRecord::Base.signed_id_verifier_secret` - not used.
- Custom Active Job serializers needing a `#klass` method - we don't define any custom serializers.
- Built-in `sidekiq`/`SuckerPunch` adapters - we use Resque.

Also confirmed none of the things 8.0->8.1 hard-removed were in use here: MySQL's `:unsigned_float`/`:unsigned_decimal` and SQLite3's `:retries`, the `:azure` Active Storage service, multi-path routes, leading-bracket param parsing, semicolon query separators, `STATS_DIRECTORIES`/`bin/rake stats`, `rails/console/methods.rb`.


config/application.rb
```diff
@@ -64,7 +64,7 @@ def self.full(text)
   # in config/environments, which are processed later.
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 8.0
+    config.load_defaults 8.1
```
Throne3d, line 67: (No associated config/initializers/new_framework_defaults_8_1.rb - added it temporarily in testing, none of the changes impacted anything, so skipped here)

